PR #497, opened 2021-01-19
Implement Accessor and Schema API

tamargrey:
- Pull request for implementing the Schema and Accessor API to which all related PRs will be merged.


README.md
```diff
@@ -39,60 +40,63 @@ conda install -c conda-forge woodwork
...
 ```python
 import woodwork as ww
 
-data = ww.demo.load_retail(nrows=100, return_dataframe=True)
-dt = ww.DataTable(data, name='retail')
-dt.set_types(logical_types={
-    'quantity': 'Double',
-    'customer_name': 'Categorical',
-    'country': 'Categorical'
+df = ww.demo.load_retail_to_accessor(nrows=100, init_woodwork=False)
```
kmax12, line 48: i wonder if we should just do `pd.read_csv('https://api.featurelabs.com/datasets/online-retail-logs-2018-08-28.csv') here? that would make it more clear what the type is compared to the function call.

I might also print out `df.head(5)` so people can see the sample of data to compare to the `df.ww` output below

thehomebrewnerd: @kmax12 We can certainly use `pd.read_csv` here. Note, we will be changing `load_retail_to_accessor` to `load_retail` in #709. We could also change the default for `init_woodwork` to be `False` which would make this call
```
df = ww.demo.load_retail(nrows=100)
```

Would you prefer that or the `read_csv` option?

gsheni: I think with `read_csv` we make it clear that this is a normal pandas DataFrame, whereas with `load_retail` its not clear to the user.

thehomebrewnerd: One other thing I just noticed - the `load_retail` function also adds an index column to the dataframe before returning it. If we used `pd.read_csv` we will need to update our examples to manually add that index column before we demonstrate any index behaviors that use that column. Alternatively, we could use `make_index` during initialization, but that might not work for all our examples - would need to investigate more.

If we do go the `read_csv` route, should we remove `load_retail` completely, or is there a need to keep it for something else?

gsheni: Let's keep `load_retail`. It will be useful to have for Guides and some other docs pages. 

However, for the ReadME and Get Started page, using a simple `read_csv` would be better. 

We can use the `index_col` when call `read_csv` to specify the index column - [pandas read csv docs](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html).

gsheni: Fixed.

gsheni (approved):

Issue #487, opened 2021-01-15
Allow initializing Accessor with schema


tamargrey:
allows for `df.ww.init(schema)` that 
- checks whether the schema is valid according to `is_valid_schema`
- initializes that schema as the schema

TBD: Do we just set the schema and or do we actually try and initialize the Schema object with the DataFrame, doing further validation checks?

PR #522, opened 2021-01-28
Allow initializing accessor with valid schema

### LINKED ISSUE #486: Add check for whether a schema is valid
Adds a check `Accessor.is_valid_schema(schema)` that checks that the schema provided is valid for the Accessor's dataframe.

An incompatible schema change will be defined as one where any of the following are true when comparing the old schema to the new DataFrame:
* The set of columns in the old schema does not exactly match the columns in the new DataFrame
* The dtype for each column in the schema does not exactly match the corresponding column’s dtype in the new DataFrame
* The index set in the old schema does not match the index of the DataFrame

tamargrey:
- Adds `_is_valid_schema` check that confirms whether a schema passed to `Accessor.init` is valid for the dataframe.
- Allows initializing the Accessor's schema with a valid Schema object.
- Closes #486 , #487

### DIFF
--- a/woodwork/table_accessor.py
+++ b/woodwork/table_accessor.py
@@ -25,7 +25,7 @@ class WoodworkTableAccessor:
         self._dataframe = dataframe
         self._schema = None
 
-    def init(self, schema=None, index=None, time_index=None, logical_types=None, make_index=False, already_sorted=False, **kwargs):
+    def init(self, index=None, time_index=None, logical_types=None, make_index=False, already_sorted=False, schema=None, **kwargs):
         '''Initializes Woodwork typing information for a DataFrame.
 
         Args:
@@ -57,15 +57,13 @@ class WoodworkTableAccessor:
             use_standard_tags (bool, optional): If True, will add standard semantic tags to columns based
                 specified logical type for the column. Defaults to True.
             column_descriptions (dict[str -> str], optional): Dictionary mapping column names to column descriptions.
+            schema (Woodwork.Schema, optional): Typing information to use for the DataFrame instead of performing inference.
+                Any other arguments provided will be ignored.
         '''
+        _validate_accessor_params(self._dataframe, index, make_index, time_index, logical_types, schema)
         if schema is not None:
-            if not _is_valid_schema(self._dataframe, schema):
-                raise ValueError('Invalid typing information presented at init. Cannot set Woodwork on DataFrame.')
-            else:
-                self._schema = schema
+            self._schema = schema
         else:
-            _validate_accessor_params(self._dataframe, index, make_index, time_index, logical_types)
-
             if make_index:
                 _make_index(self._dataframe, index)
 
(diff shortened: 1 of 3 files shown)


woodwork/table_accessor.py
```diff
@@ -57,37 +57,41 @@ def init(self, index=None, time_index=None, logical_types=None, make_index=False
...
+                Any other arguments provided will be ignored.
         '''
-        _validate_accessor_params(self._dataframe, index, make_index, time_index, logical_types)
-
-        if make_index:
-            _make_index(self._dataframe, index)
+        _validate_accessor_params(self._dataframe, index, make_index, time_index, logical_types, schema)
+        if schema is not None:
+            self._schema = schema
+        else:
```
tamargrey, line 66: Right now any extra arguments provided beyond `schema` will be ignored when a schema is passed in. I'm not sure this is ideal, but I think we're going to promote using `df.ww.sample` instead of calling `df.sample` and then needing to directly call the init, so it's probably not worth restricting which parameters you can pass in.

thehomebrewnerd: Should/could we add a validation check in `_validate_accessor_params` to confirm that if `schema` is not None, that all the other params are none and either raise a Warning or Error instead of silently ignoring? Probably not high priority at the moment, but maybe something we could do eventually.

tamargrey: We definitely could. I think it'd make a lot of sense (and probably more sense than quietly ignoring the non-schema params--maybe we just raise a warning?), but figured I'd keep things simple for now

thehomebrewnerd: Seems fine to keep simple for now and tackle this later - maybe capture in an issue so we don't forget?

tamargrey: created #523

woodwork/table_accessor.py
```diff
@@ -251,3 +264,17 @@ def _update_column_dtype(series, logical_type, name):
...
     return series
+
+
+def _is_valid_schema(dataframe, schema):
+    if set(dataframe.columns) != set(schema.columns.keys()):
+        return False
+    for name in dataframe.columns:
+        if str(dataframe[name].dtype) != schema.logical_types[name].pandas_dtype:
+            return False
+    if schema.index is not None:
```
gsheni, line 275: Do you need to check the time index?

tamargrey: I don't think so?

The things to check would be whether the data for a possible `schema.index` is an acceptable Datetime or Numeric Datetime, but since the time_index is set in the Schema, we know it has to be a `Datetime` logical type or have the `numeric` tag, and since we also check that the dtype of each series matches its Logical Type, I think that means that the column should only contain valid time index data. 

But I definitely could be missing something to check for the time index here.

gsheni: Gotcha.

gsheni (approved):
lgtm

PR #537, opened 2021-02-02
Automated Latest Dependency Updates

machineFL:
This is an auto-generated PR with dependency updates. Please do not delete the dep-update branch because it's needed by the auto-dependency bot.


gsheni (approved):

gsheni:
We realize this behavior is odd (numpy 1.20.0 is the latest). However, this is due to koalas not support numpy 1.20.0 yet.

PR #538, opened 2021-02-02
Add ability to call DataFrame methods from Accessor

### LINKED ISSUE #490: Call DataFrame methods from Accessor
Adds the ability to use DataFrame methods and return a dataframe with an initialized Schema by wrapping the DataFrame operation.
- Should be the last resort after checking if method exists on schema or accessor
- Needs to be able to handle the possible pandas outputs (df, series, other)
- Needs to be able to handle inplace operations as well

tamargrey:
- Allows for things like `df.ww.sample` while throwing errors for operation that invalidates the original DataFrame + Schema or a new DataFrame + Schema
- Closes #490

### DIFF
--- a/woodwork/exceptions.py
+++ b/woodwork/exceptions.py
@@ -31,4 +31,9 @@ class OutdatedSchemaWarning(UserWarning):
 
 class SchemaInvalidatedWarning(UserWarning):
     def get_warning_message(self, attr, invalid_reason):
-        return (f'Operation performed by {attr} has invalidated the Woodwork typing information:\n {invalid_reason}. Reinitialize the typing information with DataFrame.ww.init.')
+        return (f'Operation performed by {attr} has invalidated the Woodwork typing information:\n {invalid_reason}.\n Please reinitialize Woodwork with DataFrame.ww.init')
+
+
+class CannotInitSchemaWarning(UserWarning):
+    def get_warning_message(self, attr, invalid_reason):
+        return (f'DataFrame created by {attr} is not valid for the given typing information:\n {invalid_reason}.\n Please initialize Woodwork with DataFrame.ww.init')
--- a/woodwork/exceptions.py
+++ b/woodwork/exceptions.py
@@ -31,4 +31,4 @@ class OutdatedSchemaWarning(UserWarning):
 
 class SchemaInvalidatedWarning(UserWarning):
     def get_warning_message(self, attr, invalid_reason):
-        return (f'Operation performed by {attr} has invalidated the Woodwork typing information:\n {invalid_reason}')
+        return (f'Operation performed by {attr} has invalidated the Woodwork typing information:\n {invalid_reason}. Reinitialize the typing information with DataFrame.ww.init.')
--- a/woodwork/exceptions.py
+++ b/woodwork/exceptions.py
@@ -29,15 +29,8 @@ class OutdatedSchemaWarning(UserWarning):
                 % (saved_version_str))
 
 
-class SchemaInvalidatedWarning(UserWarning):
+class TypingInfoMismatchWarning(UserWarning):
     def get_warning_message(self, attr, invalid_reason):
         return (f'Operation performed by {attr} has invalidated the Woodwork typing information:\n '
-                f'{invalid_reason}.\n '
-                'Please reinitialize Woodwork with DataFrame.ww.init')
-
-
-class CannotInitSchemaWarning(UserWarning):
-    def get_warning_message(self, attr, invalid_reason):
-        return (f'DataFrame created by {attr} is not valid for the given typing information:\n '
                 f'{invalid_reason}.\n '
                 'Please initialize Woodwork with DataFrame.ww.init')
--- a/woodwork/exceptions.py
+++ b/woodwork/exceptions.py
@@ -27,3 +27,8 @@ class OutdatedSchemaWarning(UserWarning):
                 '%s is no longer supported by this version '
                 'of woodwork. Attempting to load woodwork.DataTable ...'
                 % (saved_version_str))
+
+
+class SchemaInvalidatedWarning(UserWarning):
+    def get_warning_message(self, attr):
+        return (f'Operation performed by {attr} has invalidated the Woodwork typing information.')
--- a/woodwork/exceptions.py
+++ b/woodwork/exceptions.py
@@ -31,9 +31,13 @@ class OutdatedSchemaWarning(UserWarning):
 
 class SchemaInvalidatedWarning(UserWarning):
     def get_warning_message(self, attr, invalid_reason):
-        return (f'Operation performed by {attr} has invalidated the Woodwork typing information:\n {invalid_reason}.\n Please reinitialize Woodwork with DataFrame.ww.init')
+        return (f'Operation performed by {attr} has invalidated the Woodwork typing information:\n '
+                f'{invalid_reason}.\n '
+                'Please reinitialize Woodwork with DataFrame.ww.init')
 
 
 class CannotInitSchemaWarning(UserWarning):
     def get_warning_message(self, attr, invalid_reason):
-        return (f'DataFrame created by {attr} is not valid for the given typing information:\n {invalid_reason}.\n Please initialize Woodwork with DataFrame.ww.init')
+        return (f'DataFrame created by {attr} is not valid for the given typing information:\n '
+                f'{invalid_reason}.\n '
+                'Please initialize Woodwork with DataFrame.ww.init')
--- a/woodwork/exceptions.py
+++ b/woodwork/exceptions.py
@@ -30,5 +30,5 @@ class OutdatedSchemaWarning(UserWarning):
 
 
 class SchemaInvalidatedWarning(UserWarning):
-    def get_warning_message(self, attr):
-        return (f'Operation performed by {attr} has invalidated the Woodwork typing information.')
+    def get_warning_message(self, attr, invalid_reason):
+        return (f'Operation performed by {attr} has invalidated the Woodwork typing information:\n {invalid_reason}')
--- a/woodwork/table_accessor.py
+++ b/woodwork/table_accessor.py
@@ -2,7 +2,7 @@ import warnings
 
 import pandas as pd
 
-from woodwork.exceptions import SchemaInvalidatedWarning
+from woodwork.exceptions import CannotInitSchemaWarning, SchemaInvalidatedWarning
 from woodwork.logical_types import Datetime, LatLong, Ordinal
 from woodwork.schema import Schema
 from woodwork.type_sys.utils import (
@@ -102,7 +102,6 @@ class WoodworkTableAccessor:
             If the method is present on Schema, uses that method.
             If the method is present on DataFrame, uses that method.
         '''
-        # schema = object.__getattribute__(self, '_schema')
         if self._schema is None:
             raise AttributeError("Woodwork not initialized for this DataFrame. Initialize by calling DataFrame.ww.init")
         if hasattr(self._schema, attr):
@@ -126,8 +125,10 @@ class WoodworkTableAccessor:
                     if isinstance(result, pd.DataFrame):
                         invalid_schema_message = _get_invalid_schema_message(result, self._schema)
                         if invalid_schema_message:
-                            raise ValueError(f'Cannot perform operation {attr}: {invalid_schema_message}')
-                        result.ww.init(schema=self._schema)
+                            warnings.warn(CannotInitSchemaWarning().get_warning_message(attr, invalid_schema_message),
+                                          CannotInitSchemaWarning)
+                        else:
+                            result.ww.init(schema=self._schema)
                     # Confirm that the Schema is still valid on original DataFrame
                     invalid_schema_message = _get_invalid_schema_message(self._dataframe, self._schema)
                     if invalid_schema_message:
@@ -309,9 +310,9 @@ def _get_invalid_schema_message(dataframe, schema):
         df_dtype = str(dataframe[name].dtype)
         schema_dtype = schema.logical_types[name].pandas_dtype
         if df_dtype != schema_dtype:
-            return f'dtype mismatch for column {name} between DataFrame dtype, {df_dtype}, and LogicalType dtype, {schema_dtype}.'
+            return f'dtype mismatch for column {name} between DataFrame dtype, {df_dtype}, and LogicalType dtype, {schema_dtype}'
     if schema.index is not None:
         if not all(dataframe.index == dataframe[schema.index]):
-            return 'Index mismatch between DataFrame and typing information.'
+            return 'Index mismatch between DataFrame and typing information'
         elif not dataframe[schema.index].is_unique:
-            return 'Index column is not unique.'
+            return 'Index column is not unique'
--- a/woodwork/table_accessor.py
+++ b/woodwork/table_accessor.py
@@ -2,10 +2,7 @@ import warnings
 
 import pandas as pd
 
-from woodwork.exceptions import (
-    CannotInitSchemaWarning,
-    SchemaInvalidatedWarning
-)
+from woodwork.exceptions import TypingInfoMismatchWarning
 from woodwork.logical_types import Datetime, LatLong, Ordinal
 from woodwork.schema import Schema
 from woodwork.type_sys.utils import (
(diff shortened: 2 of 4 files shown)


woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -844,3 +844,96 @@ def test_is_valid_schema(sample_df):
...
+    assert 'new_name' in schema_df.columns
+
+
+def test_dataframe_methods_on_accessor_returning_indexer(sample_df):
+    xfail_dask_and_koalas(sample_df)
+
+    schema_df = sample_df.copy()
+    schema_df.ww.init(name='test_schema')
+
+    iloc_df = schema_df.ww.iloc()[2:, :]
```
tamargrey, line 893: Want to highlight the behavior of indexers like `iloc`. For some reason, the behavior of the Accessor doesn't allow for `df.ww.iloc[2:, :]`. 

It throws `TypeError: 'function' object is not subscriptable`, probably because of some type transformation that happens when using a getattr like this:
image

My guess is that we might eventually want to add our own `iloc` if this behavior isn't great in the long run. Plus it'd be great to get `df.ww.iloc[:, 2]` to return a series with an initialized Accessor, which we'd definitely need to implement ourselves.

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -844,3 +844,96 @@ def test_is_valid_schema(sample_df):
...
+
+    schema_df.ww.sort_values(['full_name'], inplace=True)
+    assert schema_df.ww.name == 'test_schema'
+
+    pd.testing.assert_frame_equal(to_pandas(schema_df), to_pandas(df_pre_sort.sort_values(['full_name'])))
+
+    error = 'Woodwork typing information is not valid for this DataFrame.'
+    with pytest.raises(ValueError, match=error):
+        schema_df.ww.rename({'id': 'new_name'}, inplace=True, axis=1)
+    assert 'new_name' in schema_df.columns
```
tamargrey, line 884: Worth noting that even when the schema gets invalidated and we throw an error, it's happening after we've made the call, so in the case of inplace operations if you were to catch the error, you'd still have the action performed. I'm not sure how we'd avoid this, but maybe it's not a problem?

(can also see this in the call of `df.ww.pop` in `test_dataframe_methods_on_accessor_returning_series`)

thehomebrewnerd (changes requested):

woodwork/tests/accessor/test_table_accessor.py
tamargrey: Would you be ok with setting `df.ww.schema` to None any time the schema is invalidated? Is there a reason to keep it present at all (maybe to allow updating it so that it's valid again)?

When it comes to checking that the schema is valid before being able to access it, we can add the check at the top level of the `__getattr__`. That means that we'll need to explicitly check the validity of the schema in any Accessor methods. I do think it probably makes sense as a separate error from the `Woodwork not initialized for this DataFrame` one, so maybe that's reason enough to not just set the schema to None when it's invalidated. 

Would it be worth raising a warning when the schema gets invalidated? Trying to think about a situation in which you might perform a sequence of operations on a DataFrame and it'd be useful to know which one was causing the schema to be invalidated.

woodwork/table_accessor.py
```diff
@@ -207,7 +225,7 @@ def _check_schema(dataframe, schema):
     if not isinstance(schema, Schema):
         raise TypeError('Provided schema must be a Woodwork.Schema object.')
     if not _is_valid_schema(dataframe, schema):
-        raise ValueError('Invalid typing information presented at init. Cannot set Woodwork on DataFrame.')
+        raise ValueError('Woodwork typing information is not valid for this DataFrame.')
```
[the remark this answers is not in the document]
thehomebrewnerd: Yeah, I was thinking something along those lines so we could pass in a custom message depending on where the error happened.

woodwork/tests/accessor/test_table_accessor.py
thehomebrewnerd: So that's a good question. Maybe it would be best for now just to set the schema back to None if it is determined to be invalid. If we could provide a meaningful warning (even if it is somewhat generic) at that time we reset, that might be sufficient for now. We could always add more later if it becomes a bigger issue.

The more I think about it, setting the schema back to None seems cleaner than sprinkling a bunch of schema validation checks throughout the code.

woodwork/table_accessor.py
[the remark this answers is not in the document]
tamargrey: So if we want slightly different error messages based on when we check that the schema is valid
- before accessing the schema - `Cannot access logical_types; Woodwork typing information is invalid`
- when doing `df.ww.init(schema)` directly - `Invalid typing information presented at init`
- inside of the wrapper - `performing the replace operation has invalidated the typing information`

And each of those as a prefix to the specific message that should help with fixing the issue - something like `Column 'id' dtype 'string' does not match the listed Logical Type Categorical's dtype 'categorical'`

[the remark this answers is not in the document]
thehomebrewnerd: I think for the cases where we warn, instead of error, we might want to mention that the typing information can be reinitialized by calling df.ww.init(). Not real clear how this fits exactly with setting the schema to None on invalidation, but these messages do provide additional detail that would help the user understand what is going on.

woodwork/tests/accessor/test_table_accessor.py
tamargrey: Adde #539

woodwork/table_accessor.py
```diff
@@ -206,8 +265,9 @@ def _check_logical_types(dataframe_columns, logical_types):
 def _check_schema(dataframe, schema):
     if not isinstance(schema, Schema):
         raise TypeError('Provided schema must be a Woodwork.Schema object.')
-    if not _is_valid_schema(dataframe, schema):
-        raise ValueError('Invalid typing information presented at init. Cannot set Woodwork on DataFrame.')
+    invalid_schema_message = _get_invalid_schema_message(dataframe, schema)
+    if invalid_schema_message:
+        raise ValueError(f'Woodwork typing information is not valid for this DataFrame: {invalid_schema_message}')
```
tamargrey, line 270: @thehomebrewnerd The one place I've allowed to error for an invalid schema is when you directly initialize with a Schema (so calling `df.ww.init(schema)` directly instead of going through `df.ww.sample` or something).

The three situations where Woodwork will check for an invalid schema are as follows:

- Perform a pandas action that creates a new DataFrame for which the schema is invalid - WARNS
- Perform a pandas action that alters the DataFrame in a way that invalidates the schema - WARNS
- Directly initialize - ERRORS

My logic is that I think we want to be allowed to raise errors when users call `df.ww.init` directly. If we say that `df.ww.init(schema)` warns instead of errors, then I think it’d follow that we should also be warning users if they are setting individual parameters with `df.ww.init(logical_types={..}, index='id')`, but I think we definitely want to raise an Error if you try to set the index to an invalid column.

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -817,30 +821,109 @@ def test_accessor_repr(sample_df, sample_column_names, sample_inferred_logical_t
...
     xfail_dask_and_koalas(sample_df)
 
     schema_df = sample_df.copy()
     schema_df.ww.init(name='test_schema', index='id', logical_types={'id': 'Double', 'full_name': 'FullName'})
     schema = schema_df.ww.schema
 
-    assert _is_valid_schema(schema_df, schema)
-    assert not _is_valid_schema(sample_df, schema)
+    assert _get_invalid_schema_message(schema_df, schema) is None
+    assert _get_invalid_schema_message(sample_df, schema) == 'dtype mismatch for column id between DataFrame dtype, int64, and LogicalType dtype, float64'
```
thehomebrewnerd, line 832: Long line here and in the assertions that follow in this test.

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -817,30 +821,109 @@ def test_accessor_repr(sample_df, sample_column_names, sample_inferred_logical_t
...
+
+    copied_df = schema_df.ww.copy()
+
+    assert schema_df is not copied_df
+    assert isinstance(copied_df.ww.schema, Schema)
+    assert copied_df.ww.schema == schema_df.ww.schema
+
+    pd.testing.assert_frame_equal(to_pandas(schema_df), to_pandas(copied_df))
+
+    warning = 'DataFrame created by astype is not valid for the given typing information:\n dtype mismatch for column id between DataFrame dtype, string, and LogicalType dtype, Int64.\n Please initialize Woodwork with DataFrame.ww.init'
```
thehomebrewnerd, line 867: Long line

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -817,30 +821,109 @@ def test_accessor_repr(sample_df, sample_column_names, sample_inferred_logical_t
...
+    schema_df.ww.init(name='test_schema')
+
+    df_pre_sort = schema_df.copy()
+
+    schema_df.ww.sort_values(['full_name'], inplace=True)
+    assert schema_df.ww.name == 'test_schema'
+
+    pd.testing.assert_frame_equal(to_pandas(schema_df), to_pandas(df_pre_sort.sort_values(['full_name'])))
+
+    warning = "Operation performed by rename has invalidated the Woodwork typing information:\n The following columns in the DataFrame were missing from the typing information: {'new_name'}.\n Please reinitialize Woodwork with DataFrame.ww.init"
```
thehomebrewnerd, line 888: Long line

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -817,30 +821,109 @@ def test_accessor_repr(sample_df, sample_column_names, sample_inferred_logical_t
...
+    dtypes = schema_df.ww.dtypes
+
+    assert schema_df.ww.name == 'test_schema'
+    pd.testing.assert_series_equal(dtypes, schema_df.dtypes)
+
+    memory = schema_df.ww.memory_usage()
+    assert schema_df.ww.name == 'test_schema'
+    pd.testing.assert_series_equal(memory, schema_df.memory_usage())
+
+    warning = "Operation performed by pop has invalidated the Woodwork typing information:\n The following columns in the typing information were missing from the DataFrame: {'id'}.\n Please reinitialize Woodwork with DataFrame.ww.init"
```
thehomebrewnerd, line 910: Long line

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -817,30 +821,109 @@ def test_accessor_repr(sample_df, sample_column_names, sample_inferred_logical_t
...
+    schema_df.ww.init(name='test_schema')
+
+    shape = schema_df.ww.shape
+
+    assert schema_df.ww.name == 'test_schema'
+    assert shape == schema_df.shape
 
-    different_dtype_df = schema_df.astype({'id': 'Int64'}, copy=True)
-    assert not _is_valid_schema(different_dtype_df, schema)
+    keys = schema_df.ww.keys()
```
thehomebrewnerd, line 927: Move this immediately before the assertion where it is used, or update the assertion to use `schema_df.ww.keys()` directly to make this a bit more readable.

woodwork/exceptions.py
```diff
@@ -27,3 +27,17 @@ def get_warning_message(self, saved_version_str):
...
+        return (f'Operation performed by {attr} has invalidated the Woodwork typing information:\n '
+                f'{invalid_reason}.\n '
+                'Please reinitialize Woodwork with DataFrame.ww.init')
+
+
+class CannotInitSchemaWarning(UserWarning):
+    def get_warning_message(self, attr, invalid_reason):
+        return (f'DataFrame created by {attr} is not valid for the given typing information:\n '
+                f'{invalid_reason}.\n '
+                'Please initialize Woodwork with DataFrame.ww.init')
```
thehomebrewnerd, line 43: How would you feel about combining these into a single `SchemaMismatchWarning`. I see these are warning for slightly different situations, but the warnings are similar enough I wonder if we can get by with a single one?

thehomebrewnerd (changes requested):
Looking pretty good. Just a few minor things to clean up and a couple questions/suggestions to consider. This rendition was a lot easier for me to follow than the last one, so the code readability here has definitely improved!

woodwork/table_accessor.py
```diff
@@ -97,20 +110,14 @@ def __getattr__(self, attr):
...
-            # logical_type attr can return an uninstantiated LogicalType which we don't want to interpret as callable
-            if callable(schema_attr) and attr != 'logical_type':
-                def wrapper(*args, **kwargs):
-                    return schema_attr(*args, **kwargs)
-                return wrapper
-            else:
-                return schema_attr
+            return self._make_schema_call(attr)
+        if hasattr(self._dataframe, attr):
+            return self._make_dataframe_call(attr)
```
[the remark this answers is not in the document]
tamargrey: yay! Glad it helps clear things up

woodwork/table_accessor.py
```diff
@@ -145,6 +152,55 @@ def _set_underlying_index(self):
...
+                                      CannotInitSchemaWarning)
+                    else:
+                        result.ww.init(schema=self._schema)
+                # Confirm that the Schema is still valid on original DataFrame
+                # Important for inplace operations
+                invalid_schema_message = _get_invalid_schema_message(self._dataframe, self._schema)
+                if invalid_schema_message:
+                    warnings.warn(SchemaInvalidatedWarning().get_warning_message(attr, invalid_schema_message),
+                                  SchemaInvalidatedWarning)
+                    self._schema = None
```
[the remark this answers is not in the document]
tamargrey: I was thinking there might be a situation in which the original DataFrame gets altered even if a new DataFrame gets returned. I guess that'd be not great on pandas' part if they allowed unexpected side effects, but maybe sometimes it's expected?

woodwork/exceptions.py
tamargrey: I think if we say that the second Warning can exist in the else of `if isinstance(result, pd.DataFrame):`, then it's fine to have the Warnings be exactly the same.

But if, in theory, you can get both warnings at once (if there's a DataFrame operation that could both create a new DataFrame and alter the original one in a way that both had the schema invalidated), it'd be super confusing to get the same Warning twice.

But also I can change it so that you pass in whatever message you want to a generic `SchemaMismatchWarning`.

woodwork/table_accessor.py
```diff
@@ -266,15 +326,26 @@ def _update_column_dtype(series, logical_type, name):
...
+        return f'The following columns in the typing information were missing from the DataFrame: '\
+            f'{schema_cols_not_in_df}'
     for name in dataframe.columns:
-        if str(dataframe[name].dtype) != schema.logical_types[name].pandas_dtype:
-            return False
+        df_dtype = str(dataframe[name].dtype)
+        schema_dtype = schema.logical_types[name].pandas_dtype
+        if df_dtype != schema_dtype:
+            return f'dtype mismatch for column {name} between DataFrame dtype, '\
+                f'{df_dtype}, and LogicalType dtype, {schema_dtype}'
```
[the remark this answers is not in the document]
tamargrey: That makes a lot of sense! Changing

woodwork/tests/accessor/test_table_accessor.py
tamargrey: changing to not set the `keys` variable at all and just use it directly

woodwork/table_accessor.py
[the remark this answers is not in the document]
thehomebrewnerd: I was trying to think of that scenario to, but didn't come up with anything. I think either you get `None` for an inplace operation or a new dataframe, with the original unmodified, but that might not be guaranteed.

I would lean toward imbedding the second check in an else block for efficiency at this point until we identify a situation where it needs to be outside the else block.

woodwork/exceptions.py
thehomebrewnerd: Sure makes sense. I'm thinking we consolidate these for now until we find a situation that would require them to be separated.

woodwork/table_accessor.py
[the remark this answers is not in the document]
tamargrey: Sounds good. I'll do this and consolidate the Warnings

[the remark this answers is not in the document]
tamargrey: Also gonna not use the word `schema` in the Warning name as that'll get propagated to users who won't know what that refers to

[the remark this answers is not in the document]
thehomebrewnerd: Good call!

woodwork/table_accessor.py
```diff
@@ -145,6 +149,56 @@ def _set_underlying_index(self):
...
+        dataframe_attr = getattr(self._dataframe, attr)
+
+        if callable(dataframe_attr):
+            def wrapper(*args, **kwargs):
+                # Make DataFrame call and intercept the result
+                result = dataframe_attr(*args, **kwargs)
+
+                # Try to initialize Woodwork with the existing Schema
+                if isinstance(result, pd.DataFrame):
+                    invalid_schema_message = _get_invalid_schema_message(result, self._schema)
```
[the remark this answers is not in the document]
tamargrey: Yeah, that's a fair point. We could definitely define some DataFrame calls that will always keep the schema valid. Opened #545

gsheni (approved):
lgtm

Issue #541, opened 2021-02-02
Koalas option fails resulting in Exception


gsheni:
- An EvalML user reported this [error](https://github.com/alteryx/evalml/issues/1780), which is caused by Woodwork.
- We should look into making it so that if the `ks.set_option` cannot be safely called, `woodwork` import does not fail.
  - OR We should also look into if `set_option` can only be called when using a Koalas DataFrame with Woodwork (instead of being located at the top of a file and outside a function).

```python
~\Anaconda3\lib\site-packages\evalml\demos\breast_cancer.py in
1 import pandas as pd
----> 2 import woodwork as ww
3 from sklearn.datasets import load_breast_cancer as load_breast_cancer_sk
4
5

~\Anaconda3\lib\site-packages\woodwork_init_.py in
1 # flake8: noqa
2 from .config import config
----> 3 from .datatable import DataColumn, DataTable
4 from .type_sys import type_system
5 from .type_sys.utils import list_logical_types, list_semantic_tags

~\Anaconda3\lib\site-packages\woodwork\datatable.py in
27 ks = import_or_none('databricks.koalas')
28 if ks:
---> 29 ks.set_option('compute.ops_on_diff_frames', True)
```

dsherry:
@gsheni @thehomebrewnerd wooo, you guys are fast!

Between when I responded to https://github.com/alteryx/evalml/issues/1780 and now I worked up a draft PR which I believe will fix this problem. I'll get that ready for review. If you agree, let's merge it. If not, I will gladly cede the issue back to you @thehomebrewnerd  :)

dsherry:
@gsheni @thehomebrewnerd 

My PR #542 will prevent the error reported in https://github.com/alteryx/evalml/issues/1780 from causing the woodwork import to crash. I think its a helpful change because it'll a) prevent this issue from crashing import in woodwork and dependent projects (evalml) and b) give us more info if users encounter similar issues in the future. However, given the stack trace from https://github.com/alteryx/evalml/issues/1780, #542 is probably not enough to close this issue.

My suggestion: merge #542 and get that into evalml, keep this issue open and try to reproduce the error with `koalas.set_option` from https://github.com/alteryx/evalml/issues/1780

thehomebrewnerd:
[...]

PR #542, opened 2021-02-03
Avoid fatal errors at import-time from koalas

dsherry:
For #541 

This PR will prevent the call to `koalas.set_option` from crashing import.

This doesn't solve the root cause of #541, but it will hopefully prevent a crash at import-time resulting from that issue.

### DIFF
--- a/woodwork/utils.py
+++ b/woodwork/utils.py
@@ -1,7 +1,7 @@
-
 import ast
 import importlib
 import re
+import warnings
 
 import numpy as np
 import pandas as pd
@@ -21,6 +21,41 @@ def import_or_none(library):
         return importlib.import_module(library)
     except ImportError:
         return None
+
+
+def import_or_raise(library, error_msg):
+    '''
+    Attempts to import the requested library.  If the import fails, raises an
+    ImportError with the supplied error message.
+
+    Args:
+        library (str): the name of the library
+        error_msg (str): error message to return if the import fails
+    '''
+    try:
+        return importlib.import_module(library)
+    except ImportError:
+        raise ImportError(error_msg)
+
+
+def import_and_configure_koalas(compute_ops_on_diff_frames=True):
+    """
+    Import the koalas library. Set the compute.ops_on_diff_frames as an optimization.
+
+    Args:
+        compute_ops_on_diff_frames (bool): the koalas compute.ops_on_diff_frames argum,ent
+
+    Returns:
+        A reference to the koalas module
+    """
+    ks = import_or_none('databricks.koalas')
+    if ks is None:
+        return None
+    try:
+        ks.set_option('compute.ops_on_diff_frames', compute_ops_on_diff_frames)
+    except Exception as e:
+        warnings.warn('Unable to set compute.ops_on_diff_frames for koalas: ' + str(e))
+    return ks
 
 
 def camel_to_snake(s):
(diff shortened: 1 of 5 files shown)


woodwork/utils.py
```diff
@@ -23,6 +23,41 @@ def import_or_none(library):
         return None
 
 
+def import_or_raise(library, error_msg):
```
dsherry, line 26: I just moved this higher in the file, no code change

woodwork/utils.py
```diff
@@ -23,6 +23,41 @@ def import_or_none(library):
...
+        A reference to the koalas module
+    """
+    ks = import_or_none('databricks.koalas')
+    if ks is None:
+        return None
+    try:
+        ks.set_option('compute.ops_on_diff_frames', compute_ops_on_diff_frames)
+    except Exception as e:
+        warnings.warn('Unable to set compute.ops_on_diff_frames for koalas: ' + str(e))
+    return ks
```
dsherry, line 58: Here's the new logic I added. If the call to `ks.set_option` results in an exception, report it as a warning and continue with the import.

If there's an underlying problem with koalas here, this wouldn't fix that, but it would at least allow the import of woodwork to succeed and log a warning.

woodwork/utils.py
```diff
@@ -23,6 +23,41 @@ def import_or_none(library):
...
+        compute_ops_on_diff_frames (bool): the koalas compute.ops_on_diff_frames argum,ent
+
+    Returns:
+        A reference to the koalas module
+    """
+    ks = import_or_none('databricks.koalas')
+    if ks is None:
+        return None
+    try:
+        ks.set_option('compute.ops_on_diff_frames', compute_ops_on_diff_frames)
```
dsherry, line 55: Something which we could discuss separately is making this configurable.

I would categorize this "compute.ops_on_diff_frames" option as global state, meaning something users could set/unset independent of woodwork. Ideally, low-level libraries like woodwork and evalml would either a) not change global state or b) have an opinion on changing global state but give users a way to override.

My suggestion on how to do that would be to add a config for this option. I didn't do that in this PR as that's not the immediate issue, but that's come up in discussions with @boopboopbeepboop recently. Could be a good separate issue for woodwork. We have a similar issue to solve in evalml RE configuring the logger.

@rwedge FYI

thehomebrewnerd: Checked into this - we need to have this option set to be able to reassign new columns to a dataframe when a type change happens. In this case we create a new series with the appropriate type and replace the existing series with the new series. So without this option set, we can't do much with Woodwork and Koalas dataframes.

We can handle this more gracefully though, and avoid an error when the user isn't trying to use Koalas even though they might have it installed.

woodwork/utils.py
```diff
@@ -23,6 +23,41 @@ def import_or_none(library):
...
+        library (str): the name of the library
+        error_msg (str): error message to return if the import fails
+    '''
+    try:
+        return importlib.import_module(library)
+    except ImportError:
+        raise ImportError(error_msg)
+
+
+def import_and_configure_koalas(compute_ops_on_diff_frames=True):
```
tamargrey, line 41: It'd be cool if we could make this a generic `import_and configure_package(package, **args)`. Doesn't have to be in this PR, but maybe something to think about in the future.

thehomebrewnerd:
I've worked up an alternate solution to this issue in #543 

I think both have merits, will discuss further and decide which path to take.

woodwork/utils.py
dsherry: Totally agree! I think that work should happen if/when we add more "global config" options like this, at which point we should probably add a config object to encapsulate that stuff.

gsheni:
@dsherry thanks for your help and putting up a fix. We're going to go ahead with solution #543 because it avoids the unnecessary try/except.

PR #543, opened 2021-02-03
Fix issue with setting koalas option

### LINKED ISSUE #541: Koalas option fails resulting in Exception

thehomebrewnerd:
- Fix issue with setting koalas option
- Closes #541 

Update so that Kolas `'compute.ops_on_diff_frames'` is only set if user is working with Koalas dataframes, rather than setting in import.

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -9,6 +9,7 @@ Release Notes
     * Fixes
         * Handle missing values in Datetime columns when calculating mutual information (:pr:`516`)
         * Support numpy 1.20.0 by restricting version for koalas and changing serialization error message (:pr:`532`)
+        * Move Koalas option setting to DataTable init instead of import (:pr:`543`)
     * Changes
     * Documentation Changes
         * Add Alteryx OSS Twitter link (:pr:`519`)
--- a/woodwork/datatable.py
+++ b/woodwork/datatable.py
@@ -26,8 +26,6 @@ from woodwork.utils import (
 
 dd = import_or_none('dask.dataframe')
 ks = import_or_none('databricks.koalas')
-if ks:
-    ks.set_option('compute.ops_on_diff_frames', True)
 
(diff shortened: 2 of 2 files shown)


gsheni (approved):
lgtm

dsherry (commented):
Thanks @thehomebrewnerd . Closing #542 is fine with me.

If we're going to go this route, given that we have proof this call to `koalas.set_option` can cause a stack trace, I encourage us to add unit test coverage for this. Here's what I did in #542 : https://github.com/alteryx/woodwork/pull/542/files#diff-d975f76e08893e6ef08dbe15cf6637d26578f2933a2bdeef677d44498dfe0f98R305

thehomebrewnerd:
@dsherry In my opinion, we need to understand the root cause here before we take further action.

PR #544, opened 2021-02-03
Add Open Graph Info to Docs HTML

### LINKED ISSUE #529: Add Open Graph Info to Docs HTML
- Similar to the EvalML [issue](https://github.com/alteryx/evalml/issues/1722), we should add appropriate images for Woodwork.
- So that if someone puts a docs link on Twitter/Facebook, it shows a nice preview image with Woodwork and Alteryx logo

gsheni:
- Closes #529


thehomebrewnerd (approved):
I think this looks fine. If you wanted to, you could define the repeated content in variables like Featuretools does, so if we need to update in the future, it only needs to be changed in one place. Not critical though.

https://github.com/alteryx/featuretools/blob/c6951d513ab67284cd9435647b2d3ca74e910c42/docs/source/templates/layout.html#L12-L25

Issue #489, opened 2021-01-15
Add Woodwork methods that utilize the Schema


tamargrey:
Methods like `select` and `mutual_information` utilize the Schema's typing information and provide Woodwork-specific information, and they utilize the DataFrame.

TBD: Whether these functions live on the Accessor or Schema. Probably the Accessor, since we should only pass the DataFrame to the Schema on initialization.

`describe` and the serialization methods might fall under the definition of methods that are in the Pandas namespace, so maybe wait to add until we deal with methods like `rename`.

```python
    def select(self, include):
        '''Get a new DataFrame with Woodwork initialized with just the columns
        relevant in include.
        '''
        new_schema = self._datatable.select(include)
        col_names = new_schema.columns.keys()
        
        new_df = self._dataframe[col_names]
        new_df.ww.init(new_schema)
        
        return new_df
```

tamargrey:
probably also worth adding in the part of the getitem that gets a new df+schema from a list of column names. The single item -> series will be a part of the Series Accessor implementation.

PR #548, opened 2021-02-04
Add `select` to Accessor

tamargrey:
- Allows for selecting columns from your DataFrame based on the Woodwork typing info (logical type, semantic tags)
- Part of #489 

Adds in several helper functions that will be useful for future Woodwork methods
- Schema._filter_cols
- Schema._get_subset_schema
- WoodworkTableAccessor._get_subset_df_with_schema

Adds `use_standard_tags` attribute to the Schema so we can successfully get a subset Schema

### DIFF
--- a/woodwork/schema.py
+++ b/woodwork/schema.py
@@ -241,9 +241,9 @@ class Schema(object):
         new_semantic_tags = {}
         new_column_descriptions = {}
         new_column_metadata = {}
-        for col_name, col in self.columns.items():
-            if col_name not in subset_cols:
-                continue
+        for col_name in subset_cols:
+            col = col = self.columns[col_name]
+
             new_logical_types[col_name] = col['logical_type']
             new_semantic_tags[col_name] = col['semantic_tags']
             new_column_descriptions[col_name] = col['description']
--- a/woodwork/schema.py
+++ b/woodwork/schema.py
@@ -230,6 +230,12 @@ class Schema(object):
 
     def _new_schema_including(self, subset_cols):
         '''
+        Creates a new Schema with specified columns, retainig typing information.
+
+        Args:
+            subset_cols (list[str]): subset of columns from which to create the new Schema
+        Returns:
+            Schema: New Schema with attributes from original Schema
         '''
         new_logical_types = {}
         new_semantic_tags = {}
--- a/woodwork/schema.py
+++ b/woodwork/schema.py
@@ -228,7 +228,7 @@ class Schema(object):
 
         return list(cols_to_include)
 
-    def _new_schema_including(self, subset_cols):
+    def _get_subset_schema(self, subset_cols):
         '''
         Creates a new Schema with specified columns, retainig typing information.
 
--- a/woodwork/schema.py
+++ b/woodwork/schema.py
@@ -47,6 +47,7 @@ class Schema(object):
                          table_metadata, column_metadata, semantic_tags, column_descriptions)
 
         self.name = name
+        self.use_standard_tags = use_standard_tags
 
         # Infer logical types and create columns
         self.columns = self._create_columns(column_names,
@@ -225,8 +226,40 @@ class Schema(object):
             if _get_ltype_class(col['logical_type']) in ltypes_used or col['semantic_tags'].intersection(tags_used):
                 cols_to_include.add(col_name)
 
-        # --> need to remember to sort by dataframe column order any time we use this with a DataFrame
         return list(cols_to_include)
+
+    def _new_schema_including(self, subset_cols):
+        '''
+        '''
+        new_logical_types = {}
+        new_semantic_tags = {}
+        new_column_descriptions = {}
+        new_column_metadata = {}
+        for col_name, col in self.columns.items():
+            if col_name not in subset_cols:
+                continue
+            new_logical_types[col_name] = col['logical_type']
+            new_semantic_tags[col_name] = col['semantic_tags']
+            new_column_descriptions[col_name] = col['description']
+            new_column_metadata[col_name] = col['metadata']
+
+        new_index = self.index if self.index in subset_cols else None
+        new_time_index = self.time_index if self.time_index in subset_cols else None
+        if new_index is not None:
+            new_semantic_tags[new_index] = new_semantic_tags[new_index].difference({'index'})
+        if new_time_index is not None:
+            new_semantic_tags[new_time_index] = new_semantic_tags[new_time_index].difference({'time_index'})
+
+        return Schema(subset_cols,
+                      new_logical_types,
+                      name=self.name,
+                      index=new_index,
+                      time_index=new_time_index,
+                      semantic_tags=new_semantic_tags,
+                      use_standard_tags=self.use_standard_tags,
+                      table_metadata=self.metadata,
+                      column_metadata=new_column_metadata,
+                      column_descriptions=new_column_descriptions)
 
 
 def _validate_params(column_names, name, index, time_index, logical_types,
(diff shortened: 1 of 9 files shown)


woodwork/schema.py
```diff
@@ -181,6 +182,91 @@ def _set_index_tags(self, index):
     def _set_time_index_tags(self, time_index):
         self.columns[time_index]['semantic_tags'].add('time_index')
 
+    def _filter_cols(self, include):
+        """Return list of columns filtered in specified way. In case of collision, favors logical types
+        then semantic tag then column name.
```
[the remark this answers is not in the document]
thehomebrewnerd: That can also be handled in a separate issue/PR if we want to change.

woodwork/schema.py
```diff
@@ -181,6 +182,91 @@ def _set_index_tags(self, index):
...
+
+        Args:
+            include (str or LogicalType or list[str or LogicalType]): parameter or list of parameters to
+                filter columns by. Can be Logical Types or Semantic Tags.
```
thehomebrewnerd, line 191: Also column names?

woodwork/table_accessor.py
```diff
@@ -125,6 +125,25 @@ def __repr__(self):
...
+    def select(self, include):
+        """Create a DataFrame with Woodowork typing information initialized
+        that includes only columns whose Logical Type and semantic tags are
+        specified in the list of types and tags to include.
+        If no matching columns are found, an empty DataFrame will be returned.
+
+        Args:
+            include (str or LogicalType or list[str or LogicalType]): Logical
+                types, semantic tags to include
+                in the DataFrame.
```
thehomebrewnerd, line 137: Move end of sentence to line above.

woodwork/schema.py
```diff
@@ -181,6 +182,91 @@ def _set_index_tags(self, index):
...
+
+        for col_name, col in self.columns.items():
+            if _get_ltype_class(col['logical_type']) in ltypes_used or col['semantic_tags'].intersection(tags_used):
+                cols_to_include.add(col_name)
+
+        return list(cols_to_include)
+
+    def _get_subset_schema(self, subset_cols):
+        '''
+        Creates a new Schema with specified columns, retainig typing information.
```
thehomebrewnerd, line 233: retainig -> retaining

woodwork/schema.py
```diff
@@ -181,6 +182,91 @@ def _set_index_tags(self, index):
...
+        new_semantic_tags = {}
+        new_column_descriptions = {}
+        new_column_metadata = {}
+        for col_name, col in self.columns.items():
+            if col_name not in subset_cols:
+                continue
+            new_logical_types[col_name] = col['logical_type']
+            new_semantic_tags[col_name] = col['semantic_tags']
+            new_column_descriptions[col_name] = col['description']
+            new_column_metadata[col_name] = col['metadata']
```
thehomebrewnerd, line 250: Can we flip this loop around to loop over subset_cols instead of self.columns.items()? If we had a very wide schema - say 100 columns and we only wanted two of them, we would loop over 98 columns we don't care about. Will this work instead so we only loop over the columns we know we care about?
```python
        for col_name in subset_cols:
            col = self.columns[col_name]
            new_logical_types[col_name] = col['logical_type']
            new_semantic_tags[col_name] = col['semantic_tags']
            new_column_descriptions[col_name] = col['description']
            new_column_metadata[col_name] = col['metadata']
```

woodwork/tests/schema/test_schema.py
```diff
@@ -162,3 +167,93 @@ def test_schema_column_metadata(sample_column_names, sample_inferred_logical_typ
...
+    filtered_log_type = schema._filter_cols(include=NaturalLanguage)
+    assert filtered_log_type == filtered_log_type_string
+
+    filtered_semantic_tag = schema._filter_cols(include='numeric')
+    assert filtered_semantic_tag == ['age']
+
+    filtered_multiple = schema._filter_cols(include=['numeric'])
+    expected = ['phone_number', 'age']
+    for col in filtered_multiple:
+        assert col in expected
```
thehomebrewnerd, line 191: Could also make expected a set and use `assert set(filtered_multiple) == expected` instead of looping.

thehomebrewnerd (changes requested):
Looks pretty good. One major question -- Does/should `select` also work with column names? It is mentioned in one location, but I don't think the code will work with column names currently, unless I missed something. If it doesn't work, but we want it to, we should remove the docstring ref for now and create a new issue to update.

tamargrey:
@thehomebrewnerd Regarding column names, found [this comment](https://github.com/alteryx/woodwork/pull/228/files#r504075869) from the PR where that option was added to `DataTable._filter_cols`.

So we don't want to have the `col_names` parameter just yet. I think we'll want to add it back in when we implement `describe` which has an `include` parameter that can take all three of column names, logical types, and semantic tags

woodwork/tests/schema/test_schema.py
tamargrey: Another problem with this test: `phone_number` isn't even in `filtered_multiple`, so I'm not sure what this is even testing, since this call to `_filter_cols` doesn't even produce multiple columns.

Think I'm going to add a line where we check that `schema._filter_cols(include=NaturalLanguage)` matches an expected set since that will produce multiple columns.

woodwork/schema.py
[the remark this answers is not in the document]
tamargrey: Definitely still a relevant question for the future. Not sure why we went with that order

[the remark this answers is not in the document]
tamargrey: but going to remove the reference to column names for now

woodwork/schema.py
tamargrey: that makes a lot of sense! Changing

woodwork/tests/schema/test_schema.py
```diff
@@ -162,3 +167,90 @@ def test_schema_column_metadata(sample_column_names, sample_inferred_logical_typ
...
+        'index': 'id',
+        'time_index': 'signup_date',
+        'semantic_tags': {'age': 'test_tag'},
+        'table_metadata': {'created_by': 'user1'},
+        'column_metadata': {'phone_number': {'format': 'xxx-xxx-xxxx'}},
+        'use_standard_tags': False,
+        'column_descriptions': {'age': 'this is a description'}
+    }
+
+    # Confirm all possible params to DataTable init are present with non-default values where possible
```
gsheni, line 250: Wrong comment? Says DataTable

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -934,3 +936,302 @@ def test_dataframe_methods_on_accessor_other_returns(sample_df):
...
+                      }, semantic_tags={
+                          'full_name': ['new_tag', 'tag2'],
+                          'age': 'numeric',
+                          'signup_date': 'date_of_birth',
+                          'email': 'tag2'
+                      })
+
+    assert len(schema_df.ww.select(['doesnt_exist']).columns) == 0
+
+    df_multiple_unused = schema_df.ww.select(['doesnt_exist', 'boolean', 'category', PhoneNumber])
```
gsheni, line 1233: One test I didn't see is to make sure duplicate items in `select` doesn't break Schema.
```
schema_df.ww.select(['full_name', FullName, 'FullName'])
```

gsheni (commented):
looks good overall. Just 1 test case suggestion.

woodwork/tests/accessor/test_table_accessor.py
tamargrey: makes sense--will add

gsheni (approved):
lgtm

PR #552, opened 2021-02-05
Update branch reference in tests workflow

gsheni:
- Update branch reference in tests workflow
- Trigger tests on pull request events
- Change release branches to be named `release_0_0_9` to on not conflict with tags (named `v0.0.9`)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,4 +1,6 @@
-on: push
+on:
+  pull_request:
+    types: [opened, synchronize]
 
 name: Tests
 jobs:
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -62,43 +70,22 @@ jobs:
           from os.path import basename
           from re import compile
           main = '^main$'
-          release = '^v\d+\.\d+\.\d+$'
+          release_starts_r = 'release_\d+\_\d+\_\d+$'
```
gsheni, line 73: We want release branches to be named `release_0_0_9` so that they don't conflict with tags (ReadTheDocs gets confused).

.github/workflows/tests.yml
```diff
@@ -62,43 +70,21 @@ jobs:
           from os.path import basename
           from re import compile
           main = '^main$'
-          release = '^v\d+\.\d+\.\d+$'
+          release = 'release_\d+\_\d+\_\d+$'
```
gsheni, line 73: We want release branches to be named release_0_0_9 so that they don't conflict with tags (ReadTheDocs gets confused).

jeff-hernandez: If you add a `^` at the beginning of the pattern, it'll make sure the branch starts with `release`.

jeff-hernandez (approved):

.github/workflows/tests.yml
gsheni: Added

PR #563, opened 2021-02-08
Avoid calculating mutual information on unique columns

### LINKED ISSUE #555: Prevent Mutual Info from being calculated on unique columns
- If a column is unique, mutual info should not be calculated for that column.

tamargrey:
-  Closes #555 
- Had to update some of the existing tests to not be unique

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -77,6 +77,12 @@ jobs:
           ref = basename("${{ github.event.ref }}")
           is_dev = not any(pattern.match(ref) for pattern in patterns)
           print('::set-output name=is_dev::' + str(is_dev))
+      - if: ${{ steps.branch.outputs.is_dev == 'True' }}	
+        name: Checkout repository	
+        uses: actions/checkout@v2	
+      - if: ${{ steps.branch.outputs.is_dev == 'True' }}	
+        name: Check if release notes were updated	
+        run: cat docs/source/release_notes.rst | grep ":pr:\`${{ steps.pull_request.outputs.number }}\`"
 
   unit_tests:
     name: ${{ matrix.python_version }} unit tests ${{ matrix.optional_libraries }}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -77,13 +77,15 @@ jobs:
           ref = basename("${{ github.event.ref }}")
           is_dev = not any(pattern.match(ref) for pattern in patterns)
           print('::set-output name=is_dev::' + str(is_dev))
-      - if: ${{ steps.branch.outputs.is_dev == 'True' }}	
-        name: Checkout repository	
-        uses: actions/checkout@v2	
-      - if: ${{ steps.branch.outputs.is_dev == 'True' }}	
-        name: Check if release notes were updated	
-        run: cat docs/source/release_notes.rst | grep ":pr:\`${{ steps.pull_request.outputs.number }}\`"
-
+      - if: ${{ steps.branch.outputs.is_dev == 'True' }}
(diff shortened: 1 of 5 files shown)


.github/workflows/tests.yml
```diff
@@ -77,6 +77,12 @@ jobs:
           ref = basename("${{ github.event.ref }}")
           is_dev = not any(pattern.match(ref) for pattern in patterns)
           print('::set-output name=is_dev::' + str(is_dev))
+      - if: ${{ steps.branch.outputs.is_dev == 'True' }}	
+        name: Checkout repository	
+        uses: actions/checkout@v2	
+      - if: ${{ steps.branch.outputs.is_dev == 'True' }}	
+        name: Check if release notes were updated	
+        run: cat docs/source/release_notes.rst | grep ":pr:\`${{ steps.pull_request.outputs.number }}\`"
```
tamargrey, line 85: Added these lines back in to get the release note update CI check working again

.github/workflows/tests.yml
```diff
@@ -77,7 +77,15 @@ jobs:
...
-
+      - if: ${{ steps.branch.outputs.is_dev == 'True' }}
+        name: Checkout repository
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+      - if: ${{ steps.branch.outputs.is_dev == 'True' }}
+        name: Check if release notes were updated
+        run: cat docs/source/release_notes.rst | grep ":pr:\`${{ github.event.number }}\`"
```
tamargrey, line 88: Added these lines back in to get the release note update CI check working again

woodwork/datatable.py
```diff
@@ -1023,6 +1023,9 @@ def mutual_information_dict(self, num_bins=10, nrows=None):
         if ks and isinstance(self._dataframe, ks.DataFrame):
             data = data.to_pandas()
 
+        # Dpn't calculate mutual info for columns that aren't unique
+        data = data[[col for col in data.columns if not data[col].is_unique]]
```
gsheni, line 1027: Can we move the following line out `_handle_nans_for_mutual_info`, and do the following:
```python
        if nrows is not None and nrows < data.shape[0]:
            data = data.sample(nrows)

        # Remove fully null columns
        data = data.loc[:, data.columns[data.notnull().any()]]
        # Remove columns that are unique
        data = data[[col for col in data.columns if not data[col].is_unique]]

        data = self._handle_nans_for_mutual_info(data)
        data = self._make_categorical_for_mutual_info(data, num_bins)
```
- My thought is sample will do the most with large datasets to reduce the size. 
  - After that columns that are all NaN can be removed, and then any that are unique. (The unique column could make for a large computation...)
- Maybe we rename `_handle_nans_for_mutual_info` to `_replace_nans_for_mutual_info`?

tamargrey: Yeah, I like this. It also probably makes the behavior of mutual info/what it actually does to the data prior to calculating more clear

tamargrey: Question: is it faster to only create the new DataFrames from removing fully null columns and unique columns if there are any columns to remove? 

Right now, if there's nothing to remove we're needlessly copying the DataFrame

gsheni: @tamargrey good point. Yes, we should only do this if there is fully NULL columns and/or unique columns.

tamargrey: ok, updated to only do this when necessary

gsheni (approved):
lgtm

PR #564, opened 2021-02-09
Fix text on back arrow on install page

gsheni:
- Fixed text on back arrow on install page


tamargrey (approved):
lgtm

gsheni:
Public Docs Preview:
- https://woodwork.alteryx.com/en/fix_install_docs/

PR #571, opened 2021-02-09
Add mutual information to the Table Accessor

### LINKED ISSUE #558: Implement df.ww.mutual_information
- Should include the check to not calculate mutual info on unique columns

tamargrey:
- Adds ability to calculate mutual information for the DataFrame
- Closes #558

### DIFF
--- a/woodwork/table_accessor.py
+++ b/woodwork/table_accessor.py
@@ -298,7 +298,7 @@ class WoodworkTableAccessor:
 
     def mutual_information_dict(self, num_bins=10, nrows=None):
         """
-        Calculates mutual information between all pairs of columns in the DataTable that
+        Calculates mutual information between all pairs of columns in the DataFrame that
         support mutual information. Logical Types that support mutual information are as
         follows:  Boolean, Categorical, CountryCode, Datetime, Double, Integer, Ordinal,
         SubRegionCode, and ZIPCode
@@ -307,7 +307,7 @@ class WoodworkTableAccessor:
             num_bins (int): Determines number of bins to use for converting
                 numeric features into categorical.
             nrows (int): The number of rows to sample for when determining mutual info.
-                If specified, samples the desired number of rows from the data.
+            If specified, samples the desired number of rows from the data.
                 Defaults to using all rows.
 
         Returns:
@@ -361,7 +361,7 @@ class WoodworkTableAccessor:
 
     def mutual_information(self, num_bins=10, nrows=None):
         """
-        Calculates mutual information between all pairs of columns in the DataTable that
+        Calculates mutual information between all pairs of columns in the DataFrame that
         support mutual information. Logical Types that support mutual information are as
         follows:  Boolean, Categorical, CountryCode, Datetime, Double, Integer, Ordinal,
         SubRegionCode, and ZIPCode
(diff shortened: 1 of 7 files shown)


woodwork/table_accessor.py
```diff
@@ -232,6 +241,147 @@ def _get_subset_df_with_schema(self, cols_to_include):
...
+        """Transforms dataframe columns into numeric categories so that
+        mutual information can be calculated
+
+        Args:
+            data (pd.DataFrame): dataframe to use for caculating mutual information
+            num_bins (int): Determines number of bins to use for converting
+                numeric features into categorical.
+
+        Returns:
+            data (pd.DataFrame): Transformed data
```
thehomebrewnerd, line 280: We should update this to match the style used in `_replace_nans_for_mutual_info` -> `pd.DataFrame: Transformed data`?

woodwork/table_accessor.py
```diff
@@ -232,6 +241,147 @@ def _get_subset_df_with_schema(self, cols_to_include):
...
+        """
+
+        col_names = data.columns.to_list()
```
thehomebrewnerd, line 283: Do we need this list conversion, or can we just loop over data.columns in the next line -> `for col_name in data.columns:`?

thehomebrewnerd (commented):
Overall looks pretty good.

Topic for another day/PR: Should we move the methods `mutual_information_dict`, `_make_categorical_fo_mutual_info` and `_replace_nans_for_mutual_info` to a separate utils file? Not sure these need to reside on the accessor, but we would just have to update them to accept the column info as an additional parameter.

woodwork/table_accessor.py
tamargrey: Yeah, that makes sense to me

woodwork/table_accessor.py
tamargrey: Doesn't seem like we use the list of column names anywhere, so happy to make this change

woodwork/utils.py
```diff
@@ -136,7 +136,7 @@ def read_csv_to_accessor(filepath=None,
             information on available keywords refer to the pandas documentation.
 
     Returns:
-        woodwork.DataTable: DataTable created from the specified CSV file
+        pd.DataFrame: DataFrame created from the specified CSV file with Woodwork typing information initialized.
```
tamargrey, line 139: realized this docstring was still referencing DataTables so I changed it

tamargrey:
@thehomebrewnerd regarding moving mutual_information_dict, _make_categorical_fo_mutual_info and _replace_nans_for_mutual_info into a separate utils file, I think that's not a bad idea. Could be `statistics/utils` or something to go along with the `test_statitstics` file. 

If we're happy with that, I can start with `describe` when working on that issue, and move the mutual info stuff after.

thehomebrewnerd:
> @thehomebrewnerd regarding moving mutual_information_dict, _make_categorical_fo_mutual_info and _replace_nans_for_mutual_info into a separate utils file, I think that's not a bad idea. Could be `statistics/utils` or something to go along with the `test_statitstics` file.
> 
> If we're happy with that, I can start with `describe` when working on that issue, and move the mutual info stuff after.

Sounds good to me.

thehomebrewnerd (approved):
Looks good!

PR #573, opened 2021-02-10
Add tag modification methods to series accessor

### LINKED ISSUE #565: Implement methods to modify series semantic tags
The initial implementation of the series accessor does not include methods for adding, updating, removing or resetting semantic tags. These methods need to be implemented. The corresponding tests in `test_series_accessor.py` should be uncommented as well.

thehomebrewnerd:
- Add tag modification methods to series accessor
- Closes #565 

Adds the following methods to the series accessor, along with corresponding tests:
- `add_semantic_tags`
- `remove_semantic_tags`
- `reset_semantic_tags`
- `set_semantic_tags`

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -4,6 +4,7 @@ Release Notes
 -------------
 **Future Release**
     * Enhancements
+        * Add semantic tag update methods to column accessor (:pr:`573`)
     * Fixes
     * Changes
         * Avoid calculating mutualinfo for non-unique columns (:pr:`563`)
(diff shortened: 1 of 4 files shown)


tamargrey:
I think when we get to allowing set, reset, remove, add semantic tags on the Table Accessor, we should be able to use the bulk of the logic in the Column Accessor's methods. 

Not sure if it's better to do it now or later, but each of those methods could probably be a helper function in `schema_column.py` that takes in a column's current semantic tags and the user input tags to add or remove and returns the updated set of tags.

thehomebrewnerd:
> Not sure if it's better to do it now or later, but each of those methods could probably be a helper function in `schema_column.py` that takes in a column's current semantic tags and the user input tags to add or remove and returns the updated set of tags.

@tamargrey Yes, I think this is a good point. I generally tend to be of the opinion that we should refactor at the time the refactor becomes necessary. In this case we know we will need this, so it doesn't matter much whether we do it now or later. I can go ahead and split these out into helper functions, which will also keep the series accessor code more compact.

tamargrey (approved):
For now, I think we should allow inplace schema changes. We can change it to create a new object at some point if we run into a compelling reason.

gsheni (approved):
looks good!

PR #579, opened 2021-02-11
Add describe to the Accessor

### LINKED ISSUE #559: Implement df.ww.describe
- Adds back the col_name param to `_filter_cols`
- Overlaps in the same name space as `DataFrame.describe` so add check that it's as expected

tamargrey:
- Creates a `statistics_utils.py` file with `_get_describe_dict` that the Accessor calls for `df.ww.describe` and `df.ww.describe_dict`
- Closes #559

### DIFF
--- a/woodwork/schema.py
+++ b/woodwork/schema.py
@@ -209,28 +209,30 @@ class Schema(object):
         cols_to_include = set()
 
         for selector in include:
-            maybe_ltype = _get_ltype_class(selector)
-            if maybe_ltype in ww.logical_types.LogicalType.__subclasses__():
-                if maybe_ltype in ww.type_system.registered_types:
-                    if selector not in ww.type_system.registered_types:
-                        raise TypeError(f"Invalid selector used in include: {selector} cannot be instantiated")
-                    if selector in ltypes_in_schema:
-                        ltypes_used.add(selector)
-                else:
-                    raise TypeError(f"Specified LogicalType selector {selector} is not registered in Woodwork's type system.")
-            elif isinstance(selector, str):
-                # String can be interpreted as a LogicalType
-                ltype = ww.type_system.str_to_logical_type(selector, raise_error=False)
-                if ltype and ltype in ltypes_in_schema:
-                    ltypes_used.add(ltype)
-                # String can be interpreted as a semantic tag
-                if selector in tags_in_schema:
-                    tags_used.add(selector)
+            # Determine if the selector is a registered, uninstantiated LogicalType
+            maybe_ltype = selector
+            if isinstance(selector, str):
+                # Convert possible string to LogicalType - unregistered LogicalTypes return None
+                maybe_ltype = ww.type_system.str_to_logical_type(selector, raise_error=False)
+            # Get the class - unregistered LogicalTypes return LogicalTypeMetaClass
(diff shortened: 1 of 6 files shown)


woodwork/schema.py
```diff
@@ -219,8 +221,15 @@ def _filter_cols(self, include):
...
+                elif col_names and selector in self.columns:
+                    cols_to_include.add(selector)
             else:
-                raise TypeError(f"Invalid selector used in include: {selector} must be either a string or LogicalType")
+                # Non-string column names are relevant if using col_names
+                if col_names and selector in self.columns:
+                    cols_to_include.add(selector)
+                else:
+                    raise TypeError(f"Invalid selector used in include: {selector} must be a "
+                                    "string, LogicalType, or valid column name")
```
tamargrey, line 232: Want to highlight that `_filter_cols` was previously not allowing non-string column names to be filtered for. Changed that as well as the error language here.

woodwork/schema.py
```diff
@@ -219,8 +221,15 @@ def _filter_cols(self, include):
                     continue
                 elif selector in tags_in_schema:
                     tags_used.add(selector)
+                elif col_names and selector in self.columns:
+                    cols_to_include.add(selector)
```
tamargrey, line 225: @thehomebrewnerd Thought about why column order would be last in ltype > semantic tag > column name preference, and I think it's because there can only ever be one column with that column name, but there can many columns with a certain ltype or tag.

That said, it's not ideal that, if you have a column named `full_name` but its logical type is `NaturalLanguage`, it'll never get selected for even if you put `full_name` in the include list. But if we have it come first, then any column's with that logical type but not that column name will not get included.

My suggestion is that if a selector is a column name, it gets included. And then after that, we go and check logical types and semantic tags.

thehomebrewnerd: Wondering if the root of the issue is that we have a method that isn't doing just one thing, it's potentially doing multiple things depending on the input? Not suggesting we do this now, but maybe we need to revisit this and refactor at some point so this behavior is separated more cleanly somehow.

Maybe for now this should be an AND type selection, such that if there were three columns, one with the name `full_name`, one with the LogicalType `FullName` and one with the semantic tag of `full_name` all three would get selected with a `full_name` selector. Then the burden is on the user to avoid name conflicts in their data, types and tags, and we don't have to make the decision about what should be returned in the case of collisions.

tamargrey: Yeah, it reminds me how with `DataTable.set_types` we specifically have separate logical types and semantic tags parameters in order to not blur the lines between semantic tags and logical types.

For now, I think having an AND selection makes a lot of sense.

tamargrey: @thehomebrewnerd Changed `_filter_cols` to add the selector if it matches any of logical type, column name, or semantic tag.

I'm not feeling great about how that code is looking at the moment, though, since it feels overly complicated/hard to understand. I'm going to spend some more time thinking about ways to simplify, but open to any suggestions!

thehomebrewnerd (approved):
Seems like we should be able to clean up `_filter_cols`, but I'm not coming up with any specific suggestions at this point, so I think this is good!

As we discussed, let's be sure to create an issue to look at refactoring the way the `include` parameter is handled at some point in the future.

Issue #568, opened 2021-02-09
Updates for proper handling of LatLong columns


thehomebrewnerd:
In the previous Woodwork implementation using DataTables and DataColumns, any columns that were designates as LatLong columns would be reformatted and properly stored in a standard format. We need to implement this functionality in the new accessor approach as well. 

Note that reformatting of a standalone series cannot be done in place, and may require the use of the helper function created in #554.

tyler3991:
@thehomebrewnerd , putting this to top of dev backlog. Let me know if it should go somewhere else.

thehomebrewnerd:
@tyler3991 This is something we will need for feature parity with Woodwork prior to the accessor implementation, but I feel like this is lower priority that many of the other issues, since it only impacts one very specific data type, which isn't as common as most others.

tyler3991:
Ok, got it. Thanks @thehomebrewnerd.

thehomebrewnerd:
This code can be adapted for this conversion:
```
    if _get_ltype_class(self.logical_type) == LatLong:
        # Reformat LatLong columns to be a length two tuple (or list for Koalas) of floats
        if dd and isinstance(self._series, dd.Series):
            name = self._series.name
            meta = (self._series, tuple([float, float]))
            self._series = self._series.apply(_reformat_to_latlong, meta=meta)
            self._series.name = name
        elif ks and isinstance(self._series, ks.Series):
            formatted_series = self._series.to_pandas().apply(_reformat_to_latlong, use_list=True)
            self._series = ks.from_pandas(formatted_series)
        else:
            self._series = self._series.apply(_reformat_to_latlong)
```

thehomebrewnerd:
Closed by #598

Issue #515, opened 2021-01-26
Repetitive Validation checks among Schemas and Accessors


tamargrey:
Since WoodworkTableAccessor and Schema are separate objects and a Schema can be created independently of an Accessor, we have two sets of checks in each object that overlap some.

This will probably also be the case in the Series Accessor when that's created, as it will have validation checks of its own that will overlap with the Table Accessor and the column dictionary.

We can consider refactoring the validation in a way that doesn't re-validate unnecessary checks.

thehomebrewnerd:
Ordinal validation is currently happening in both `WoodworkColumnAccessor._validate_logical_type` and in `accessor_utils.py::_update_column_dtype`.

PR #581, opened 2021-02-11
Add util function for initializing series with dtype change

### LINKED ISSUE #554: Add helper function for initializing series with new accessor method
Because it is not possible to change a series `dtype` in place, we need a way to initialize a series with a logical type that will cause a series dtype change -- for example,  initializing a series of dtype `object` with a logical type of `Categorical`. Because we will need to change the dtype, we cannot do this via the normal series accessor initialization process with `series.ww.init()`.

This can be solved by creating a helper function that creates a new series object with the proper dtype, before initializing Woodwork, and then returning this new, initialized series to the user.

#### Code Example

```python
series = pd.Series(['a', 'b', 'c', 'a'], dtype='object')
ww_series = ww.init_series(series, logical_type='categorical')
assert ww_series.ww.schema is not None
assert ww_series.ww.logical_type is Categorical
assert ww_series.dtype == 'category'
assert series.dtype == 'object'
```

We should also update the error message in the `_validate_logical_type` method of the column accessor to suggest to users to use this function if they need to convert dtypes.

thehomebrewnerd:
- Add util function for initializing series with dtype change
- Closes #554

This PR implements a util function, `ww.init_series` that can be used to initialize a series with Woodwork typing information on a series, when a dtype conversion is required on the series. This function takes a series as an input along with the other initialization inputs (logical_type, semantic_tags, metadata, etc) and returns a new series with the dtype converted to match the logical type (if possible), initialized with Woodwork typing info.


woodwork/tests/accessor/test_acessor_utils.py
```diff
@@ -0,0 +1,38 @@
+import pytest
```
tamargrey, line 1: typo in the filename - should be `test_accessor_utils`

thehomebrewnerd: Fixed!

woodwork/accessor_utils.py
```diff
@@ -0,0 +1,74 @@
+import pandas as pd
+
+from woodwork.logical_types import Datetime
+from woodwork.type_sys.utils import _get_ltype_class
+from woodwork.utils import _get_column_logical_type
+
+
+def _update_dtype(series, logical_type):
```
tamargrey, line 8: Would it be possible to take `_update_column_dtype` from `table_accessor.py` and also use it here? I get that we're pulling out the LatLong logic to handle in #568, but I'm not sure I understand what needs to change. 

Maybe we'd remove the Ordinal data validation since that doesn't need to happen in this case (and maybe the data validation doesn't belong in the function that updates the series to match the dtype??), but it seems to me like it'd work.

thehomebrewnerd: I think we can use a singular function for both. I pulled this code from the original DataColumn code, not realizing it was already pulled out into the `table_accessor.py` file. I will clean that up so they are all using the same dtype update function.

thehomebrewnerd: @tamargrey I updated this to use a single `_convert_column_dtype` function for both table and column conversions now. Thanks for pointing that out!

thehomebrewnerd: Also, for now I left the Ordinal validation in - it's also in the `_validate_logical_type` method in the column accessor. Need to take a closer look at that, to see if there is a better way to handle that.

tamargrey: Awesome! Glad we're able to reuse so much code. I guess we can take Ordinal validation into account when we get to #515 as it's double validation between the `init_series` and Series Accessor init

tamargrey: Also, I like that name `_convert_column_dtype`!

woodwork/accessor_utils.py
```diff
@@ -0,0 +1,91 @@
...
+                else:
+                    series = pd.to_datetime(series, format=logical_type.datetime_format)
+            else:
+                if ks and isinstance(series, ks.Series) and logical_type.backup_dtype:
+                    new_dtype = logical_type.backup_dtype
+                else:
+                    new_dtype = logical_type.pandas_dtype
+                series = series.astype(new_dtype)
+        except (TypeError, ValueError):
+            error_msg = f'Error converting datatype for {series.name} from type {str(series.dtype)} ' \
```
tamargrey, line 87: makes sense that we don't need to pass the name through!

woodwork/tests/accessor/test_accessor_utils.py
```diff
@@ -0,0 +1,56 @@
+import pytest
+
+from woodwork.accessor_utils import init_series
+from woodwork.logical_types import Categorical, Datetime
+from woodwork.tests.testing_utils import xfail_dask_and_koalas
+
+
+def test_init_series_valid_conversion_specified_ltype(sample_series):
```
tamargrey, line 8: Maybe also worth adding a check for specifying a logical type that's different from the one this series will get inferred as.

tamargrey (commented):
Would this be the time to update the error message in `_validate_logical_type` at the column accessor to suggest initializing with `init_series`?

woodwork/accessor_utils.py
thehomebrewnerd: Yeah, sounds good. I made a comment about the Ordinal stuff in Issue #515 so we don't forget about it.

thehomebrewnerd:
> Would this be the time to update the error message in `_validate_logical_type` at the column accessor to suggest initializing with `init_series`?

Yes, I think that makes sense!

woodwork/tests/accessor/test_accessor_utils.py
thehomebrewnerd: Added!

tamargrey (approved):
LGTM!

woodwork/accessor_utils.py
```diff
@@ -0,0 +1,91 @@
...
+                f'to type {logical_type.pandas_dtype}. Please confirm the underlying data is consistent with ' \
+                f'logical type {logical_type}.'
+            raise TypeError(error_msg)
```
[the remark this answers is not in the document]
thehomebrewnerd: I'm fine with that. I'll update.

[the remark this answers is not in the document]
thehomebrewnerd: @gsheni Added a custom `TypeConversionError` exception for this.

gsheni (approved):
lgtm

PR #584, opened 2021-02-12
Move mutual info functions to statistics_utils file

### LINKED ISSUE #577: Move mutual information core logic to statistics utils file
We're creating a statistics utils file that'll contain our `describe` logic, `_get_mode`, and other functions that draw statistical insights into Woodwork data. The mutual information logic currently lives entirely on the Accessor, so we should move that over to the statistics file.

tamargrey:
- Moves `_replace_nans_for_mutual_info`, `_make_categorical_for_mutual_info`, and `_get_mutual_information_dict` from `table_accessor.py`
- also moves `_get_mode` from `utils.py`
- Closes #577

### DIFF
--- a/woodwork/table_accessor.py
+++ b/woodwork/table_accessor.py
@@ -5,15 +5,13 @@ from sklearn.metrics.cluster import normalized_mutual_info_score
 
 from woodwork.accessor_utils import _update_column_dtype
 from woodwork.exceptions import TypingInfoMismatchWarning
-from woodwork.logical_types import Datetime, Double
+from woodwork.logical_types import Datetime
 from woodwork.schema import Schema
-from woodwork.schema_column import (
-    _is_col_boolean,
-    _is_col_categorical,
-    _is_col_datetime,
-    _is_col_numeric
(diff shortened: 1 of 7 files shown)


woodwork/table_accessor.py
```diff
@@ -379,7 +275,7 @@ def mutual_information(self, num_bins=10, nrows=None):
             Mutual information values are between 0 (no mutual information) and 1
             (perfect dependency).
         """
-        mutual_info = self.mutual_information_dict(num_bins, nrows)
+        mutual_info = _get_mutual_information_dict(self._dataframe, num_bins=num_bins, nrows=nrows)
```
thehomebrewnerd, line 278: Can this call remain the same since `self.mutual_information_dict` is still a valid method? Doesn't matter too much, but the original call looks slightly cleaner.

thehomebrewnerd (approved):
Looks good. Just one minor comment.

woodwork/table_accessor.py
tamargrey: Good call. Changing

PR #588, opened 2021-02-12
Preserve underlying index

### LINKED ISSUE #569: Preserve the underlying pandas DataFrame's index in initialization of Woodwork data structure
- As a user, I would like Woodwork to preserve the underlying pandas DataFrame's index if I did not specify a specific `index` column to use. Right now, in WW 0.0.8 and 0.0.9, the index is replaced with `RangeIndex`. However, in EvalML, we want to preserve the user input's index. This is currently tricky to do without adding a new index column and then removing it.

Talked to @tamargrey about https://github.com/alteryx/woodwork/pull/464; we agreed a happy balance could be: 
- If user doesn't specify `index` parameter in initialization of WW DataTable, use underlying Index.
- If user later sets `index` and then removes it, `RangeIndex` is a good default.

#### Code Example

```python
X_df = pd.DataFrame([0, 1, 2, 3, 4], index=list(range(-5, 0)))
X_ww = ww.DataTable(X_df)
assert X_ww.to_dataframe().index == list(range(-5, 0))
```

jeff-hernandez:
Closes #569 by preserving the underlying index when returning to a data frame.


docs/source/release_notes.rst
```diff
@@ -7,14 +7,15 @@ Release Notes
     * Fixes
     * Changes
         * Avoid calculating mutualinfo for non-unique columns (:pr:`563`)
+        * Preserve underlying index (:pr:`588`)
```
thehomebrewnerd, line 10: Maybe add a little more detail: "Preserve underlying DataFrame index if index column is not specified"

woodwork/datatable.py
```diff
@@ -423,7 +424,11 @@ def drop(self, columns):
         if not_present:
             raise ValueError(f'{not_present} not found in DataTable')
 
-        return self._new_dt_from_cols([col for col in self._dataframe.columns if col not in columns])
+        dt = self._new_dt_from_cols([col for col in self._dataframe.columns if col not in columns])
+        if any('index' in self[column].semantic_tags for column in columns):
```
thehomebrewnerd, line 428: Can we use the index property here to simplify this code?
`if self.index in columns:`

thehomebrewnerd (changes requested):
Looks pretty good, just a few relatively minor suggestions.

tamargrey:
Did we decide in the end that `pop`, `drop`, and setting `dt.index = None` are the only actions where we want to reset the index?

Thinking through some other ways you can lose the index:
- removing the tag directly with `remove_semantic_tags`, `reset_semantic_tags`, or `set_types`
    - this seems more like actually setting index to None than the ones below
- `select`
- `iloc`
- getitem - `dt[list_of_cols]`

If we're trying to minimize the number of assumptions Woodwork is making, I can understand maybe letting the last three keep the underlying index since losing the index is more of a side effect than anything else.

jeff-hernandez:
@tamargrey I think it makes sense to also reset the index when removing the tags directly with `remove_semantic_tags`, `reset_semantic_tags`, or `set_types`.

woodwork/datatable.py
```diff
@@ -563,7 +571,12 @@ def remove_semantic_tags(self, semantic_tags):
             woodwork.DataTable: DataTable with the specified semantic tags removed
         """
         _check_semantic_tags(self._dataframe, semantic_tags)
-        return self._update_cols_and_get_new_dt('remove_semantic_tags', semantic_tags)
+        dt = self._update_cols_and_get_new_dt('remove_semantic_tags', semantic_tags)
+        for _, tags in semantic_tags.items():
```
thehomebrewnerd, line 575: Can just do `for tags in semantic_tags.values()` here since we discarding the keys.

thehomebrewnerd (approved):
Made one small comment, but otherwise this looks good. Please wait to merge until @tamargrey has had a chance to review though.

woodwork/datatable.py
```diff
@@ -524,16 +527,21 @@ def set_types(self, logical_types=None, semantic_tags=None, retain_index_tags=Tr
...
+            if 'index' in col.semantic_tags and not retain_index_tags:
+                reset_index = True
             if col_name in logical_types:
                 col = col.set_logical_type(logical_types[col_name], retain_index_tags)
             if col_name in semantic_tags:
                 col = col.set_semantic_tags(semantic_tags[col_name], retain_index_tags)
             cols_to_update[col_name] = col
 
         new_dt._update_columns(cols_to_update)
+        if reset_index:
```
tamargrey, line 543: Another option for this check is to have `if new_dt.index is None and self.index is not None`. Plus this check could be used here, `DataTable.drop`, and the other semantic tag update methods, since all of them return a new DataTable, so it might be nice to have these lines be similar between all of them.

tamargrey (changes requested):
Can we expand `test_underlying_index_on_update` to include tests that confirm that doing things like `pop`, `drop`, `rename`, and semantic tag updates reset the underlying index like we expect?

Maybe also worth documenting with tests the other DataTable behaviors that don't reset the index (iloc, getitem, etc)

woodwork/tests/datatable/test_datatable_indexing.py
```diff
@@ -162,6 +162,34 @@ def test_underlying_index_on_update(sample_df):
     assert type(dt._dataframe.index) == pd.Int64Index
     assert type(dt.to_dataframe().index) == pd.Int64Index
 
+    actual = dt.iloc[[0, 1]]
+    assert type(actual._dataframe.index) == pd.Index
+    assert type(actual.to_dataframe().index) == pd.Index
```
jeff-hernandez, line 167: A quirk comes up when setting the underlying index after using iloc of the datatable. Pandas converts Int64 to objects but preserves the dtype if set as int64.

```
   id         full_name
0   2       James Brown
1   3  Ms. Paige Turner
```
```python
assert str(df.id.dtype) == 'int64'
df.set_index('id', inplace=True)
assert str(df.index.dtype) == 'int64'
```
```python
df.reset_index(inplace=True)
df['id'] = df['id'].astype('Int64')
df.set_index('id', inplace=True)
assert str(df.index.dtype) == 'object'
```

tamargrey (approved):

Issue #582, opened 2021-02-12
Create new Schema objects when initing Woodwork with a Schema


tamargrey:
Currently, when doing `new_df.ww.init(df.ww.schema)`, the exact same Object is used for `df.ww.schema` and `new_df.ww.schema`, so if we allow Schemas to be mutable (which seems to be the plan), this will cause any changes to `new_df`'s schema to occur on `df`'s schema as well.

It should be investigated whether this exists with series' `col_dict` that stores typing information (probably does, but maybe the ability to use `**` when doing `new_series.ww.init(**series.ww._schema)` will mitigate that. 

We should make sure that series description and series metadata and table metadata dictionaries are all copied sufficiently that they're separate objects as well.

for Table Accessor can use `Schema._get_subset_schema(df.columns)` to copy the schema

tamargrey:
The Series Accessor now passes a **deep** copy of the schema dict through to `init` in the `__getattr__` when handling series operations so that metadata and semantic tags get new objects as well. We should still make sure to test this, though.

PR #589, opened 2021-02-16
Allow series calls from the accessor

### LINKED ISSUE #547: Allow Series method calls from the Series Accessor
Similar to how we allow `df.ww.sample`, we want to allow `series.ww.sample` in a way that returns a new series with the typing information initialized.

Note that `WoodworkTableAccessor._make_dataframe_call` will be a useful resource in making the Series Accessor's getattr.

thehomebrewnerd:
- Allow series calls from the accessor
- Closes #547 

Adds a `__getattr__` method to the series accessor to pass down any valid attribute/method calls from the accessor to the series, for anything not implemented on the accessor. For calls the return a new series, reinitializes Woodwork if the series dtype has not changed.

### DIFF
--- a/woodwork/column_accessor.py
+++ b/woodwork/column_accessor.py
@@ -1,5 +1,8 @@
+import warnings
+
 import pandas as pd
 
+from woodwork.exceptions import TypingInfoMismatchWarning
 from woodwork.logical_types import Ordinal
 from woodwork.schema_column import (
     _add_semantic_tags,
@@ -85,12 +88,57 @@ class WoodworkColumnAccessor:
             return False
         return self._series.equals(other._series)
 
+    def __getattr__(self, attr):
+        '''
+            If the method is present on the Accessor, uses that method.
+            If the method is present on DataFrame, uses that method.
+        '''
+        if self._schema is None:
+            raise AttributeError("Woodwork not initialized for this Series. Initialize by calling Series.ww.init")
+        if hasattr(self._series, attr):
+            return self._make_series_call(attr)
+        else:
+            raise AttributeError(f"Woodwork has no attribute '{attr}'")
+
     def __repr__(self):
         msg = u"<Series: {} ".format(self._series.name)
         msg += u"(Physical Type = {}) ".format(self._series.dtype)
         msg += u"(Logical Type = {}) ".format(self.logical_type)
         msg += u"(Semantic Tags = {})>".format(self.semantic_tags)
         return msg
+
+    def _make_series_call(self, attr):
+        '''
+        Forwards the requested attribute onto the series object.
+        Intercepts return value, attempting to initialize Woodwork with the current schema
+        when a new Series is returned.
+        Confirms schema is still valid for the original Series.
+        '''
+        series_attr = getattr(self._series, attr)
+
+        if callable(series_attr):
+            def wrapper(*args, **kwargs):
+                # Make Series call and intercept the result
+                result = series_attr(*args, **kwargs)
+
+                # Try to initialize Woodwork with the existing Schema
+                if isinstance(result, pd.Series):
+                    if result.dtype == self._schema['logical_type'].pandas_dtype:
+                        # We don't need to pass dtype from the schema to init
+                        schema = {key: value for key, value in self._schema.items() if key != 'dtype'}
+                        result.ww.init(**schema)
+                    else:
+                        invalid_schema_message = f'Operation performed by {attr} has invalidated the Woodwork ' \
+                            'typing information:\n' \
+                            f'dtype mismatch between original dtype, {self._schema["logical_type"].pandas_dtype}, ' \
+                            f'and returned dtype, {result.dtype}.\n' \
+                            'Please initialize Woodwork with Series.ww.init'
+                        warnings.warn(invalid_schema_message, TypingInfoMismatchWarning)
+                # Always return the results of the Series operation whether or not Woodwork is initialized
+                return result
+            return wrapper
+        # Directly return non-callable Series attributes
+        return series_attr
 
     def _validate_logical_type(self, logical_type):
         """Validates that a logical type is consistent with the series dtype. Performs additional type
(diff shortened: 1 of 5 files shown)


woodwork/column_accessor.py
```diff
@@ -85,13 +88,58 @@ def __eq__(self, other):
...
+                        result.ww.init(**schema)
+                    else:
+                        invalid_schema_message = f'Operation performed by {attr} has invalidated the Woodwork ' \
+                            'typing information:\n' \
+                            f'dtype mismatch between original dtype, {self._schema["logical_type"].pandas_dtype}, ' \
+                            f'and returned dtype, {result.dtype}.\n' \
+                            'Please initialize Woodwork with Series.ww.init'
+                        warnings.warn(invalid_schema_message, TypingInfoMismatchWarning)
+                # Always return the results of the Series operation whether or not Woodwork is initialized
+                return result
```
tamargrey, line 138: Is the reason that we don't need to worry about the schema getting invalidated inplace because you can't change the dtype of a series inplace?

woodwork/column_accessor.py
```diff
@@ -85,13 +88,58 @@ def __eq__(self, other):
...
+            def wrapper(*args, **kwargs):
+                # Make Series call and intercept the result
+                result = series_attr(*args, **kwargs)
+
+                # Try to initialize Woodwork with the existing Schema
+                if isinstance(result, pd.Series):
+                    if result.dtype == self._schema['logical_type'].pandas_dtype:
+                        # We don't need to pass dtype from the schema to init
+                        schema = {key: value for key, value in self._schema.items() if key != 'dtype'}
+                        result.ww.init(**schema)
```
tamargrey, line 129: This'll create a new object for the column dictionary, but it won't make a deepcopy. 

We can wait to handle this in #582 if we want, but I just wanted to point out that with this implementation, changes to `metadata` and `semantic_tags` would happen to both schemas. (The same issue exists with the Table Schema, but worse because the Schema is the exact same object at the moment)

woodwork/column_accessor.py
thehomebrewnerd: Basically yes. A dtype change was the only scenario I thought of that would invalidate (since we don't have column changes to worry about here), and since you can't modify the dtype in place I think we should be good. We can do inplace modifications (like `series.pop()` for example), but I think that should be safe.

woodwork/column_accessor.py
thehomebrewnerd: Ah, good point.

woodwork/column_accessor.py
tamargrey: Cool! 

Also not storing the name on the schema means that we can't invalidate it that way, which was the other way I was originally thinking could invalidate the schema!

woodwork/column_accessor.py
thehomebrewnerd: I went ahead and added a deepcopy operation here to resolve this issue.

tamargrey: Awesome! I'll add a note in 582 so that we remember to test this, but glad we've got it here!

woodwork/column_accessor.py
```diff
@@ -85,13 +89,59 @@ def __eq__(self, other):
             return False
         return self._series.equals(other._series)
 
+    def __getattr__(self, attr):
+        '''
+            If the method is present on the Accessor, uses that method.
+            If the method is present on DataFrame, uses that method.
```
tamargrey, line 95: Series instead of DataFrame

woodwork/column_accessor.py
```diff
@@ -85,13 +89,59 @@ def __eq__(self, other):
...
+            def wrapper(*args, **kwargs):
+                # Make Series call and intercept the result
+                result = series_attr(*args, **kwargs)
+
+                # Try to initialize Woodwork with the existing Schema
+                if isinstance(result, pd.Series):
+                    if result.dtype == self._schema['logical_type'].pandas_dtype:
+                        schema = copy.deepcopy(self._schema)
+                        # We don't need to pass dtype from the schema to init
+                        _ = schema.pop('dtype')
```
tamargrey, line 130: Bit of a nitpick: do we need to set the dummy `_` here? I figure we can just pop without setting the result to anything or, since we know `dtype` is in the schema dict, I'd be fine with doing `del schema['dtype']`

woodwork/column_accessor.py
```diff
@@ -85,13 +89,59 @@ def __eq__(self, other):
...
+                        result.ww.init(**schema)
+                    else:
+                        invalid_schema_message = f'Operation performed by {attr} has invalidated the Woodwork ' \
+                            'typing information:\n' \
+                            f'dtype mismatch between original dtype, {self._schema["logical_type"].pandas_dtype}, ' \
+                            f'and returned dtype, {result.dtype}.\n' \
+                            'Please initialize Woodwork with Series.ww.init'
+                        warnings.warn(invalid_schema_message, TypingInfoMismatchWarning)
```
tamargrey, line 138: Might not be worth changing the `get_warning_message` for this warning so that it works for both Series and Dataframe, but since the warning messages are otherwise the same, it'd be nice to have.

woodwork/column_accessor.py
thehomebrewnerd: Fixed!

woodwork/column_accessor.py
thehomebrewnerd: Not sure what's preferred...matter of style, I guess. I generally thought it was good to set a dummy to acknowledge that you were intentionally discarding the return, but not a big deal either way. Maybe the `del scheam['dtype']` is better here though. I'll change.

woodwork/column_accessor.py
thehomebrewnerd: Yeah, that thought crossed my mind as well. I think it should be a fairly simple change. I'll update.

tamargrey (approved):
lgtm!

PR #590, opened 2021-02-17
Add set_logical_type method to series accessor

### LINKED ISSUE #567: Implement method to update series logical type after init
The current series accessor implementation does not allow users to change the logical type after a series has been initialized. We should implement a `set_logical_type` method to allow users to change the series logical type, as was done previously on the DataColumn object.

This method should only succeed if the new logical type is compatible with the series dtype as in place modifications to the series are not possible. If dtype changes are required, the user will likely need to use the helper function created in #554.

thehomebrewnerd:
- Add set_logical_type method to series accessor
- Closes #567 

This PR adds a set_logical_type method to the series accessor. A few things to note, as this behaves somewhat differently than the previous DataColumn method:

- Tests related to retaining index and time_index tags have been removed. Since the table accessor does not rely on series having index or time_index tags any longer, these tags are now treated just like any other tags on the column accessor
- Test related to updating the dtype on init or on a logical type change have been removed. Since it is not possible to change a series dtype in place, these tests no longer apply and users will need to use the `ww.init_series()` helper function to initialize a series with a dtype change.
- Attempting to set a logical type that is incompatible with the dtype of the underlying series will not result in an error.

### DIFF
--- a/woodwork/accessor_utils.py
+++ b/woodwork/accessor_utils.py
@@ -73,8 +73,8 @@ def _update_column_dtype(series, logical_type):
                 series.name = name
             elif ks and isinstance(series, ks.Series):
                 series = ks.Series(ks.to_datetime(series.to_numpy(),
-                                                    format=logical_type.datetime_format),
-                                    name=series.name)
+                                                  format=logical_type.datetime_format),
+                                   name=series.name)
             else:
                 series = pd.to_datetime(series, format=logical_type.datetime_format)
         else:
--- a/woodwork/accessor_utils.py
+++ b/woodwork/accessor_utils.py
@@ -63,29 +63,29 @@ def _update_column_dtype(series, logical_type):
             series = ks.from_pandas(formatted_series)
         else:
             series = series.apply(_reformat_to_latlong)
-
-    # Update the underlying series
-    try:
-        if _get_ltype_class(logical_type) == Datetime:
-            if dd and isinstance(series, dd.Series):
-                name = series.name
-                series = dd.to_datetime(series, format=logical_type.datetime_format)
-                series.name = name
-            elif ks and isinstance(series, ks.Series):
-                series = ks.Series(ks.to_datetime(series.to_numpy(),
-                                                  format=logical_type.datetime_format),
-                                   name=series.name)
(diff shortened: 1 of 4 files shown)


woodwork/column_accessor.py
```diff
@@ -184,6 +185,25 @@ def reset_semantic_tags(self):
...
+        """Update the logical type for the series, clearing any previously set semantic tags,
+        and returning a new Series.
+
+        Args:
+            logical_type (LogicalType, str): The new logical type to set for the series.
+
+        Returns:
+            Series: A new series with the updated logical type.
+        """
+        logical_type = _get_column_logical_type(self._series, logical_type, self._series.name)
```
tamargrey, line 198: `init_series` is going to just call `_get_column_logical_type` on the logical type. So not sure if this line is necessary.

And the fact that this can basically just be a wrapper for `init_series` makes me think even more that `set_logical_type` on the column isn't super important....

thehomebrewnerd: Good catch. I'll remove the duplicate call here.

I agree on the usefulness of this method, but will still create an issue to review in more detail before we remove completely.

woodwork/accessor_utils.py
```diff
@@ -64,29 +64,28 @@ def _update_column_dtype(series, logical_type):
         else:
             series = series.apply(_reformat_to_latlong)
 
-    if logical_type.pandas_dtype != str(series.dtype):
```
tamargrey, line 67: Guessing this is being removed because, for `init_series`, we need a new series object to be created even if the dtype doesn't need to change.

With DataFrame's, though, where we're happy to keep the same series object, we should probably be cognizant of saving time where possible and not unnecessarily calling `astype`

thehomebrewnerd: Yes that is correct. If you prefer, we can leave this statement in place, and we can handle the creation of the new series object inside the set_logical_type method instead.

tamargrey: I think that'd make sense

thehomebrewnerd: Updated

tamargrey (approved):
Lgtm!

PR #593, opened 2021-02-18
523 ignored parameters warning

### LINKED ISSUE #523: Don't allow passing other parameters to `init` when initializing with a Schema
- Currently if a user passes parameters that are not `schema` to `df.ww.init` along with the `schema` param, the other parameters get quietly ignored.

We should consider handling this with one of the following:
- Raising an error, not letting users do this
- Raising a warning letting users know that the other parameters are being ignored
- Incorporate these parameters into the schema somehow

johnbridstrup:
- Implements a warning if additional parameters are passed along with a schema. Ignored parameters are listed
- Added test_accessor_with_schema_parameter_warning to test suite
- Closes #523

-----

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -18,6 +18,7 @@ Release Notes
         * Add semantic tag update methods to column accessor (:pr:`573`)
         * Add ``describe`` and ``describe_dict`` to WoodworkTableAccessor (:pr:`579`)
         * Add ``init_series`` util function for initializing a series with dtype change (:pr:`581`)
+        * Add warning if additional parameters are passed along with schema (:pr:`593`)
 
     * Fixes
     * Changes
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -20,7 +20,6 @@ Release Notes
         * Add ``init_series`` util function for initializing a series with dtype change (:pr:`581`)
         * Add semantic tag update methods to table schema (:pr:`591`)
         * Add warning if additional parameters are passed along with schema (:pr:`593`)
-
     * Fixes
     * Changes
         * Avoid calculating mutualinfo for non-unique columns (:pr:`563`)
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -22,7 +22,6 @@ Release Notes
         * Add semantic tag update methods to table schema (:pr:`591`)
         * Add warning if additional parameters are passed along with schema (:pr:`593`)
         * Better warning when accessing column properties before init (:pr:`596`)
-
     * Fixes
         * Create new Schema object when performing pandas operation on Accessors (:pr:`595`)
     * Changes
--- a/woodwork/table_accessor.py
+++ b/woodwork/table_accessor.py
@@ -3,7 +3,10 @@ import warnings
 import pandas as pd
 
 from woodwork.accessor_utils import _update_column_dtype
-from woodwork.exceptions import TypingInfoMismatchWarning, ParametersIgnoredWarning
+from woodwork.exceptions import (
+    ParametersIgnoredWarning,
+    TypingInfoMismatchWarning
+)
 from woodwork.logical_types import Datetime
 from woodwork.schema import Schema
 from woodwork.statistics_utils import (
--- a/woodwork/table_accessor.py
+++ b/woodwork/table_accessor.py
@@ -75,19 +75,15 @@ class WoodworkTableAccessor:
             extra_params = []
             if index is not None:
                 extra_params.append('index')
-                print('index')
             if make_index:
                 extra_params.append('make_index')
-                print('make_index')
             if time_index is not None:
                 extra_params.append('time_index')
-                print('time_index')
             if logical_types is not None:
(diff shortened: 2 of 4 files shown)


docs/source/release_notes.rst
```diff
@@ -19,6 +19,8 @@ Release Notes
         * Add ``describe`` and ``describe_dict`` to WoodworkTableAccessor (:pr:`579`)
         * Add ``init_series`` util function for initializing a series with dtype change (:pr:`581`)
         * Add semantic tag update methods to table schema (:pr:`591`)
+        * Add warning if additional parameters are passed along with schema (:pr:`593`)
+
```
thehomebrewnerd, line 23: Remove extra blank line

woodwork/table_accessor.py
```diff
@@ -72,6 +72,20 @@ def init(self,
...
+            if make_index:
+                extra_params.append('make_index')
+            if time_index is not None:
+                extra_params.append('time_index')
+            if logical_types is not None:
+                extra_params.append('logical_types')
+            for key in kwargs:
+                extra_params.append(key)
+            if extra_params:
+                warnings.warn("A schema was provided and the following parameters were ignored: " + ", ".join(extra_params), Warning)
```
thehomebrewnerd, line 87: We have started to implement custom warning classes to easily allow users to suppress specific warnings they don't care about. I think we should follow that pattern here. Maybe create a new warning class inside `exceptions.py`, maybe something like `ParameterIgnoredWarning`, and use that here instead of the base `Warning` class.

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -162,6 +162,21 @@ def test_init_accessor_with_schema_errors(sample_df):
...
 
+def test_accessor_with_schema_parameter_warning(sample_df):
+    xfail_dask_and_koalas(sample_df)
+
+    schema_df = sample_df.copy()
+    schema_df.ww.init(name='test_schema', semantic_tags={'id': 'test_tag'}, index='id')
+    schema = schema_df.ww.schema
+
+    head_df = schema_df.head(2)
+    assert head_df.ww.schema is None
```
thehomebrewnerd, line 173: I think we can remove this assertion as it isn't important for this test.

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -162,6 +162,21 @@ def test_init_accessor_with_schema_errors(sample_df):
...
+
+    warning = "A schema was provided and the following parameters were ignored: index, semantic_tags"
+    with pytest.warns(Warning, match=warning):
+        head_df.ww.init(index='id', semantic_tags={'id': 'test_tag'}, schema=schema)
```
thehomebrewnerd, line 177: Let's use some different values for index and semantic tags, just so we can confirm below that the values from the schema are being used and not the values we are passing for these parameters.

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -162,6 +162,21 @@ def test_init_accessor_with_schema_errors(sample_df):
...
+
```
thehomebrewnerd, line 178: Let's also check that the head_df name and semantic tags are what we expect, since the initialization should still succeed even though we warn the user.
```
    assert head_df.ww.name == 'test_schema'
    assert head_df.ww.semantic_tags['id'] == {'index', 'test_tag'}
```

thehomebrewnerd (changes requested):
@johnbridstrup Thanks for making another contribution! I have a few suggested changes to make here, but generally this looks good.

tamargrey (changes requested):
Thanks for contributing @johnbridstrup!

woodwork/tests/accessor/test_table_accessor.py
johnbridstrup: That's a good call. There are issues with _validate_accessor_params if we use different values since it doesn't know they are being ignored. I'll get to this later today.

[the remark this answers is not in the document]
johnbridstrup: See my response to @thehomebrewnerd. Params have to match the original dataframe column labels or _validate_accessor_params throws a LookupError. I will fix this later today. Thanks!

johnbridstrup: Just kidding they were all quick fixes. Let me know if there's anything else!

thehomebrewnerd (changes requested):
One more small update to make.

thehomebrewnerd:
> One more small update to make.

I'm not sure what happened to the comment I made, but we also need to add a check for the `already_sorted` parameter in `init` as we are also ignoring this if it is passed.

thehomebrewnerd:
@johnbridstrup Looks like codecov is complaining about some of the new checks being untested. You should be able to fix this by adding all the parameters that are checked in the `test_accessor_with_schema_parameter_warning` test.

thehomebrewnerd (approved):
LGTM!

tamargrey (approved):

gsheni:
Thanks for the contribution @johnbridstrup

PR #595, opened 2021-02-18
Create new schema object when going through pandas operations

### LINKED ISSUE #582: Create new Schema objects when initing Woodwork with a Schema

tamargrey:
- Creates a new schema object when we do operations like `new_df = df.ww.head()` so that any changes don't impact `df`'s schema and `new_df`'s schema. 
- Uses deepcopy where necessary (metadata is the big one) 
- adds tests that confirm we have new objects and that changes don't impact both
- Closes #582

### DIFF
--- a/woodwork/table_accessor.py
+++ b/woodwork/table_accessor.py
@@ -68,7 +68,7 @@ class WoodworkTableAccessor:
             column_descriptions (dict[str -> str], optional): Dictionary mapping column names to column descriptions.
             schema (Woodwork.Schema, optional): Typing information to use for the DataFrame instead of performing inference.
                 Any other arguments provided will be ignored. Note that any changes made to the typing information
-                after passing in the typing information will propogate to the original object. It is not advised to 
+                after passing in the typing information will propogate to the original object. It is not advised to
                 share typing information between DataFrames.
         '''
         _validate_accessor_params(self._dataframe, index, make_index, time_index, logical_types, schema)
@@ -124,7 +124,10 @@ class WoodworkTableAccessor:
 
     @property
     def schema(self):
-        return self._schema
+        ''' A copy of the Woodwork typing information for the DataFrame.
+        '''
+        if self._schema:
+            return self._schema._get_subset_schema(list(self.columns.keys()))
 
     def select(self, include):
         """Create a DataFrame with Woodowork typing information initialized
--- a/woodwork/table_accessor.py
+++ b/woodwork/table_accessor.py
@@ -67,7 +67,9 @@ class WoodworkTableAccessor:
                 specified logical type for the column. Defaults to True.
             column_descriptions (dict[str -> str], optional): Dictionary mapping column names to column descriptions.
             schema (Woodwork.Schema, optional): Typing information to use for the DataFrame instead of performing inference.
-                Any other arguments provided will be ignored.
+                Any other arguments provided will be ignored. Note that any changes made to the typing information
+                after passing in the typing information will propogate to the original object. It is not advised to 
+                share typing information between DataFrames.
         '''
         _validate_accessor_params(self._dataframe, index, make_index, time_index, logical_types, schema)
         if schema is not None:
--- a/woodwork/table_accessor.py
+++ b/woodwork/table_accessor.py
@@ -67,9 +67,9 @@ class WoodworkTableAccessor:
                 specified logical type for the column. Defaults to True.
             column_descriptions (dict[str -> str], optional): Dictionary mapping column names to column descriptions.
             schema (Woodwork.Schema, optional): Typing information to use for the DataFrame instead of performing inference.
-                Any other arguments provided will be ignored. Note that any changes made to the typing information
-                after passing in the typing information will propogate to the original object. It is not advised to
-                share typing information between DataFrames.
+                Any other arguments provided will be ignored. Note that any changes made to the schema object after
+                initialization will propagate to the DataFrame. Similarly, to avoid unintended typing information changes,
+                the same schema object should not be shared between DataFrames.
         '''
         _validate_accessor_params(self._dataframe, index, make_index, time_index, logical_types, schema)
         if schema is not None:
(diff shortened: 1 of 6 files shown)


woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -129,13 +129,15 @@ def test_init_accessor_with_schema(sample_df):
     assert head_df.ww.schema is None
     head_df.ww.init(schema=schema)
 
+    assert head_df.ww.schema is schema
```
thehomebrewnerd, line 132: I think in this case we have head_df and schema_df both sharing the same underlying schema, so that changes to either one will be propagated to the other. Is this the behavior we want or should we also break that linkage here?

tamargrey: Currently this is intentional. I was thinking originally that it should be up to the user to create a new object since I'm not sure that it's expected that Woodwork would create a new object at init. 

The situation where users are likely to pass in a schema themselves is when the Schema gets invalidated with an operation. In that case, users will want to take the original Schema and change it around until it's valid. Since schemas are mutable, they'll need to copy the schema or they'll be changing it for the original DataFrame as well--this creates a new object anyway. So I'm not sure when users would be passing a Schema from another DataFrame through without having to first make a new object since they should be going through the accessor for those.

But whether or not Woodwork creates a new object, we need to give users a way to copy a Schema if Schemas aren't going to be immutable.

thehomebrewnerd: Reading the initial issue, I was under the impression that we would create a new schema object during init as well. I don't have a strong preference here, but we might want to make sure the behavior is clear in the documentation.

tamargrey: I guess it occurred to me when I saw the Series Accessor getattr PR that it would make sense to create the new Table schema at the getattr instead of having it in the init. (if we do decide to have a new object at accessor init then we’ll need a corresponding change in the series accessor to copy metadata and semantic tags inits)

Will definitely make sure to mention in the `init` that the schema passed in gets used and that a new object is not created so users should be careful about passing in a schema that's used elsewhere.

thehomebrewnerd: Just thinking out loud here a bit, but what if we were to update the user facing `.schema` property on the table accessor to return a new schema object with copies of the dictionary. That way, any time the user is trying to get the schema to modify it, like in the scenario that you described above, they are working with a copy. That way, users don't have to worry about these side effects and we don't make a new object during init unnecessarily.

tamargrey: I like that!

tamargrey: Added the copying behavior for the schema property and updated relevant tests to make sure we were referencing `_schema` if necessary

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -123,19 +132,21 @@ def test_init_accessor_with_schema(sample_df):
 
     schema_df = sample_df.copy()
     schema_df.ww.init(name='test_schema', semantic_tags={'id': 'test_tag'}, index='id')
-    schema = schema_df.ww.schema
+    schema = schema_df.ww._schema
```
thehomebrewnerd, line 135: Really minor point, but this can stay as `schema_df.ww.schema` since we are not using the schema_df schema in any of the assertions that follow.

woodwork/table_accessor.py
```diff
@@ -67,7 +67,9 @@ def init(self,
                 specified logical type for the column. Defaults to True.
             column_descriptions (dict[str -> str], optional): Dictionary mapping column names to column descriptions.
             schema (Woodwork.Schema, optional): Typing information to use for the DataFrame instead of performing inference.
-                Any other arguments provided will be ignored.
+                Any other arguments provided will be ignored. Note that any changes made to the typing information
+                after passing in the typing information will propogate to the original object. It is not advised to
```
thehomebrewnerd, line 71: Hmm. Is this true now assuming the original schema was accessed through `df.ww.schema`? Maybe I'm getting tripped up by the work `original` here. If I'm understanding things correctly these hold true:
- If a user initializes with a schema object obtained from another dataframe, and then makes changes to the schema after initialization, the changes to the schema will propagate to the second object, but not the first object from which the schema originated
- If any two or more dataframes are initialized with the same schema object, changes to one will be propagated to the others

If I'm following properly, maybe we should update this to something like: 
```
Any other arguments provided will be ignored. Note that any changes made to the schema object after
initialization will propagate to the DataFrame. Similarly, to avoid unintended typing information changes,
the same schema object should not be shared between DataFrames.
```

thehomebrewnerd (approved):
A couple minor suggestions, but otherwise looks good. If you don't think the docstring update I proposed makes sense, feel free to ignore.

woodwork/tests/accessor/test_table_accessor.py
tamargrey: haha yeah I forgot that by passing `schema` through that that'd be the object no matter what. Thanks for pointing this out!

woodwork/table_accessor.py
tamargrey: Yes both of those are correct, and I think your suggestion displays that really well. I'll change.

PR #596, opened 2021-02-18
Better error when accessing column accessor properties before init

### LINKED ISSUE #572: Improve error when accessing series properties before init
In the series accessor implementation, users can attempt to access various properties without first initializing Woodwork. For most properties, this will result in an error that is not very meaningful for the user: `TypeError: 'NoneType' object is not subscriptable`

We should update the implementation of the properties to return a better error to the user, instructing them to first initialize Woodwork.

thehomebrewnerd:
- Pull Request Description (replace this text with your description)
- Closes #572 

Adds a better warning message when trying to access properties on Column Accessor before initialization.

### DIFF
--- a/woodwork/column_accessor.py
+++ b/woodwork/column_accessor.py
@@ -110,6 +110,8 @@ class WoodworkColumnAccessor:
             raise AttributeError(f"Woodwork has no attribute '{attr}'")
 
     def __repr__(self):
+        if self._schema is None:
+            _raise_init_warning()
         msg = u"<Series: {} ".format(self._series.name)
         msg += u"(Physical Type = {}) ".format(self._series.dtype)
         msg += u"(Logical Type = {}) ".format(self.logical_type)
--- a/woodwork/column_accessor.py
+++ b/woodwork/column_accessor.py
@@ -57,6 +57,8 @@ class WoodworkColumnAccessor:
     @property
     def description(self):
         """The description of the series"""
+        if self._schema is None:
+            _raise_init_warning()
         return self._schema['description']
 
     @description.setter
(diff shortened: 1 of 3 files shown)


tamargrey:
I think it could be useful to handle the uninitialized schema case on the `__repr__` as well. Currently if you do `series.ww` before initializing it errors with `TypeError: 'NoneType' object is not subscriptable` as well. Not sure if a better error message or some other handling is best, though

thehomebrewnerd:
> I think it could be useful to handle the uninitialized schema case on the `__repr__` as well. Currently if you do `series.ww` before initializing it errors with `TypeError: 'NoneType' object is not subscriptable` as well. Not sure if a better error message or some other handling is best, though

I think this case should be handled by default without changes to the code, but it is probably better to explicitly catch this in the code. Since `__repr__` uses `self.logical_type`, it will still raise the error if not initialized. I added a test case to cover this.

woodwork/column_accessor.py
```diff
@@ -195,3 +205,7 @@ def set_semantic_tags(self, semantic_tags):
         self._schema['semantic_tags'] = _set_semantic_tags(semantic_tags,
                                                            self.logical_type.standard_tags,
                                                            self.use_standard_tags)
+
+
+def _raise_init_warning():
```
tamargrey, line 210: This name makes it sound like we're warning users and not stopping them with an Error

thehomebrewnerd: Yep - I'll change. I don't know why I got tripped up on warning versus error here. I also did the same thing with the title of the pr...

tamargrey (approved):
Looks good!

Issue #493, opened 2021-01-15
Add Koalas Table Accessor


tamargrey:
Allows Koalas DataFrames to use the `ww` namespace for accessors.

Involves adding any additional tests or validation that might be necessary with the new API.

tamargrey:
In checking `_is_valid_schema`, we check that all they dtypes for each column match. We should make sure to use the backup dtype for koalas dataframes.

tamargrey:
The `ks.set_option('compute.ops_on_diff_frames', True)` line will need to be added to `table_accessor.py` in a way that works in accordance with #543--so likely in the `WoodworkTableAccessor.init` method.

thehomebrewnerd:
When adding support for Koalas, the function `_is_valid_latlong_value` will need to be updated to account for the need to use lists instead of tuples for latlong values in Koalas.

thehomebrewnerd:
Closed by #652

PR #598, opened 2021-02-19
Column accessor update for LatLong columns

### LINKED ISSUE #568: Updates for proper handling of LatLong columns

thehomebrewnerd:
- Column accessor update for LatLong columns
- Closes #568

Allows for direct initialization of LatLong columns via `series.ww.init` if the data is properly formatted, otherwise raises an error. LatLong columns that require reformatting can be initialized via `ww.init_series`.

### DIFF
--- a/woodwork/tests/test_utils.py
+++ b/woodwork/tests/test_utils.py
@@ -32,6 +32,8 @@ from woodwork.utils import (
     _is_null_latlong,
     _is_s3,
     _is_url,
+    _is_valid_latlong_series,
+    _is_valid_latlong_value,
     _new_dt_including,
     _reformat_to_latlong,
     _to_latlong_float,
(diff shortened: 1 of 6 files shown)


woodwork/tests/test_utils.py
```diff
@@ -436,6 +438,39 @@ def test_is_null_latlong():
     assert not _is_null_latlong(False)
 
 
+def test_is_valid_latlong_value():
+    values = [
+        (1.0, 2.0),
+        (np.nan, np.nan),
+        [1.0, 2.0],
+        np.nan,
```
tamargrey, line 446: Based off of these lines in `_reformat_to_latlong`, should this be a valid latlong?

```python
        # (np.nan, np.nan) should be counted as a single null value
        if pd.isnull(latitude) and pd.isnull(longitude):
            return np.nan
```

woodwork/utils.py
```diff
@@ -264,6 +264,23 @@ def _to_latlong_float(val):
...
+    otherwise returns False'''
+    if series.apply(_is_valid_latlong_value).all():
+        return True
+    return False
+
+
+def _is_valid_latlong_value(val):
+    '''Returns True if the value provided is a properly formatted LatLong value, otherwise
+    returns False'''
+    if isinstance(val, tuple) and len(val) == 2:
```
tamargrey, line 278: Something to make note of for #493, but we'll need the definition of a valid latlong to use lists instead of tuples for koalas

woodwork/tests/test_utils.py
thehomebrewnerd: Ah, yes. I was thinking `(np.nan, np.nan)` was the valid null value, but it's just `np.nan`.So I think our valid data should be:
- a tuple with two floats, provided they are not both `np.nan`
- a single `np.nan` value

Is that consistent with your thoughts, @tamargrey?

tamargrey: yep!

woodwork/utils.py
thehomebrewnerd: Added a note in that issue

woodwork/utils.py
```diff
@@ -274,10 +274,14 @@ def _is_valid_latlong_series(series):
 
 def _is_valid_latlong_value(val):
     '''Returns True if the value provided is a properly formatted LatLong value, otherwise
-    returns False'''
+    returns False.'''
     if isinstance(val, tuple) and len(val) == 2:
         if isinstance(val[0], float) and isinstance(val[1], float):
+            if pd.isnull(val[0]) and pd.isnull(val[1]):
```
tamargrey, line 280: bit of a nitpick, but it might read nicely to have `latitude, longitude = val` and then reference those variables in this line and the once above.

thehomebrewnerd: Updated

tamargrey (approved):
lgtm!

PR #601, opened 2021-02-19
Update tests.yml to fix release notes tests

gsheni:
- The release note tests were running on dependency update MRs (which they should not be).


jeff-hernandez (approved):
Looks good!

Issue #602, opened 2021-02-19
@property does not pass from Schema to Accessor


tamargrey:
We access Schema `@properties` like `index`, `semantic_tags`, etc via the WoodworkTableAccessor with a `__getattr__` that makes the Schema call. However, the `property` is not being propagated, so users can set those values and interact with them in ways that they couldn't if they directly accessed them through the Schema.

Here's an example of behavior that we don't want to be possible:
```python
df.ww.init()
assert df.ww.index is None

df.ww.index = 'fake_column'

assert df.ww.index == 'fake_column'
assert df.ww.schema.index is None

df.ww.set_index('id')
assert df.ww.index == 'fake_column'
assert df.ww.schema.index == 'id'
```

We can check whether a attr being accessed is a `property` with the following `isinstance(getattr(Schema, attr), property)`, but I'm not sure how we'd go about making that attr a property on the WoodworkTableAccessor class programmatically.

We may need to define the necessary Schema properties on the Accessor. Whether that would mean keeping the properties on the Schema class and wrapping them in the Accessor class or defining them fully on the Accessor is tbd. Likely, we'll still want those properties on the Schema class so that users can interact with a Schema by itself.

Worth noting that to have `index` and `time_index` setters on the Table Accessor, we need to have corresponding getters.

PR #603, opened 2021-02-19
Allow updating index 

### LINKED ISSUE #586: Implement index and time index updates on Table Accessor
Once semantic tag updates are allowed, we want to allow both `set_index` and @ setter's that facilitate changing the index and resetting it to None

The DataFrame will change when an index is set so we should consider if we want this to happen in place or not.

tamargrey:
- Adds `set_index` that allows for creating, updating, or removing an index column. Performs relevant checks for whether the column is a valid index column.
- Closes #586

### DIFF
--- a/docs/source/api_reference.rst
+++ b/docs/source/api_reference.rst
@@ -44,7 +44,7 @@ Schema
     Schema.add_semantic_tags
     Schema.remove_semantic_tags
     Schema.reset_semantic_tags
-    Schema.set_semantic_tags
+    Schema.set_types
 
 
 DataTable
--- a/docs/source/api_reference.rst
+++ b/docs/source/api_reference.rst
@@ -10,8 +10,13 @@ WoodworkTableAccessor
     :toctree: generated/
 
     WoodworkTableAccessor
+    WoodworkTableAccessor.describe
+    WoodworkTableAccessor.describe_dict
     WoodworkTableAccessor.init
+    WoodworkTableAccessor.mutual_information_dict
+    WoodworkTableAccessor.mutual_information
     WoodworkTableAccessor.select
+    WoodworkTableAccessor.set_index
 
 WoodworkColumnAccessor
 ======================
@@ -22,6 +27,11 @@ WoodworkColumnAccessor
 
     WoodworkColumnAccessor
     WoodworkColumnAccessor.init
+    WoodworkColumnAccessor.add_semantic_tags
+    WoodworkColumnAccessor.remove_semantic_tags
+    WoodworkColumnAccessor.reset_semantic_tags
+    WoodworkColumnAccessor.set_logical_type
+    WoodworkColumnAccessor.set_semantic_tags
 
 Schema
 ======
@@ -31,6 +41,10 @@ Schema
     :toctree: generated/
 
     Schema
+    Schema.add_semantic_tags
+    Schema.remove_semantic_tags
+    Schema.reset_semantic_tags
+    Schema.set_semantic_tags
 
 
 DataTable
--- a/woodwork/table_accessor.py
+++ b/woodwork/table_accessor.py
@@ -159,7 +159,6 @@ class WoodworkTableAccessor:
         """
         self._schema.set_index(new_index)
 
-        # --> by putting this after it follows separation paradigm but means the error message we get will come from schema
         if new_index is not None:
             _check_index(self._dataframe, new_index)
         self._set_underlying_index()
--- a/woodwork/table_accessor.py
+++ b/woodwork/table_accessor.py
@@ -148,6 +148,22 @@ class WoodworkTableAccessor:
         if self._schema:
             return self._schema._get_subset_schema(list(self.columns.keys()))
 
+    def set_index(self, new_index):
+        """Sets the index column of the DataFrame. Adds the 'index' semantic tag to the column
+        and clears the tag from any previously set index column.
+        Setting a column as the index column will also cause any previously set standard
+        tags for the column to be removed.
+
+        Args:
+            index (str): The name of the column to set as the index
+        """
+        self._schema.set_index(new_index)
+
+        # --> by putting this after it follows separation paradigm but means the error message we get will come from schema
+        if new_index is not None:
+            _check_index(self._dataframe, new_index)
+        self._set_underlying_index()
+
     def select(self, include):
         """Create a DataFrame with Woodowork typing information initialized
         that includes only columns whose Logical Type and semantic tags are
--- a/woodwork/table_accessor.py
+++ b/woodwork/table_accessor.py
@@ -159,8 +159,8 @@ class WoodworkTableAccessor:
         """
         self._schema.set_index(new_index)
 
-        if new_index is not None:
-            _check_index(self._dataframe, new_index)
+        if self._schema.index is not None:
+            _check_index(self._dataframe, self._schema.index)
         self._set_underlying_index()
 
     def select(self, include):
--- a/woodwork/tests/accessor/test_table_accessor.py
+++ b/woodwork/tests/accessor/test_table_accessor.py
@@ -1348,6 +1348,20 @@ def test_accessor_set_index(sample_df):
     # --> add setter test
 
 
+def test_accessor_set_index_errors(sample_df):
+    xfail_dask_and_koalas(sample_df)
+
+    sample_df.ww.init()
+
+    error = 'Specified index column `testing` not found in Schema.'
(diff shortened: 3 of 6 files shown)


woodwork/table_accessor.py
```diff
@@ -129,6 +129,21 @@ def schema(self):
         if self._schema:
             return self._schema._get_subset_schema(list(self.columns.keys()))
 
+    def set_index(self, new_index):
```
tamargrey, line 132: I'm not sure that we need an index `@setter`. Previously, the difference between `DataTable.set_index` and the setter was that the setter made the change inplace while `set_index` created a new DataTable. We've been sticking with inplace changes so far, so it seems redundant to me to have both.

My reasoning for going with `set_index` over the setter was that we'd need to add the getter as well, and it already exists on the Schema.

This thought process also lead me to discover that the `@property` tag doesn't propagate with the way we get Schema attrs through the Accessor. I documented this in #602

woodwork/table_accessor.py
```diff
@@ -129,6 +129,21 @@ def schema(self):
...
+        tags for the column to be removed.
+
+        Args:
+            index (str): The name of the column to set as the index
+        """
+        self._schema.set_index(new_index)
+
+        if self._schema.index is not None:
+            _check_index(self._dataframe, self._schema.index)
+        self._set_underlying_index()
```
tamargrey, line 145: @thehomebrewnerd here's an example of Schema/Accessor separation: Setting the index on the schema and then using that updated schema to change the dataframe. 

We don't have to keep it like this, but I thought I'd show an example of it.

thehomebrewnerd: So this pattern makes me think about a scenario where the _check_index validation might fail (say a column is not unique). We will have already set the index in the schema to an invalid column and it will remain that way. Not suggesting we need to fix at this point, because presumably, the user would then do something to fix the error from _check_index, which could then resync the information.

Just something to think about I guess.

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -677,39 +677,37 @@ def test_underlying_index(sample_df):
...
-    # assert (schema._dataframe.index == schema.to_dataframe()['full_name']).all()
-    # assert schema._dataframe.index.name is None
-    # assert type(schema._dataframe.index) == specified_index
-    # assert type(schema.to_dataframe().index) == specified_index
+    schema_df.ww.set_index('full_name')
+    assert (schema_df.index == schema_df['full_name']).all()
+    assert schema_df.index.name is None
+    assert type(schema_df.index) == specified_index
 
+    # --> add back in with setter
```
thehomebrewnerd, line 693: We should delete completely if we decide we don't need a setter for now.

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -677,39 +677,37 @@ def test_underlying_index(sample_df):
...
-    # schema = Schema(sample_df.copy())
-    # schema = schema.set_index('full_name')
-    # assert (schema._dataframe.index == schema.to_dataframe()['full_name']).all()
-    # assert schema._dataframe.index.name is None
-    # assert type(schema._dataframe.index) == specified_index
-    # assert type(schema.to_dataframe().index) == specified_index
+    schema_df.ww.set_index('full_name')
+    assert (schema_df.index == schema_df['full_name']).all()
+    assert schema_df.index.name is None
+    assert type(schema_df.index) == specified_index
```
thehomebrewnerd, line 691: I'm not sure where it belongs exactly, but I think we should make sure somewhere that when we set the underlying dataframe index that the index column also remains present in the DataFrame columns as this is the behavior we expect. Maybe I missed it, but I'm not seeing that check.

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -1304,3 +1302,78 @@ def test_select_repetitive(sample_df):
...
+
+
+def test_accessor_set_index(sample_df):
+    xfail_dask_and_koalas(sample_df)
+
+    sample_df.ww.init()
+
+    sample_df.ww.set_index('id')
+    assert sample_df.ww.index == 'id'
+    assert (sample_df.index == [0, 1, 2, 3]).all()
```
thehomebrewnerd, line 1314: Maybe update this to use the pattern below where we check that the index is equal to sample_df['id'] values instead of hard-coding the numbers, which would break if the id column in sample df ever got changed for some reason.

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -1304,3 +1302,78 @@ def test_select_repetitive(sample_df):
...
+    error = 'Specified index column `testing` not found in Schema.'
+    with pytest.raises(LookupError, match=error):
+        sample_df.ww.set_index('testing')
+
+    error = "Index column must be unique"
+    with pytest.raises(LookupError, match=error):
+        sample_df.ww.set_index('age')
+
+
+# def test_set_index(sample_df):
```
thehomebrewnerd, line 1341: Can we delete these old tests now?

woodwork/tests/schema/test_schema.py
```diff
@@ -551,3 +551,65 @@ def test_removes_time_index_via_tags(sample_column_names, sample_inferred_logica
...
+    assert schema.index is None
+    assert schema.semantic_tags['age'] == {'numeric'}
+    assert schema.semantic_tags['id'] == {'numeric'}
+
+
+def test_set_index_errors(sample_column_names, sample_inferred_logical_types):
+    schema = Schema(sample_column_names, sample_inferred_logical_types)
+
+    schema.set_index(None)
+    assert schema.index is None
```
thehomebrewnerd, line 581: Is this related to the test for errors or can it be removed since these two lines are in the preceding test?

thehomebrewnerd (changes requested):

docs/source/api_reference.rst
```diff
@@ -31,6 +41,10 @@ Schema
     :toctree: generated/
 
     Schema
+    Schema.add_semantic_tags
+    Schema.remove_semantic_tags
+    Schema.reset_semantic_tags
+    Schema.set_semantic_tags
```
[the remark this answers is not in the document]
tamargrey: yuppp that would be why the docs are failing to build 🤦‍♀️

woodwork/tests/accessor/test_table_accessor.py
tamargrey: yep! removing

woodwork/tests/schema/test_schema.py
tamargrey: It's more checking that there's no issue with setting the index to None when it's already None... though maybe that belongs in `set_index_twice`

woodwork/table_accessor.py
tamargrey: Sounds good!

woodwork/tests/accessor/test_table_accessor.py
tamargrey: makes sense to me!

woodwork/tests/schema/test_schema.py
thehomebrewnerd: Yeah, seems like it might fit there better.

woodwork/tests/accessor/test_table_accessor.py
tamargrey: I think this test is a fine place to add that check. Makes sense to add

woodwork/tests/schema/test_schema.py
```diff
@@ -551,3 +551,59 @@ def test_removes_time_index_via_tags(sample_column_names, sample_inferred_logica
...
+        schema.set_index('testing')
+
+
+def test_set_index_twice(sample_column_names, sample_inferred_logical_types):
+    schema = Schema(sample_column_names, sample_inferred_logical_types, index='id')
+
+    original_schema = schema._get_subset_schema(list(schema.columns.keys()))
+
+    schema.set_index(None)
+    assert schema.index is None
```
thehomebrewnerd, line 591: This flow is setting the index to `id` on init, then changing to `None`, then changing back to `id`. Is that what you wanted here? I was originally thinking this was confirming there were no issues with setting the index to the same value twice, but I might have misunderstood.

tamargrey: you're totally right! Mistake on my part--thanks for catching.

thehomebrewnerd (approved):
Looks good!

PR #612, opened 2021-02-22
Allow updating time index

### LINKED ISSUE #586: Implement index and time index updates on Table Accessor

tamargrey:
- Adds `set_time_index method that allows for creating, changing, or removing a time index column. Performs relevant checks for whether the column is a valid time index column.
- Closes #586

### DIFF
--- a/woodwork/table_accessor.py
+++ b/woodwork/table_accessor.py
@@ -164,19 +164,6 @@ class WoodworkTableAccessor:
             _check_index(self._dataframe, self._schema.index)
         self._set_underlying_index()
 
-    def set_time_index(self, new_time_index):
-        '''Sets the time index column of the DataFrame. Adds the 'time index' semantic tag to the column
-        and clears the tag from any previously set time index column.
-        Clears the DataFrame's time index by passing in None.
-
-        Args:
-            new_time_index (str): The name of the column to set as the time index
-        '''
-        self._schema.set_time_index(new_time_index)
-
-        if self._schema.time_index is not None:
-            _check_time_index(self._dataframe, self._schema.time_index)
-
     def select(self, include):
         """Create a DataFrame with Woodowork typing information initialized
         that includes only columns whose Logical Type and semantic tags are
--- a/woodwork/table_accessor.py
+++ b/woodwork/table_accessor.py
@@ -153,15 +153,29 @@ class WoodworkTableAccessor:
         and clears the tag from any previously set index column.
         Setting a column as the index column will also cause any previously set standard
         tags for the column to be removed.
+        Clears the DataFrame's index by passing in None.
 
         Args:
-            index (str): The name of the column to set as the index
+            new_index (str): The name of the column to set as the index
         """
         self._schema.set_index(new_index)
 
         if self._schema.index is not None:
             _check_index(self._dataframe, self._schema.index)
         self._set_underlying_index()
+
+    def set_time_index(self, new_time_index):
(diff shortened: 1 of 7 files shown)


woodwork/table_accessor.py
```diff
@@ -153,16 +153,30 @@ def set_index(self, new_index):
...
+        and clears the tag from any previously set time index column.
+        Clears the DataFrame's time index by passing in None.
+
+        Args:
+            new_time_index (str): The name of the column to set as the time index
+        '''
+        self._schema.set_time_index(new_time_index)
+
+        if self._schema.time_index is not None:
+            _check_time_index(self._dataframe, self._schema.time_index)
```
tamargrey, line 178: I'm not 100% convinced that there needs to be a method on the Accessor. The `_check_time_index` in `schema.py` which gets called first, checks whether the column is a Datetime logical type or has the `numeric` standard tag, which I think might be good enough in this case, since we'll have previously confirmed that the logical type being used is valid.

Our checks in the accessor's `_check_time_index` basically make sure that the potential time index column is a valid Datetime or numeric.

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -317,21 +317,16 @@ def test_accessor_with_numeric_time_index(time_index_df):
...
-    # assert date_col.semantic_tags == {'numeric'}
-
-    # Set numeric time index after init
-    # schema = Schema(time_index_df, logical_types={'ints': 'Double'})
-    # schema = schema.set_time_index('ints')
-    # date_col = schema['ints']
-    # assert schema.time_index == 'ints'
-    # assert date_col.logical_type == Double
-    # assert date_col.semantic_tags == {'time_index', 'numeric'}
+    # Change time index to normal datetime time index
```
thehomebrewnerd, line 320: Comment wrong? Should this be `Set numeric time index after init` instead?

woodwork/table_accessor.py
thehomebrewnerd: Makes sense. If we can eliminate this, I'd say let's do it.

woodwork/tests/schema/test_schema.py
```diff
@@ -582,28 +582,73 @@ def test_set_index_errors(sample_column_names, sample_inferred_logical_types):
         schema.set_index('testing')
 
 
+def test_set_time_index(sample_column_names, sample_inferred_logical_types):
+    schema = Schema(sample_column_names, sample_inferred_logical_types)
+    assert schema.time_index is None
+    assert schema.semantic_tags['id'] == {'numeric'}
```
thehomebrewnerd, line 588: Is this assertion needed here?

woodwork/table_accessor.py
tamargrey: Yep, I haven't thought of a reason to keep it. Removed!

woodwork/tests/accessor/test_table_accessor.py
tamargrey: yep! Changing

woodwork/tests/schema/test_schema.py
tamargrey: I don't think so... Removing

thehomebrewnerd (approved):
Looks good. The only other minor thing would be to add this method to the API Reference document for the Schema.

PR #613, opened 2021-02-22
Implement loc and iloc for Column Accessor

### LINKED ISSUE #594: Implement loc and iloc for Column Accessor
Similar to #539, we need to implement `.loc` and `.iloc` on the Column Accessor.

thehomebrewnerd:
- Implement loc and iloc for Column Accessor
- Closes #594 

This PR adds `.loc` and `.iloc` to the ColumnAccessor along with corresponding unit tests for these methods. Code and tests that will be needed for the table implementation of `loc` and `iloc` have been left in place but commented out.

### DIFF
--- a/woodwork/column_accessor.py
+++ b/woodwork/column_accessor.py
@@ -5,6 +5,7 @@ import pandas as pd
 
 from woodwork.accessor_utils import init_series
 from woodwork.exceptions import TypingInfoMismatchWarning
+from woodwork.indexers import _iLocIndexerAccessor, _locIndexerAccessor
 from woodwork.logical_types import LatLong, Ordinal
 from woodwork.schema_column import (
     _add_semantic_tags,
@@ -16,7 +17,6 @@ from woodwork.schema_column import (
     _validate_metadata
 )
 from woodwork.utils import _get_column_logical_type, _is_valid_latlong_series
-from woodwork.indexers import _iLocIndexerAccessor, _locIndexerAccessor
 
 
 @pd.api.extensions.register_series_accessor('ww')
@@ -87,7 +87,6 @@ class WoodworkColumnAccessor:
         """
         return _iLocIndexerAccessor(self._series)
 
-    
     @property
     def loc(self):
         """
--- a/woodwork/column_accessor.py
+++ b/woodwork/column_accessor.py
@@ -16,6 +16,7 @@ from woodwork.schema_column import (
     _validate_metadata
 )
 from woodwork.utils import _get_column_logical_type, _is_valid_latlong_series
+from woodwork.indexers import _iLocIndexerAccessor, _locIndexerAccessor
 
 
 @pd.api.extensions.register_series_accessor('ww')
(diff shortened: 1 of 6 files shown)


woodwork/indexers.py
```diff
@@ -45,3 +45,52 @@ def __getitem__(self, key):
...
+        # else:
+        logical_type = original_data.ww.logical_type or None
+        semantic_tags = original_data.ww.semantic_tags or None
+        # if semantic_tags is not None:
+        #     semantic_tags = semantic_tags - {'index'} - {'time_index'}
+        selection.ww.init(logical_type=logical_type,
+                          semantic_tags=semantic_tags,
+                          use_standard_tags=original_data.ww.use_standard_tags,
+                          description=original_data.ww.description,
+                          metadata=original_data.ww.metadata)
```
tamargrey, line 90: Glad we're adding metadata and description in! I think we hadn't been with DataColumn's but do with DataTable, so it'll be good to be consistent.

woodwork/tests/accessor/test_indexers.py
```diff
@@ -0,0 +1,161 @@
...
+    logical_type = Categorical
+    semantic_tags = ['tag1', 'tag2']
+    description = 'custom column description'
+    metadata = {'meta_key': 'custom metadata'}
+    series.ww.init(logical_type=logical_type,
+                   semantic_tags=semantic_tags,
+                   description=description,
+                   metadata=metadata)
+
+    sliced = series.ww.iloc[2:]
```
tamargrey, line 42: Can we check that changing the schema of `sliced` doesn't also change `series`'  attributes? The ones I'm potentially worried about are `semantic_tags` and `metadata`.

tamargrey (commented):
Looking good so far! Loving the versatility of `_process_selection` for both loc and iloc

woodwork/column_accessor.py
```diff
@@ -67,6 +68,49 @@ def description(self, description):
...
+        return _iLocIndexerAccessor(self._series)
+
+    @property
+    def loc(self):
+        """
+        Access a group of rows and columns by label(s) or a boolean array.
+        ``.loc[]`` is primarily label based, but may also be used with a
+        boolean array.
+        Allowed inputs are:
+        - A single label, e.g. ``5`` or ``'a'``, (note that ``5`` is
```
tamargrey, line 97: Seems like the docs are failing maybe because of the indentation here? We should probably have the format match `iloc`

woodwork/tests/accessor/test_indexers.py
thehomebrewnerd: Yeah, good call. Probably need to add some copy operations. I'll add a test for this.

woodwork/column_accessor.py
thehomebrewnerd: Should be fixed now. I hope.

woodwork/indexers.py
```diff
@@ -45,3 +47,52 @@ def __getitem__(self, key):
...
+        #     return selection
+        # if isinstance(self.ww_data, ww.DataTable):
+        #     logical_type = self.ww_data.logical_types.get(col_name, None)
+        #     semantic_tags = self.ww_data.semantic_tags.get(col_name, None)
+        # else:
+        logical_type = original_data.ww.logical_type or None
+        semantic_tags = original_data.ww.semantic_tags.copy() or None
+        # if semantic_tags is not None:
+        #     semantic_tags = semantic_tags - {'index'} - {'time_index'}
+        selection.ww.init(logical_type=logical_type,
```
tamargrey, line 88: Would it be possible to do what we do in `_make_series_call`?
```python
                        schema = copy.deepcopy(self._schema)
                        # We don't need to pass dtype from the schema to init
                        del schema['dtype']
                        result.ww.init(**schema)
```
I figure lines 84 and 85 may make it not possible here, but thought I'd suggest it

thehomebrewnerd: I was actually thinking it is some of the uncommented code above that will complicate using this approach (when we are pulling a series off a dataframe), but I can take a closer look.

tamargrey: Okay! If we do end up leaving it like this, we'll just need to remember to update it if the parameters to the Series Accessor init change (probably why we don't have description or metadata on the DataColumn tbh)

Maybe we could add a test like `test_get_subset_schema_all_params` that makes sure we test with all possible params to the init.

thehomebrewnerd: @tamargrey I think your original suggestion here is the better approach. Changing for now and will worry about the commented code later when we get there.

Also, making this change highlighted an issue with the `_make_series_call` code when we have `use_standard_tags=False`, so I fixed that as well and added a new test to cover that situation.

woodwork/column_accessor.py
```diff
@@ -67,6 +68,49 @@ def description(self, description):
         _validate_description(description)
         self._schema['description'] = description
 
+    @property
+    def iloc(self):
+        """
+        Integer-location based indexing for selection by position.
+        ``.iloc[]`` is primarily integer position based (from ``0`` to
+        ``length-1`` of the axis), but may also be used with a boolean array.
```
tamargrey, line 76: Would it make sense to have a line in here about how if the result is a Series it has Woodwork typing information still initialized?

tamargrey (approved):
One small comment about the docstrings but other than that looks good!

woodwork/column_accessor.py
thehomebrewnerd: Sure, I can make that update.

Issue #597, opened 2021-02-18
Replicate preserve index behavior implemented on DataTable


thehomebrewnerd:
PR #588 implemented changed to the way indexes are handled for DataTables. We should make sure this behavior is replicated with the new accessor approach.

tamargrey:
We might need to move the semantic tag update methods to the Table Accessor. I stuck them on the Schema class because I was thinking that they don't care at all about the dataframe, but this change means they have to be able to reset the underlying index....

PR #619, opened 2021-02-23
Update set_types to also set logical types

### LINKED ISSUE #587: Implement `set_types` on Table Accessor
This allows for updating logical types on the Table Accessor.

As this will cause changes to the underlying DataFrame, we might want to consider returning a new DataFrame and Schema here instead of doing things inplace

tamargrey:
- Updates `df.ww.set_types` so that it can also take in a logical types parameter
- Updates underlying dtype of DataFrame is logical type is changed
- Also removes underlying index if index has been removed from the Schema 
- Closes #587

### DIFF
--- a/woodwork/schema.py
+++ b/woodwork/schema.py
@@ -155,18 +155,17 @@ class Schema(object):
         return None
 
     def set_types(self, logical_types=None, semantic_tags=None, retain_index_tags=True):
-        """Update the semantic tags for any columns names in the provided types dictionary,
-        updating the Woodwork typing information for the DataFrame. Any columns whose Logical Type
-        or semantic tags are updated will have its existing semantic tags reset.
+        """Update the logical tpye and semantic tags for any columns names in the provided types dictionaries,
+        updating the Schema at those columns.
 
         Args:
-            # --> add ltypes param back in
+            logical_types (dict[str -> LogicalType], optional): A dictionary defining the new logical types for the
+                specified columns.
             semantic_tags (dict[str -> str/list/set], optional): A dictionary defining the new semantic_tags for the
                 specified columns.
             retain_index_tags (bool, optional): If True, will retain any index or time_index
-                semantic tags set on the column. If False, will replace all semantic tags. Defaults to
-                True.
-
+                semantic tags set on the column. If False, will replace all semantic tags any time a column's
+                semantic tags or logical type changes. Defaults to True.
         """
         logical_types = logical_types or {}
         _check_logical_types(self.columns.keys(), logical_types, require_all_cols=False)
--- a/woodwork/schema.py
+++ b/woodwork/schema.py
@@ -176,7 +176,7 @@ class Schema(object):
 
         # Set logical types
         # --> we should confirm they're ltype objects and not strings - but probably below
-        for col_name, column in self.columns.items():
+        for col_name in (logical_types.keys() | semantic_tags.keys()):
             if col_name not in logical_types and col_name not in semantic_tags:
                 continue
 
--- a/woodwork/schema.py
+++ b/woodwork/schema.py
@@ -156,7 +156,8 @@ class Schema(object):
 
     def set_types(self, logical_types=None, semantic_tags=None, retain_index_tags=True):
         """Update the semantic tags for any columns names in the provided types dictionary,
-        updating the Woodwork typing information for the DataFrame.
+        updating the Woodwork typing information for the DataFrame. Any columns whose Logical Type
+        or semantic tags are updated will have its existing semantic tags reset.
 
         Args:
             # --> add ltypes param back in
@@ -173,28 +174,33 @@ class Schema(object):
         semantic_tags = semantic_tags or {}
         _check_semantic_tags(self.columns.keys(), semantic_tags)
 
-        # Set logical types
-        # --> we should confirm they're ltype objects and not strings - but probably below
         for col_name in (logical_types.keys() | semantic_tags.keys()):
-            if col_name not in logical_types and col_name not in semantic_tags:
-                continue
-
             original_tags = self.semantic_tags[col_name]
-            if col_name in logical_types:
-                self.columns[col_name]['logical_type'] = logical_types.get(col_name)
-                # Changing the logical type resets the semantic tags
-                self.reset_semantic_tags(col_name, retain_index_tags=retain_index_tags)
-            if col_name in semantic_tags:
-                new_semantic_tags = _set_semantic_tags(semantic_tags.get(col_name),
+
+            # Update Logical Type for the Schema, getting new semantic tags
+            ltype_update_tags = original_tags
+            new_logical_type = logical_types.get(col_name)
+            if new_logical_type is not None:
+                self.columns[col_name]['logical_type'] = new_logical_type
+                ltype_update_tags = _reset_semantic_tags(new_logical_type.standard_tags, self.use_standard_tags)
+
+            # Get new semantic tags, removing existing tags
+            new_semantic_tags = semantic_tags.get(col_name)
+            if new_semantic_tags is not None:
+                new_semantic_tags = _set_semantic_tags(new_semantic_tags,
                                                        self.logical_types[col_name].standard_tags,
                                                        self.use_standard_tags)
                 _validate_not_setting_index_tags(new_semantic_tags, col_name)
-                self.columns[col_name]['semantic_tags'] = new_semantic_tags
 
-                if retain_index_tags and 'index' in original_tags:
-                    self._set_index_tags(col_name)
-                if retain_index_tags and 'time_index' in original_tags:
-                    self._set_time_index_tags(col_name)
+            # Update the tags for the Schema
+            if new_semantic_tags is None:
+                new_semantic_tags = ltype_update_tags
+            self.columns[col_name]['semantic_tags'] = new_semantic_tags
+
+            if retain_index_tags and 'index' in original_tags:
+                self._set_index_tags(col_name)
+            if retain_index_tags and 'time_index' in original_tags:
+                self._set_time_index_tags(col_name)
 
     def add_semantic_tags(self, semantic_tags):
         """Adds specified semantic tags to columns, updating the Woodwork typing information.
--- a/woodwork/schema.py
+++ b/woodwork/schema.py
@@ -155,7 +155,7 @@ class Schema(object):
         return None
 
     def set_types(self, logical_types=None, semantic_tags=None, retain_index_tags=True):
-        """Update the logical tpye and semantic tags for any columns names in the provided types dictionaries,
+        """Update the logical type and semantic tags for any columns names in the provided types dictionaries,
         updating the Schema at those columns.
 
         Args:
--- a/woodwork/schema.py
+++ b/woodwork/schema.py
@@ -7,6 +7,7 @@ import woodwork as ww
 from woodwork.schema_column import (
     _add_semantic_tags,
     _get_column_dict,
+    _get_column_tags,
     _remove_semantic_tags,
     _reset_semantic_tags,
     _set_semantic_tags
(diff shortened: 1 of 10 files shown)


woodwork/table_accessor.py
```diff
@@ -164,6 +164,37 @@ def set_index(self, new_index):
...
+            logical_types (dict[str -> str], optional): A dictionary defining the new logical types for the
+                specified columns.
+            semantic_tags (dict[str -> str/list/set], optional): A dictionary defining the new semantic_tags for the
+                specified columns.
+            retain_index_tags (bool, optional): If True, will retain any index or time_index
+                semantic tags set on the column. If False, will replace all semantic tags any time a column's
+                semantic tags or logical type changes. Defaults to True.
+        """
+        logical_types = logical_types or {}
+        logical_types = {col_name: _get_column_logical_type(None, ltype, col_name) for col_name, ltype in logical_types.items()}
```
tamargrey, line 181: Felt slightly weird using `_get_column_logical_type` here, since, technically, if `ltype` wasn't defined it'd try to infer the logical type, which we definitely don't want (this is happening in the DataTables' set types as well). To avoid that, I passed `None` in for the series so that we definitely wouldn't quietly infer a LogicalType during `set_types`.

I'd be open to pulling out the logic from this function that parses a logical type user input, but not sure how useful that'd be.

woodwork/table_accessor.py
```diff
@@ -164,6 +164,37 @@ def set_index(self, new_index):
...
+        # go through changed ltypes and update dtype if necessary
+        for col_name, logical_type in logical_types.items():
+            series = self._dataframe[col_name]
+            updated_series = _update_column_dtype(series, logical_type)
+            if updated_series is not series:
+                self._dataframe[col_name] = updated_series
+
+        # if index has been reset, reset underlying index of DataFrame
+        if original_index is not None and self._schema.index is None:
+            self._set_underlying_index()
```
tamargrey, line 196: Felt it was important to follow our behavior of resetting the underlying index when it's removed, but this isn't getting done for `reset_semantic_tags` or `remove_semantic_tags` since they only exist on the Schema class, and this isn't actually behavior that we want (#597 ), so maybe it's better to not have this here at all.

woodwork/tests/schema/test_schema.py
```diff
@@ -309,6 +310,141 @@ def test_get_subset_schema_all_params(sample_column_names, sample_inferred_logic
...
+    semantic_tags = {
+        'full_name': 'tag1',
+        'age': 'test_tag'
+    }
+    schema = Schema(sample_column_names, sample_inferred_logical_types,
+                    index='full_name',
+                    semantic_tags=semantic_tags, use_standard_tags=True)
+
+    # If the tags for a column are None, nothing should change
+    schema.set_types(semantic_tags={'full_name': None}, retain_index_tags=False)
```
tamargrey, line 356: I don't know why a user would do this, but figured it'd be better to just do nothing than to error or still reset. Happy to discuss other options, though

woodwork/schema.py
```diff
@@ -154,27 +154,46 @@ def time_index(self):
...
 
-    def set_types(self, semantic_tags=None, retain_index_tags=True):
-        """Update the semantic tags for any columns names in the provided types dictionary,
-        updating the Woodwork typing information for the DataFrame.
+    def set_types(self, logical_types=None, semantic_tags=None, retain_index_tags=True):
+        """Update the logical tpye and semantic tags for any columns names in the provided types dictionaries,
+        updating the Schema at those columns.
 
         Args:
+            logical_types (dict[str -> LogicalType], optional): A dictionary defining the new logical types for the
```
tamargrey, line 162: I don't know that we want the `Schema.set_types` to only take in LogicalType objects. It follows the way that the Schema only takes in LogicalType objects, but maybe that's less important here.

woodwork/table_accessor.py
thehomebrewnerd: My initial reaction is that we should pull out the logic that parses the logical type into a separate function that we could use here and inside `_get_column_logical_type`. That would make the code a bit more understandable I think rather than trying to understand why `None` is being passed for the series here.

woodwork/table_accessor.py
thehomebrewnerd: TBH the underlying index behavior is getting me a bit confused. I'd vote we have some good discussion about how we really want it to work in all these different scenarios and tackle it separate from this issue.

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -791,10 +790,16 @@ def test_ordinal_with_incomplete_ranking(sample_series):
...
                           "present in the order values provided: ['c']")
+
+    schema_df = pd.DataFrame(sample_series)
+
     with pytest.raises(ValueError, match=error_msg):
-        schema_df = pd.DataFrame(sample_series)
         schema_df.ww.init(logical_types={'sample_series': ordinal_incomplete_order})
 
+    with pytest.raises(ValueError, match=error_msg):
+        schema_df.ww.init()
```
thehomebrewnerd, line 800: Maybe move this line outside the context manager so we know clearly which statement we are testing. Not sure why the df creation was inside before...

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -1350,3 +1355,42 @@ def test_accessor_set_index_errors(sample_df):
     error = "Index column must be unique"
     with pytest.raises(LookupError, match=error):
         sample_df.ww.set_index('age')
+
+
+def test_set_types(sample_df):
```
thehomebrewnerd, line 1360: Should we also test setting a semantic tag somewhere in this test?

woodwork/tests/schema/test_schema_init.py
```diff
@@ -90,6 +90,27 @@ def test_check_logical_types_errors(sample_column_names):
     with pytest.raises(LookupError, match=error_message):
         _check_logical_types(sample_column_names, bad_logical_types_keys)
 
+    bad_logical_types_keys = {
+        'email': 1,
+
```
thehomebrewnerd, line 95: Extra blank line

woodwork/schema.py
tamargrey: Even in the `Schema.__init__`?

I agree that we want to avoid people having to import more LogicalTypes than they have to, but, in theory, they should rarely be directly creating Schemas or calling Schema.set_types, so it probably won't be that big of an issue.

then again, I can't think of a reason that a Schema shouldn't be able to parse logical type strings other than trying to simplify it a bit.

woodwork/table_accessor.py
tamargrey: Cool. I'm happy to pull that logic out! (and also maybe just move it to the Schema.set_types)

woodwork/table_accessor.py
tamargrey: I think that's very fair

tamargrey: Going to remove this behavior so that there's not inconsistency with index removal via tags

woodwork/schema.py
thehomebrewnerd: Sorry - I'm getting confused between the methods. I think it is fine for the Schema methods to only take in LogicalType classes, but the accessor methods should be more flexible.

tamargrey: Ahh, got it. So to confirm: We'll allow string inputs in `WoodworkTableAccessor.set_types` but not for `Schema.set_types`

thehomebrewnerd: Yep, that's what I was thinking.

woodwork/tests/accessor/test_table_accessor.py
tamargrey: Makes sense. I'll update the checks to include one that sets semantic tags as well

woodwork/tests/schema/test_schema.py
tamargrey: I think I'd like to catch the None case in `_check_semantic_tags`--that way it'd also get caught at init.

thehomebrewnerd: Makes sense to me

woodwork/schema.py
```diff
@@ -154,27 +154,46 @@ def time_index(self):
...
+            new_semantic_tags = semantic_tags.get(col_name)
+            if new_semantic_tags is not None:
+                new_semantic_tags = _set_semantic_tags(new_semantic_tags,
+                                                       self.logical_types[col_name].standard_tags,
+                                                       self.use_standard_tags)
+                _validate_not_setting_index_tags(new_semantic_tags, col_name)
+
+            # Update the tags for the Schema
+            if new_semantic_tags is None:
+                new_semantic_tags = ltype_update_tags
```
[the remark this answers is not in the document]
tamargrey: I think we still need some intermediate like `ltype_update_tags` so that `new_semantic_tags` can take precedence when it's present, but if it's `None`, it'll mean that there was an ltype update, so we need `ltype_update_tags` to reset the semantic tags`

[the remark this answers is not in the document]
thehomebrewnerd: Not sure if it is better or not, but this is what I was thinking for the `for` loop:
```python
        for col_name in (logical_types.keys() | semantic_tags.keys()):
            original_tags = self.semantic_tags[col_name]

            # Update Logical Type for the Schema, getting new semantic tags
            new_logical_type = logical_types.get(col_name)
            if new_logical_type is not None:
                self.columns[col_name]['logical_type'] = new_logical_type
                # ltype_update_tags = _reset_semantic_tags(new_logical_type.standard_tags, self.use_standard_tags)

            # Get new semantic tags, removing existing tags
            new_semantic_tags = semantic_tags.get(col_name)
            if new_semantic_tags is not None:
                new_semantic_tags = _set_semantic_tags(new_semantic_tags,
                                                       self.logical_types[col_name].standard_tags,
                                                       self.use_standard_tags)
                _validate_not_setting_index_tags(new_semantic_tags, col_name)
            else:
                new_semantic_tags = _reset_semantic_tags(new_logical_type.standard_tags, self.use_standard_tags)

            self.columns[col_name]['semantic_tags'] = new_semantic_tags

            if retain_index_tags and 'index' in original_tags:
                self._set_index_tags(col_name)
            if retain_index_tags and 'time_index' in original_tags:
                self._set_time_index_tags(col_name)
```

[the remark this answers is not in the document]
thehomebrewnerd: I think that works because we know if `new_semantic_tags` is `None` that means `new_logical_type` is not `None`.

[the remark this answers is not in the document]
tamargrey: Ahh, got it. That'll definitely work, but I think it might be beneficial to keep the actual reset call next to where we update the logical type for clarity.

Maybe if we do `if new_semantic_tags is None` then `new_semantic_tags = _reset_semantic_tags(..)` then we can have the concept of "if we change the logical type the tags get reset" more closely tied together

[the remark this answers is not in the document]
thehomebrewnerd: Yep, sounds good!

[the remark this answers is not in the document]
thehomebrewnerd: As a side note here, related to a prior conversation, I think some of this confusion stems from having a method that is responsible for doing multiple things - if we had this logic completely separated into two methods this wouldn't be quite as confusing.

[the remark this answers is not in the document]
tamargrey: With DataTables, I think we started out with separate methods and combined them to a single `set_types`... maybe @gsheni remembers why??

woodwork/utils.py
```diff
@@ -320,14 +320,19 @@ def get_valid_mi_types():
...
-            return logical_type
-        else:
-            raise TypeError(f"Invalid logical type specified for '{name}'")
+        return _parse_logical_type(logical_type, name)
     else:
         return ww.type_system.infer_logical_type(series)
+
+
+def _parse_logical_type(logical_type, name):
+    # --> pull out parsing logic to own tests
```
thehomebrewnerd, line 329: Old comment?

thehomebrewnerd (approved):
One minor thing, but otherwise looks good to merge after fixing the merge conflicts.

woodwork/utils.py
tamargrey: haha yeppp

PR #625, opened 2021-02-24
Add DaskColumnAccessor

### LINKED ISSUE #611: Add Dask Column Accessor
Related to #493

Allows Dask Series to use the ww namespace for accessors.

thehomebrewnerd:
- Add DaskColumnAccessor
- Closes #611 

This PR implements various changes to add a Dask compatible column accessor.

### DIFF
--- a/woodwork/accessor_utils.py
+++ b/woodwork/accessor_utils.py
@@ -89,3 +89,23 @@ def _update_column_dtype(series, logical_type):
                 f'logical type {logical_type}.'
             raise TypeConversionError(error_msg)
     return series
+
+
+def _is_series(data):
+    if isinstance(data, pd.Series):
+        return True
+    elif dd and isinstance(data, dd.Series):
+        return True
+    elif ks and isinstance(data, ks.Series):
+        return True
+    return False
+
+
+def _is_dataframe(data):
+    if isinstance(data, pd.DataFrame):
+        return True
+    elif dd and isinstance(data, dd.DataFrame):
+        return True
+    elif ks and isinstance(data, ks.DataFrame):
+        return True
+    return False
--- a/woodwork/indexers.py
+++ b/woodwork/indexers.py
@@ -3,6 +3,7 @@ import copy
 import pandas as pd
 
 import woodwork as ww
+from woodwork.accessor_utils import _is_dataframe, _is_series
 from woodwork.utils import _new_dt_including, import_or_none
 
 dd = import_or_none('dask.dataframe')
@@ -93,23 +94,3 @@ def _process_selection(selection, original_data):
         selection.ww.init(schema=new_schema)
     # Selecting a single value or return selection from above
     return selection
-
-
-def _is_series(data):
-    if isinstance(data, pd.Series):
-        return True
-    elif dd and isinstance(data, dd.Series):
-        return True
-    elif ks and isinstance(data, ks.Series):
-        return True
-    return False
-
-
-def _is_dataframe(data):
-    if isinstance(data, pd.DataFrame):
-        return True
-    elif dd and isinstance(data, dd.DataFrame):
-        return True
-    elif ks and isinstance(data, ks.DataFrame):
-        return True
-    return False
--- a/woodwork/indexers.py
+++ b/woodwork/indexers.py
@@ -72,11 +72,11 @@ class _locIndexerAccessor:
 
 
 def _process_selection(selection, original_data):
-    if isinstance(selection, pd.Series) or (ks and isinstance(selection, ks.Series)):
-        if isinstance(original_data, pd.DataFrame) and set(selection.index.values) == set(original_data.columns):
+    if _is_series(selection):
+        if _is_dataframe(original_data) and set(selection.index.values) == set(original_data.columns):
             # Selecting a single row from a DataFrame, returned as Series without Woodwork initialized
             schema = None
-        elif isinstance(original_data, pd.DataFrame):
+        elif _is_dataframe(original_data):
             # Selecting a single column from a DataFrame
             schema = original_data.ww.schema.columns[selection.name]
             schema['semantic_tags'] = schema['semantic_tags'] - {'index'} - {'time_index'}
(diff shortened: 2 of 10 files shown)


woodwork/utils.py
```diff
@@ -267,6 +267,9 @@ def _to_latlong_float(val):
 def _is_valid_latlong_series(series):
     '''Returns True if all elements in the series contain properly formatted LatLong values,
     otherwise returns False'''
+    dd = import_or_none('dask.dataframe')
+    if dd and isinstance(series, dd.Series):
+        series = series.compute()
```
thehomebrewnerd, line 272: Not sure if we want to pull these values into memory to check them, or if we would be better off to skip this validation for Dask? It can always be addressed in a future issue if this is problematic.

woodwork/tests/accessor/test_column_accessor.py
```diff
@@ -380,40 +382,43 @@ def test_remove_semantic_tags_raises_error_with_invalid_tag(sample_series):
...
+    assert replace_series.ww._schema == series.ww._schema
+    assert replace_series.ww._schema is not series.ww._schema
+    pd.testing.assert_series_equal(to_pandas(replace_series), to_pandas(series.replace('a', 'd')))
 
 
 def test_series_methods_on_accessor_inplace(sample_series):
-    xfail_dask_and_koalas(sample_series)
+    # TODO: Try to find a supported inplace method for Dask, if one exists
+    if dd and isinstance(sample_series, dd.Series):
+        pytest.xfail('Dask does not support pop.')
```
thehomebrewnerd, line 420: Dask generally does not support inplace operations - not sure if there is something else we can use here instead, or if we just leave it like this.

woodwork/indexers.py
```diff
@@ -86,10 +86,30 @@ def _process_selection(selection, original_data):
...
+    elif _is_dataframe(selection):
         # Selecting a new DataFrame from an existing DataFrame
         schema = original_data.ww.schema
         new_schema = schema._get_subset_schema(list(selection.columns))
         selection.ww.init(schema=new_schema)
     # Selecting a single value or return selection from above
     return selection
+
+
+def _is_series(data):
```
tamargrey, line 98: nifty! This might be useful in the column_accessor's `_make_series_call` (and the dataframe one eventually with the `_make_dataframe_call` in the table accessor)

woodwork/tests/accessor/test_column_accessor.py
tamargrey: that's interesting... we were able to have it in DataTables I guess since we were actually creating a new DataFrame object and dropping the relevant column.

But I'm fine leaving it like this if Dask itself doesn't support inplace operations

thehomebrewnerd: Yeah, I think we got away with it in DataTables since we weren't actually using a `pop` call on the underlying dataframe, and as such we weren't attempting any inplace modifications.

woodwork/indexers.py
thehomebrewnerd: The `if` conditions in `_process_selection` were getting really messy without this. Seems like it is more readable this way.

thehomebrewnerd: And, yeah ,we could add this to utils and use it elsewhere if that makes sense.

woodwork/utils.py
gsheni: Let's create an issue to revisit this later on. We can leave compute as is.

woodwork/indexers.py
```diff
@@ -72,11 +72,11 @@ def __getitem__(self, key):
 
 
 def _process_selection(selection, original_data):
-    if isinstance(selection, pd.Series) or (ks and isinstance(selection, ks.Series)):
-        if isinstance(original_data, pd.DataFrame) and set(selection.index.values) == set(original_data.columns):
+    if _is_series(selection):
```
gsheni, line 75: Nice, this looks much cleaner!

woodwork/tests/accessor/test_column_accessor.py
```diff
@@ -440,20 +445,28 @@ def test_series_methods_on_accessor_returning_series_invalid_schema(sample_serie
 
 
 def test_series_methods_on_accessor_other_returns(sample_series):
-    xfail_dask_and_koalas(sample_series)
+    xfail_koalas(sample_series)
     series = sample_series.astype('category')
     series.ww.init()
     col_shape = series.ww.shape
     series_shape = series.shape
+    if dd and isinstance(series, dd.Series):
```
gsheni, line 453: You could create a helper function
```python
def _is_dask_series(data):
    if dd and isinstance(data, dd.Series):
        return True
    return False
```

woodwork/utils.py
thehomebrewnerd: Created #627 for this and added it to the Accessor Improvements epic

woodwork/tests/accessor/test_column_accessor.py
thehomebrewnerd: Yeah, we've got these calls all throughout the code as well as a dataframe version (and will also have them for Koalas). Maybe we should clean this up in a separate issue?

gsheni: Sure, can go in the improvements epic

thehomebrewnerd: Created #628  for this update

woodwork/indexers.py
thehomebrewnerd: Moved these to accessor_utils and updated _make_series_call to use _is_series.

woodwork/tests/accessor/test_accessor_utils.py
```diff
@@ -56,8 +59,20 @@ def test_init_series_all_parameters(sample_series):
...
+    xfail_koalas(sample_series)
+    if dd and isinstance(sample_series, dd.Series):
+        pytest.xfail('Dask type conversion with astype does not fail until compute is called')
     error_message = "Error converting datatype for sample_series from type object to type Int64. " \
         "Please confirm the underlying data is consistent with logical type Integer."
     with pytest.raises(TypeConversionError, match=error_message):
         init_series(sample_series, logical_type='integer')
+
+
+def test_is_series(sample_series, sample_df):
```
tamargrey, line 71: I seem to remember needing to parameterize tests differently when they take in more than one fixture so they don't double up weirdly. 

For example--this produces 9 tests since it's getting every df-series pair. Should we do that here as well? Maybe just grab a single column from `sample_df` for the series check?

thehomebrewnerd: Yes, you are correct. I think we also have other tests like this that need to be fixed. Probably should create an issue for that. For now, I like your idea to just use sample_df and pull a series off of that.

woodwork/accessor_utils.py
```diff
@@ -89,3 +89,23 @@ def _update_column_dtype(series, logical_type):
                 f'logical type {logical_type}.'
             raise TypeConversionError(error_msg)
     return series
+
+
+def _is_series(data):
```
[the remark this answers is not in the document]
thehomebrewnerd: I'll create an issue for this.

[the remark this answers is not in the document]
thehomebrewnerd: Created #631

woodwork/tests/accessor/test_accessor_utils.py
thehomebrewnerd: I did a quick review for other tests that used multiple fixtures and didn't see the one I was thinking of...either I missed it or I was confused about other tests using this pattern. Regardless, I updated this test and if I come across others that use multiple fixtures that would create a matrix of tests unnecessarily, I'll create a new issue.

tamargrey: Sounds good!

tamargrey (approved):

Issue #604, opened 2021-02-22
Documentation - Review and update docstrings and API Reference


thehomebrewnerd:
Review all new accessor docstrings for accuracy/completeness after implementation is complete. Review all methods and helper functions to confirm which should be public and which should be private. Make sure all public methods/functions are included in the API Reference, and make sure the methods are in alphabetical order.

As part of this, it would also be good to review all code comments for accuracy and update or delete as appropriate.

thehomebrewnerd:
Closed by #660

PR #626, opened 2021-02-24
Deserialize accessor from csv, pickle, and parquet to Woodwork table

### LINKED ISSUE #617: Implement deserialization to WoodworkTableAccessor
This issue implements the deserialization method to Woodwork table accessor

tamargrey:
- creates `deserialize_accessor.py` and updates tests in `test_accessor_serialization.py` to test serialization and deserialization
- Closes #617

### DIFF
--- a/woodwork/deserialize_accessor.py
+++ b/woodwork/deserialize_accessor.py
@@ -19,10 +19,10 @@ def read_table_typing_information(path):
     '''Read Woodwork typing information from disk, S3 path, or URL.
 
         Args:
-            path (str): Location on disk, S3 path, or URL to read `table_description.json`.
+            path (str): Location on disk, S3 path, or URL to read `woodwork_typing_info.json`.
 
         Returns:
-            description (dict) : Woodwork typing information
+            dict: Woodwork typing information dictionary
     '''
     path = os.path.abspath(path)
     assert os.path.exists(path), '"{}" does not exist'.format(path)
@@ -33,7 +33,7 @@ def read_table_typing_information(path):
     return typing_info
 
 
-def typing_information_to_woodwork_table(table_typing_info, **kwargs):
+def _typing_information_to_woodwork_table(table_typing_info, **kwargs):
     '''Deserialize Woodwork table from table description.
 
     Args:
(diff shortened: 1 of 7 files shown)


woodwork/deserialize_accessor.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+def read_table_typing_information(path):
+    '''Read Woodwork typing information from disk, S3 path, or URL.
+
+        Args:
+            path (str): Location on disk, S3 path, or URL to read `table_description.json`.
+
+        Returns:
+            description (dict) : Woodwork typing information
```
thehomebrewnerd, line 25: I don't think we need to name the return object so this could just be `dict: Woodwork typing information dictionary`

woodwork/deserialize_accessor.py
```diff
@@ -0,0 +1,176 @@
...
+
+def read_woodwork_table(path, profile_name=None, **kwargs):
+    '''Read Woodwork table from disk, S3 path, or URL.
+
+        Args:
+            path (str): Directory on disk, S3 path, or URL to read `table_description.json`.
+            profile_name (str, bool): The AWS profile specified to write to S3. Will default to None and search for AWS credentials.
+                Set to False to use an anonymous profile.
+            kwargs (keywords): Additional keyword arguments to pass as keyword arguments to the underlying deserialization method.
+    '''
```
thehomebrewnerd, line 137: Maybe we should add a return type here as well, same as  in `typing_information_to_woodwork_table`, since this is part of the public API.

woodwork/tests/datatable/test_accessor_serialization.py
```diff
@@ -172,226 +175,241 @@ def test_to_csv(sample_df, tmpdir):
...
 
 
-# def make_public(s3_client, s3_bucket):
-#     obj = list(s3_bucket.objects.all())[0].key
-#     s3_client.ObjectAcl(BUCKET_NAME, obj).put(ACL='public-read-write')
+def test_to_parquet(sample_df, tmpdir):
+    xfail_dask_and_koalas(sample_df)
 
+    sample_df.ww.init(index='id')
+    print(sample_df.ww.index)
```
thehomebrewnerd, line 236: Remove

thehomebrewnerd (changes requested):
We should also add all the appropriate serialization/deserialization methods and functions to the API Reference.

woodwork/deserialize_accessor.py
tamargrey: Makes sense!

tamargrey: makes sense to me! 

Now I'm thinking about which of these functions should be public, though. 

Having all of them present feels like it detracts from the useful ones (though `read_woodwork_table` is the only one that should get used with any regularity anyway, so maybe they all should be private but that one).

woodwork/deserialize_accessor.py
```diff
@@ -0,0 +1,176 @@
...
+from woodwork.s3_utils import get_transport_params, use_smartopen
+from woodwork.serialize_accessor import FORMATS, SCHEMA_VERSION
+from woodwork.utils import _is_s3, _is_url, import_or_raise
+
+
+def read_table_typing_information(path):
+    '''Read Woodwork typing information from disk, S3 path, or URL.
+
+        Args:
+            path (str): Location on disk, S3 path, or URL to read `table_description.json`.
```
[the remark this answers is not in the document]
tamargrey: changed this and another usage below!

woodwork/deserialize_accessor.py
thehomebrewnerd: Yeah, I think `read_woodwork_table` is the most important one. We could review all our functions and methods as part of #604 though.

thehomebrewnerd (approved):
LGTM!

PR #633, opened 2021-02-25
Adds __getitem__ to table accessor

### LINKED ISSUE #525: Add __getitem__ to Table Accessor
This issue completely implements the __getitem__ to table accessor

jeff-hernandez:
Closes #525 by adding `__getitem__` to table accessor


thehomebrewnerd (changes requested):
@jeff-hernandez I think we need to build this out a bit more and replicate the current tests we have in place for DataTable `__getitem__`. We want to make sure this works for a variety of different input types.

For reference you can take a look at the DataTable implementation here: https://github.com/alteryx/woodwork/blob/c3f73aebca99ebdea5506b30b6d19a5dc99b2e60/woodwork/datatable.py#L139-L147

And you can see the current tests that are in place here: https://github.com/alteryx/woodwork/blob/c3f73aebca99ebdea5506b30b6d19a5dc99b2e60/woodwork/tests/datatable/test_datatable_get_set.py#L20-L110

tamargrey:
> For reference you can take a look at the DataTable implementation

One thing that'll be new is that instead of this line `return self.columns[key] ` which would return a DataColumn, we'll want the single column to return a series with the Woodwork typing info initialized.

docs/source/release_notes.rst
```diff
@@ -32,6 +32,7 @@ Release Notes
         * Add DaskColumnAccessor (:pr:`625`)
         * Allow deserialization from csv, pickle, and parquet to Woodwork table (:pr:`626`)
         * Add ``value_counts`` to WoodworkTableAccessor (:pr:`632`)
+        * Adds ``__getitem__`` to table accessor (:pr:`633`)
```
gsheni, line 35:
```suggestion
        * Adds ``__getitem__`` to WoodworkTableAccessor (:pr:`633`)
```

woodwork/table_accessor.py
```diff
@@ -145,6 +145,30 @@ def __getattr__(self, attr):
...
+            df.ww.init(schema=schema)
+            return df
+
+        if key not in self._dataframe:
+            raise KeyError(f"Column with name '{key}' not found in DataTable")
+
+        series = self._dataframe[key]
+        column = self._dataframe.ww.columns[key].copy()
+        del column['dtype']
+        series.ww.init(**column)
```
tamargrey, line 169: we'll need to pass in `use_standard_tags=self.use_standard_tags` here as well as make a deepcopy of the column dict so that the semantic tags match and we can mutuate the series' typing info without changing the original.

woodwork/table_accessor.py
```diff
@@ -145,6 +145,30 @@ def __getattr__(self, attr):
...
+            df = self._dataframe[key]
+            schema = self._schema._get_subset_schema(key)
+            df.ww.init(schema=schema)
+            return df
+
+        if key not in self._dataframe:
+            raise KeyError(f"Column with name '{key}' not found in DataTable")
+
+        series = self._dataframe[key]
+        column = self._dataframe.ww.columns[key].copy()
```
tamargrey, line 167: We can get the column with `self._schema.columns[key]` (though it's super interesting how we can access the accessor's attrs within the accessor like that--kind of mind bending)

woodwork/table_accessor.py
```diff
@@ -145,6 +145,30 @@ def __getattr__(self, attr):
...
+            columns = self._dataframe.columns
+            in_columns = index.isin(columns).all()
+
+            if not in_columns:
+                diff = list(index.difference(columns).sort_values())
+                raise KeyError(f"Column(s) '{diff}' not found in DataTable")
+
+            df = self._dataframe[key]
+            schema = self._schema._get_subset_schema(key)
+            df.ww.init(schema=schema)
```
tamargrey, line 160: I think `self._get_subset_df_with_schema(key)` should work here

woodwork/table_accessor.py
```diff
@@ -145,6 +145,30 @@ def __getattr__(self, attr):
...
             raise AttributeError(f"Woodwork has no attribute '{attr}'")
 
+    def __getitem__(self, key):
+        if isinstance(key, list):
+            index = pd.Index(key)
+            columns = self._dataframe.columns
+            in_columns = index.isin(columns).all()
+
+            if not in_columns:
+                diff = list(index.difference(columns).sort_values())
```
tamargrey, line 155: If you move `diff`'s definition out of the if statement, you can then check `if diff` before raising the error, I think, which could be good if there's a lot of columns so we don't have to check twice for `in_columns` and then `diff`

woodwork/table_accessor.py
```diff
@@ -145,6 +146,27 @@ def __getattr__(self, attr):
         else:
             raise AttributeError(f"Woodwork has no attribute '{attr}'")
 
+    def __getitem__(self, key):
+        if isinstance(key, list):
+            index = pd.Index(key)
+            columns = self._dataframe.columns
+            diff = index.difference(columns)
+            in_columns = diff.isin(columns).all()
```
tamargrey, line 154: Anything in `diff` shouldn't be in columns, so the only way in_columns is True is if `diff` is empty, so at line 156, it can just be
```python
if diff:
    raise KeyError(...)
```

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -223,6 +223,42 @@ def test_accessor_getattr(sample_df):
...
 
 
+def test_getitem(sample_df):
+    xfail_koalas(sample_df)
+    df = sample_df.copy()
+    df.ww.init(time_index='signup_date', index='id', name='dt_name')
+    assert list(df.columns) == list(df.ww.schema.columns)
+
+    subset = ['id', 'email', 'signup_date']
+    subset_ww = df.ww[subset].ww
```
tamargrey, line 233: It's probably worth checking that the contents of the subset DataFrames and Series' are as expected in addition to the Schema checks, so we might as well get `subset_df = df.ww[subset]` so we can confirm `subset_df` is the same as `df[subset]`.

Plus then we can access the index and whatnot with `subset_df.ww.index`

tamargrey (commented):
Looking good! Two additional things that will probably be good to test are `df.ww['id']` so we can confirm that the index tag gets passed down to that column and a subset that has columns but none of the index columns `df.ww[['age','email']]` to confirm the resulting Schema doesn't have an index or time index

woodwork/table_accessor.py
```diff
@@ -145,6 +146,27 @@ def __getattr__(self, attr):
...
+
+            if not in_columns:
+                raise KeyError(f"Column(s) '{list(diff)}' not found in DataTable")
```
thehomebrewnerd, line 157: Thinking maybe it would be good to be consistent with other places where we do checks like this and use sets to identify differences between the supplied values and the columns in the DataFrame. 

Existing code for example: https://github.com/alteryx/woodwork/blob/f18395909c57061415242b8b65cabed99477bd5c/woodwork/table_accessor.py#L736-L741

Not sure which is most efficient...

thehomebrewnerd, line 157: `DataTable` -> `DataFrame`

woodwork/table_accessor.py
```diff
@@ -145,6 +146,27 @@ def __getattr__(self, attr):
...
+
+            return self._get_subset_df_with_schema(key)
+
+        if key not in self._dataframe:
+            raise KeyError(f"Column with name '{key}' not found in DataTable")
```
thehomebrewnerd, line 162: `DataTable` -> `DataFrame`

woodwork/table_accessor.py
```diff
@@ -145,6 +146,27 @@ def __getattr__(self, attr):
...
+
+        series = self._dataframe[key]
+        column = deepcopy(self._schema.columns[key])
```
thehomebrewnerd, line 165: If you use the `self.schema` property here instead of `self._schema` we should be able to eliminate the deepcopy and the associated import.

woodwork/table_accessor.py
thehomebrewnerd: Another note, if we do use sets, we should probably sort the list of columns that gets displayed in the error message to make sure they are always in the same order.

thehomebrewnerd (commented):
Just a few more things to consider.

woodwork/table_accessor.py
gsheni: Can we use `ColumnNotPresentError` error?
- https://github.com/alteryx/woodwork/blob/f18395909c57061415242b8b65cabed99477bd5c/woodwork/exceptions.py#L47
- We can modify that custom Exception to have an argument, something like this:

```python
class ColumnNotPresentError(KeyError):
    def __init__(self, column):
        message = "Column with name '{key}' not found in DataTable".format(key=column)
        super().__init__(self.message)
```
- Then raise the custom exception like this
```python
    raise ColumnNotPresentError(key)
```

woodwork/tests/accessor/test_table_accessor.py
jeff-hernandez: ✅

woodwork/table_accessor.py
jeff-hernandez: ✅   0901d4d

jeff-hernandez: ✅  df1aa1e

woodwork/table_accessor.py
jeff-hernandez: ✅  c488348

woodwork/table_accessor.py
jeff-hernandez: ✅  479d5a2

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -206,6 +207,54 @@ def test_accessor_getattr(sample_df):
...
+
+    series = df.ww['age']
+    pd.testing.assert_series_equal(to_pandas(series), to_pandas(df['age']))
+    assert series.ww.logical_type == Integer
+    assert series.ww.semantic_tags == {'numeric'}
+
+    series = df.ww['id']
+    pd.testing.assert_series_equal(to_pandas(series), to_pandas(df['id']))
+    assert series.ww.logical_type == Integer
+    assert series.ww.semantic_tags == {'index', 'numeric'}
```
tamargrey, line 242: @thehomebrewnerd I just realized we do `schema['semantic_tags'] = schema['semantic_tags'] - {'index'} - {'time_index'}` on iloc and loc so that the index tags don't propagate to the series. 

We should probably have that here as well when we get a series from the dataframe,  right?

thehomebrewnerd: The `index` and `time_index` tags are not meaningful on a Series outside the context of a DataFrame, so I think it makes sense to remove them.

woodwork/table_accessor.py
```diff
@@ -458,7 +478,7 @@ def _get_subset_df_with_schema(self, cols_to_include):
         retaining all typing information and maintaining the DataFrame's column order.
         '''
         assert all([col_name in self._schema.columns for col_name in cols_to_include])
-        cols_to_include = [col_name for col_name in self._dataframe.columns if col_name in cols_to_include]
+        cols_to_include = [col_name for col_name in cols_to_include if col_name in self._dataframe.columns]
```
tamargrey, line 481: Was the reason for switching this so that the order of the columns of a list input to `__getitem__` would be maintained in the new dataframe?

It's causing column order issues with `test_select_ltypes_no_match_and_all`, because, in that case, we want the column order to match the DataFrame's order....

I'm thinking it might be best to go back to the original way you were creating the new DataFrame with Woodwork so that we can use this method for maintaining column order. Or we could add a param for `use_dataframe_order` or something so that we can decide when to follow DataFrame vs input order. 

Sorry for the confusion here!!

woodwork/tests/accessor/test_table_accessor.py
jeff-hernandez: We want to update this test so that the semantic tags in the series only contain numeric. Is that correct?

woodwork/table_accessor.py
jeff-hernandez: Yes, that was the reason for the switch. Okay, I'll look into the param option.

woodwork/tests/accessor/test_table_accessor.py
tamargrey: Yep!

woodwork/table_accessor.py
jeff-hernandez: ✅ 3d0e771

woodwork/tests/accessor/test_table_accessor.py
jeff-hernandez: ✅ 80481cb

woodwork/table_accessor.py
```diff
@@ -473,16 +473,19 @@ def wrapper(*args, **kwargs):
         # Directly return non-callable DataFrame attributes
         return dataframe_attr
 
-    def _get_subset_df_with_schema(self, cols_to_include):
+    def _get_subset_df_with_schema(self, cols_to_include, use_dataframe_order=True):
```
thehomebrewnerd, line 476: I think we should add a specific test that verifies this parameter produces the desired ordering based on the `use_datafarme_order` value. 

It can be a simpler version of this test, focusing on just verifying the column ordering: https://github.com/alteryx/woodwork/blob/f18395909c57061415242b8b65cabed99477bd5c/woodwork/tests/accessor/test_table_accessor.py#L982

thehomebrewnerd (commented):
Looking pretty good, just one more suggestion.

woodwork/table_accessor.py
jeff-hernandez: ✅ bf3f23f

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -206,6 +207,54 @@ def test_accessor_getattr(sample_df):
         sample_df.ww.not_present
 
 
+def test_getitem(sample_df):
```
thehomebrewnerd, line 210: One thing I notice in looking at this test again, is that we are not verifying that any user types set during init are carried through to the final subset dataframe. We can modify this test slightly to confirm this happens. My suggestion would be set the logical type of Age to `Double` on init and also add a custom semantic tag, something like `custom_tag`, and then update the two tests that include age to check that these values are set correctly on the subset.

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -206,6 +207,54 @@ def test_accessor_getattr(sample_df):
...
+    df = sample_df.copy()
```
thehomebrewnerd, line 211: I don't think we need these copy statements in the new tests, since we aren't modifying `sample_df` during the test.  If you want to leave df = sample_df, that's fine as it makes the statements that follow a little cleaner, or you could remove this completely and just use `sample_df` in place of `df`.

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -206,6 +207,54 @@ def test_accessor_getattr(sample_df):
...
+    assert series.ww.semantic_tags == {'numeric'}
+
+    series = df.ww['id']
+    pd.testing.assert_series_equal(to_pandas(series), to_pandas(df['id']))
+    assert series.ww.logical_type == Integer
+    assert series.ww.semantic_tags == {'numeric'}
+
+
+def test_getitem_invalid_input(sample_df):
+    df = sample_df.copy()
```
thehomebrewnerd, line 246: Same as above

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -1019,6 +1068,18 @@ def test_get_subset_df_with_schema(sample_df):
     validate_subset_schema(transfer_schema.ww.schema, schema)
 
 
+def test_use_dataframe_order(sample_df):
+    df = sample_df.copy()
```
thehomebrewnerd, line 1072: Same as above

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -1019,6 +1068,18 @@ def test_get_subset_df_with_schema(sample_df):
     validate_subset_schema(transfer_schema.ww.schema, schema)
 
 
+def test_use_dataframe_order(sample_df):
```
thehomebrewnerd, line 1071: Maybe update this test name to `test_get_subset_df_use_dataframe_order` to make it more clear what this test is about.

thehomebrewnerd (commented):
This is looking pretty good. I think there are a couple minor things in the tests that we could tweak.

woodwork/tests/accessor/test_table_accessor.py
jeff-hernandez: ✅  7ac4fa4

woodwork/tests/accessor/test_table_accessor.py
jeff-hernandez: ✅  470d855

woodwork/tests/accessor/test_table_accessor.py
jeff-hernandez: ✅  8126f37

thehomebrewnerd (approved):
LGTM!

PR #634, opened 2021-02-25
Implement KoalasColumnAccessor

### LINKED ISSUE #609: Add Koalas Column Accessor
Related to #493

Allows Koalas Series to use the ww namespace for accessors.

thehomebrewnerd:
- Implement KoalasColumnAccessor
- Closes #609

This PR adds a KoalasColumnAccessor which allows Woodwork to be used with koalas.Series objects. Also adds a utility function for determining the valid dtype for a series based on the logical type and whether or not the series is pandas/dask or koalas.

### DIFF
--- a/woodwork/column_accessor.py
+++ b/woodwork/column_accessor.py
@@ -222,9 +222,11 @@ class WoodworkColumnAccessor:
         valid_dtype = _get_valid_dtype(self._series, logical_type)
         if valid_dtype != str(self._series.dtype):
             if ks and isinstance(self._series, ks.Series) and logical_type.backup_dtype:
+                # Koalas will have a dtype of `object` even after `ks.Series.astype('str')` but we want to inform
+                # users to try to convert to the backup dtype not the dtype considered valid for the series
                 convert_dtype = logical_type.backup_dtype
             else:
-                convert_dtype = logical_type.pandas_dtype
+                convert_dtype = valid_dtype
             raise ValueError(f"Cannot initialize Woodwork. Series dtype '{self._series.dtype}' is "
                              f"incompatible with {logical_type} dtype. Try converting series "
                              f"dtype to '{convert_dtype}' before initializing or use the "
--- a/woodwork/column_accessor.py
+++ b/woodwork/column_accessor.py
@@ -3,7 +3,7 @@ import warnings
 
 import pandas as pd
 
-from woodwork.accessor_utils import init_series
+from woodwork.accessor_utils import _is_series, init_series
 from woodwork.exceptions import TypingInfoMismatchWarning
 from woodwork.indexers import _iLocIndexerAccessor, _locIndexerAccessor
 from woodwork.logical_types import LatLong, Ordinal
@@ -16,10 +16,15 @@ from woodwork.schema_column import (
     _validate_description,
     _validate_metadata
 )
-from woodwork.utils import _get_column_logical_type, _is_valid_latlong_series
+from woodwork.utils import (
+    _get_column_logical_type,
+    _is_valid_latlong_series,
+    import_or_none
+)
+
+dd = import_or_none('dask.dataframe')
 
 
-@pd.api.extensions.register_series_accessor('ww')
 class WoodworkColumnAccessor:
     def __init__(self, series):
         self._series = series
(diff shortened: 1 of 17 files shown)


woodwork/column_accessor.py
```diff
@@ -217,10 +219,15 @@ def wrapper(*args, **kwargs):
     def _validate_logical_type(self, logical_type):
         """Validates that a logical type is consistent with the series dtype. Performs additional type
         specific validation, as required."""
-        if logical_type.pandas_dtype != str(self._series.dtype):
+        valid_dtype = _get_valid_dtype(self._series, logical_type)
+        if valid_dtype != str(self._series.dtype):
+            if ks and isinstance(self._series, ks.Series) and logical_type.backup_dtype:
+                convert_dtype = logical_type.backup_dtype
+            else:
+                convert_dtype = logical_type.pandas_dtype
```
thehomebrewnerd, line 227: This is needed because of a Koalas quirk. If you run `astype('str')` on a koalas series you will get back a Series with `object` dtype. However, if you run `.astype('object')` on a Koalas series you will get an error. 

In the case that we are using Koalas and `valid_dtype` ends up being `object` we want to tell the user to try to run `astype('str')`, corresponding to the `backup_dtype` defined on the LogcialType, so we can't just use the value of `valid_dtype` in the error message.

woodwork/tests/testing_utils/column_utils.py
```diff
@@ -0,0 +1,12 @@
+from woodwork.utils import import_or_none
+
+ks = import_or_none('databricks.koalas')
+
+
+def convert_series(series, logical_type):
```
thehomebrewnerd, line 6: Added this function because for our column accessor tests we need to convert sample series to type `category` for pandas/dask or `object` for Koalas. Once we have completely eliminated DataTable and DataColumn, we might just be able to update our test fixtures to do this instead and remove this function and all the conversion calls in the test.

woodwork/column_accessor.py
tamargrey: Okay, so if I'm understanding: We define some backup dtypes as `str` but the way the dtype is stored is as `object`, so when checking the if the underlying dtype would have to change to init the Woodwork typing info, we need to look for `object`, but the actual conversion is still to `str` so that's what we want to tell users about?

I see we have a comment in `_get_valid_dtype` about this. Can we add one here as well?

thehomebrewnerd: Yes, you have understood correctly. I can add a comment here as well.

woodwork/tests/testing_utils/column_utils.py
tamargrey: yeah, I think updating the fixtures to do this later makes sense!

woodwork/utils.py
```diff
@@ -268,17 +268,27 @@ def _is_valid_latlong_series(series):
...
+    ks = import_or_none('databricks.koalas')
     if dd and isinstance(series, dd.Series):
         series = series.compute()
-    if series.apply(_is_valid_latlong_value).all():
+    if ks and isinstance(series, ks.Series):
+        series = series.to_pandas()
+        series_type = 'koalas'
+    else:
+        series_type = 'pandas'
+    if series.apply(_is_valid_latlong_value, args=(series_type,)).all():
```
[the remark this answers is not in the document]
thehomebrewnerd: Good idea. I'll change!

woodwork/utils.py
```diff
@@ -268,17 +268,23 @@ def _is_valid_latlong_series(series):
...
     return False
 
 
-def _is_valid_latlong_value(val):
-    '''Returns True if the value provided is a properly formatted LatLong value, otherwise
-    returns False.'''
-    if isinstance(val, tuple) and len(val) == 2:
+def _is_valid_latlong_value(val, bracket_type=tuple):
+    '''Returns True if the value provided is a properly formatted LatLong value for a
+    pandas or Dask Series, otherwise returns False.'''
```
tamargrey, line 286: should it now be `for a pandas, Dask, or Koalas Series`?

tamargrey (approved):
Small docstring comment, but other than that looks good!

woodwork/utils.py
thehomebrewnerd: Yes, good catch!

PR #640, opened 2021-02-26
Add drop to WoodworkTableAccessor

### LINKED ISSUE #615: Add WoodworkTableAccessor.drop
- Creates a new DataFrame without the specified columns that has Woodwork initialized using the typing information from the original DataFrame

tamargrey:
- Allows for `df.ww.drop`, returning a new dataframe with Woodwork initialized
- Closes #615

### DIFF
--- a/woodwork/table_accessor.py
+++ b/woodwork/table_accessor.py
@@ -468,7 +468,8 @@ class WoodworkTableAccessor:
 
         Note:
             This method is used for removing columns only. To remove rows with ``drop``, go through the
-            DataFrame directly instead of calling ``DataFrame.ww.drop``.
+            DataFrame directly and then reinitialize Woodwork with ``DataFrame.ww.init``
+            instead of calling ``DataFrame.ww.drop``.
         """
         if not isinstance(columns, (list, set)):
             columns = [columns]
--- a/woodwork/table_accessor.py
+++ b/woodwork/table_accessor.py
@@ -457,6 +457,32 @@ class WoodworkTableAccessor:
 
         return series
 
+    def drop(self, columns):
+        """Drop specified columns from a DataFrame.
+
+        Args:
+            columns (str or list[str]): Column name or names to drop. Must be present in the DataTable.
+
+        Returns:
+            (pd.DataFrame): DataFrame with the specified columns removed, maintaining Woodwork typing information.
+
+        Note:
+            This method is used for removing columns only. To remove rows with ``drop``, go through the
+            DataFrame directly instead of calling ``DataFrame.ww.drop``.
+        """
+        if not isinstance(columns, (list, set)):
+            columns = [columns]
+
+        not_present = [col for col in columns if col not in self._dataframe.columns]
+        if not_present:
+            raise ValueError(f'{not_present} not found in DataFrame')
+
+        new_schema = self._schema._get_subset_schema([col for col in self._dataframe.columns if col not in columns])
+        new_df = self._dataframe.drop(columns, axis=1)
+        new_df.ww.init(schema=new_schema)
+
+        return new_df
+
     def mutual_information_dict(self, num_bins=10, nrows=None):
         """
         Calculates mutual information between all pairs of columns in the DataFrame that
--- a/woodwork/tests/accessor/test_table_accessor.py
+++ b/woodwork/tests/accessor/test_table_accessor.py
@@ -721,13 +721,6 @@ def test_underlying_index(sample_df):
     assert schema_df.index.name is None
     assert type(schema_df.index) == specified_index
 
-    # --> add back when we figure out how to handle removing indices
-    # schema_dropped = schema.drop('made_index')
-    # assert 'made_index' not in schema_dropped.columns
-    # assert 'made_index' not in schema_dropped._dataframe.columns
-    # assert type(schema_dropped._dataframe.index) == unspecified_index
-    # assert type(schema_dropped.to_dataframe().index) == unspecified_index
-
 
 def test_accessor_already_sorted(sample_unsorted_df):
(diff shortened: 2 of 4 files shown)


woodwork/table_accessor.py
```diff
@@ -457,6 +457,32 @@ def pop(self, column_name):
...
+
+        Args:
+            columns (str or list[str]): Column name or names to drop. Must be present in the DataTable.
+
+        Returns:
+            (pd.DataFrame): DataFrame with the specified columns removed, maintaining Woodwork typing information.
+
+        Note:
+            This method is used for removing columns only. To remove rows with ``drop``, go through the
+            DataFrame directly instead of calling ``DataFrame.ww.drop``.
```
thehomebrewnerd, line 471: Think it might be worth mentioning that after using the DataFrame method that reinitialization of Woodwork will be required if typing information is needed?

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -1448,3 +1448,76 @@ def test_pop_error(sample_df):
...
+
+def test_accessor_drop(sample_df):
+    xfail_dask_and_koalas(sample_df)
+
+    original_columns = sample_df.columns.copy()
+    schema_df = sample_df.copy()
+    schema_df.ww.init()
+
+    single_input_df = schema_df.ww.drop('is_registered')
+    assert len(single_input_df.ww.columns) == (len(original_columns) - 1)
```
thehomebrewnerd, line 1461: What do you think about this type of assertion instead:
`assert len(single_input_df.ww.columns) == (len(schema_df.columns) - 1)`? 

I think this would also help us confirm that the `schema_df` is not being modified by the drop operation, which we wouldn't identify by comparing against a list that is a copy of the original columns in `sample_df`.

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -1448,3 +1448,76 @@ def test_pop_error(sample_df):
...
+    assert 'is_registered' not in single_input_df.ww.columns
+    assert to_pandas(schema_df).drop('is_registered', axis='columns').equals(to_pandas(single_input_df))
+
+    list_input_df = schema_df.ww.drop(['is_registered'])
+    assert len(list_input_df.ww.columns) == (len(original_columns) - 1)
```
thehomebrewnerd, line 1466: Same as above

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -1448,3 +1448,76 @@ def test_pop_error(sample_df):
...
+    assert 'is_registered' not in list_input_df.ww.columns
+    assert to_pandas(schema_df).drop('is_registered', axis='columns').equals(to_pandas(list_input_df))
+    # should be equal to the single input example above
+    assert single_input_df.ww.schema == list_input_df.ww.schema
+    assert to_pandas(single_input_df).equals(to_pandas(list_input_df))
+
+    multiple_list_df = schema_df.ww.drop(['age', 'full_name', 'is_registered'])
+    assert len(multiple_list_df.ww.columns) == (len(original_columns) - 3)
```
thehomebrewnerd, line 1474: Same as above

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -1448,3 +1448,76 @@ def test_pop_error(sample_df):
...
+    assert to_pandas(schema_df).drop(['is_registered', 'age', 'full_name'], axis='columns').equals(to_pandas(multiple_list_df))
+
+    # Drop the same columns in a different order and confirm resulting DataFrame column order doesn't change
+    different_order_df = schema_df.ww.drop(['is_registered', 'age', 'full_name'])
+    assert different_order_df.ww.schema == multiple_list_df.ww.schema
+    assert to_pandas(multiple_list_df).equals(to_pandas(different_order_df))
+
+
+def test_accessor_drop_rows(sample_df):
+    xfail_dask_and_koalas(sample_df)
```
thehomebrewnerd, line 1488: Remove?

thehomebrewnerd (commented):
Looking pretty good, just a few minor suggestions to consider.

woodwork/tests/accessor/test_table_accessor.py
tamargrey: oh oops--was thinking we might want to test the extra pandas drop parameters, but not necessary for this pr

woodwork/table_accessor.py
tamargrey: Yes, that's a good idea. Changing to `...go through the DataFrame directly instead of calling ``DataFrame.ww.drop`` and then reinitialize Woodwork with DataFrame.ww.init`

woodwork/tests/accessor/test_table_accessor.py
tamargrey: Good point--I think it makes sense to make this change here and the other places below

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -721,7 +721,7 @@ def test_underlying_index(sample_df):
     assert schema_df.index.name is None
     assert type(schema_df.index) == specified_index
 
-    # --> add back when df.ww.drop is implemented
+    # --> add back when we figure out how to handle removing indices
```
thehomebrewnerd, line 724: Is this need covered by an issue already?

thehomebrewnerd (approved):
Looks good - just one question.

woodwork/tests/accessor/test_table_accessor.py
tamargrey: Yeah, that's a fair point. Can probably just delete this commented out code block since it'll get handled by that issue.

PR #646, opened 2021-03-01
Add rename to WoodworkTableAccessor

### LINKED ISSUE #616: Add WoodworkTableAccessor.rename
- Allows users to rename columns in their DataFrame, returning a new DataFrame with the same Woodwork typing info initialized.

tamargrey:
- Adds `Schema.rename` that handles input validation and updating the schema and `WoodworkTableAccessor.rename` that updates the schema and the dataframe
- Closes #616

### DIFF
--- a/woodwork/schema.py
+++ b/woodwork/schema.py
@@ -336,9 +336,6 @@ class Schema(object):
 
         Returns:
             woodwork.Schema: Schema with the specified columns renamed.
-
-        Note:
-            Index and time index columns cannot be renamed.
         """
         if not isinstance(columns, dict):
             raise TypeError('columns must be a dictionary')
@@ -348,8 +345,6 @@ class Schema(object):
                 raise ColumnNotPresentError(f"Column to rename must be present. {old_name} cannot be found.")
             if new_name in self.columns and new_name not in columns.keys():
                 raise ValueError(f"The column {new_name} is already present. Please choose another name to rename {old_name} to or also rename {old_name}.")
-            if old_name == self.index or old_name == self.time_index:
-                raise KeyError(f"Cannot rename index or time index columns such as {old_name}.")
 
         if len(columns) != len(set(columns.values())):
             raise ValueError('New columns names must be unique from one another.')
--- a/woodwork/schema.py
+++ b/woodwork/schema.py
@@ -4,6 +4,7 @@ import copy
 import pandas as pd
 
 import woodwork as ww
+from woodwork.exceptions import ColumnNotPresentError
 from woodwork.schema_column import (
     _add_semantic_tags,
     _get_column_dict,
@@ -344,7 +345,7 @@ class Schema(object):
 
         for old_name, new_name in columns.items():
             if old_name not in self.columns:
-                raise KeyError(f"Column to rename must be present. {old_name} cannot be found.")
+                raise ColumnNotPresentError(f"Column to rename must be present. {old_name} cannot be found.")
             if new_name in self.columns and new_name not in columns.keys():
                 raise ValueError(f"The column {new_name} is already present. Please choose another name to rename {old_name} to or also rename {old_name}.")
             if old_name == self.index or old_name == self.time_index:
--- a/woodwork/schema.py
+++ b/woodwork/schema.py
@@ -327,6 +327,43 @@ class Schema(object):
             _check_time_index(self.columns.keys(), new_time_index, self.logical_types.get(new_time_index))
             self._set_time_index_tags(new_time_index)
 
+    def rename(self, columns):
+        """Renames columns in a Schema
+
+        Args:
+            columns (dict[str -> str]): A dictionary mapping columns whose names
+                we'd like to change to the name to which we'd like to change them.
+
+        Returns:
+            woodwork.Schema: Schema with the specified columns renamed.
+
+        Note:
+            Index and time index columns cannot be renamed.
+        """
+        if not isinstance(columns, dict):
+            raise TypeError('columns must be a dictionary')
+
+        for old_name, new_name in columns.items():
+            if old_name not in self.columns:
+                raise KeyError(f"Column to rename must be present. {old_name} cannot be found.")
+            if new_name in self.columns and new_name not in columns.keys():
+                raise ValueError(f"The column {new_name} is already present. Please choose another name to rename {old_name} to or also rename {old_name}.")
+            if old_name == self.index or old_name == self.time_index:
+                raise KeyError(f"Cannot rename index or time index columns such as {old_name}.")
+
+        if len(columns) != len(set(columns.values())):
+            raise ValueError('New columns names must be unique from one another.')
+
+        new_schema = self._get_subset_schema(list(self.columns.keys()))
+
+        cols_to_update = {}
+        for old_name, new_name in columns.items():
+            col = new_schema.columns.pop(old_name)
+            cols_to_update[new_name] = col
+
+        new_schema.columns.update(cols_to_update)
+        return new_schema
+
     def _set_index_tags(self, index):
         '''
         Updates the semantic tags of the index by removing any standard tags
--- a/woodwork/schema.py
+++ b/woodwork/schema.py
@@ -331,8 +331,7 @@ class Schema(object):
         """Renames columns in a Schema
 
         Args:
-            columns (dict[str -> str]): A dictionary mapping columns whose names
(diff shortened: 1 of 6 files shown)


woodwork/schema.py
```diff
@@ -327,6 +327,43 @@ def set_time_index(self, new_time_index):
...
+
+        Note:
+            Index and time index columns cannot be renamed.
+        """
+        if not isinstance(columns, dict):
+            raise TypeError('columns must be a dictionary')
+
+        for old_name, new_name in columns.items():
+            if old_name not in self.columns:
+                raise KeyError(f"Column to rename must be present. {old_name} cannot be found.")
```
gsheni, line 348: Raise a custom woodwork exceptions for these (which are subclasses properly `class ColumnNotPresent(KeyError)`)?

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -825,15 +825,17 @@ def test_accessor_with_falsy_column_names(falsy_names_df):
...
     # dt[''] = popped_col
     # assert dt[''].name == ''
     # assert dt['']._series.name == ''
 
-    # --> add back in once rename is implemented on the accessor
-    # dt = dt.rename({'': 'col_with_name'})
-    # assert '' not in dt.columns
-    # assert 'col_with_name' in dt.columns
+    renamed_df = schema_df.ww.rename({0: 'col_with_name'})
+    assert 0 not in renamed_df.columns
```
gsheni, line 837: Check that the original schema has not been modified?

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -1509,3 +1511,45 @@ def test_accessor_drop_errors(sample_df):
...
+    assert new_df.ww.metadata == table_metadata
+    assert new_df.ww.columns['id']['description'] == id_description
+
+    old_col = sample_df.ww.columns['age']
+    new_col = new_df.ww.columns['birthday']
+    assert old_col['logical_type'] == new_col['logical_type']
+    assert old_col['semantic_tags'] == new_col['semantic_tags']
+    assert old_col['dtype'] == new_col['dtype']
+
+    new_df = sample_df.ww.rename({'age': 'full_name', 'full_name': 'age'})
```
gsheni, line 1549: good idea to check this!

woodwork/tests/accessor/test_table_accessor.py
gsheni: actually this might not be necessary. We are already checking this below and in the Schema test.

woodwork/schema.py
tamargrey: I wonder if this should subclass LookupError? 

We have a couple of other exceptions that we throw when a column that a user asks for isn't present (rename, drop, pop, reset_semantic_tags, _check_logical_types) and we use LookupError for most of those, but not all of them. Drop uses a ValueError, and rename is using a KeyError. Ideally, I think we'd be able to use the same exception in all of those cases.

gsheni: Looking at the Python [docs](https://docs.python.org/3/library/exceptions.html#LookupError), `LookupError` is actually the base of KeyError. Since KeyError is specifically for a dictionaries, let's use LookupError to specify the column was not found in the DataFrame. 

### exception LookupError
> The base class for the exceptions that are raised when a key or index used on a mapping or sequence is invalid: IndexError, KeyError. This can be raised directly by codecs.lookup().
> 

###  exception KeyError
> Raised when a mapping (dictionary) key is not found in the set of existing keys.

woodwork/schema.py
```diff
@@ -327,6 +327,43 @@ def set_time_index(self, new_time_index):
             _check_time_index(self.columns.keys(), new_time_index, self.logical_types.get(new_time_index))
             self._set_time_index_tags(new_time_index)
 
+    def rename(self, columns):
+        """Renames columns in a Schema
+
+        Args:
+            columns (dict[str -> str]): A dictionary mapping columns whose names
+                we'd like to change to the name to which we'd like to change them.
```
thehomebrewnerd, line 335: Maybe we can simplify this a bit?  Perhaps something like:
```
columns (dict[str -> str]): A dictionary mapping current column names to new column names.
```

woodwork/schema.py
```diff
@@ -327,6 +327,43 @@ def set_time_index(self, new_time_index):
...
+        if not isinstance(columns, dict):
+            raise TypeError('columns must be a dictionary')
+
+        for old_name, new_name in columns.items():
+            if old_name not in self.columns:
+                raise KeyError(f"Column to rename must be present. {old_name} cannot be found.")
+            if new_name in self.columns and new_name not in columns.keys():
+                raise ValueError(f"The column {new_name} is already present. Please choose another name to rename {old_name} to or also rename {old_name}.")
+            if old_name == self.index or old_name == self.time_index:
+                raise KeyError(f"Cannot rename index or time index columns such as {old_name}.")
```
thehomebrewnerd, line 352: I don't recall now, but is there a reason we need to prevent renaming index or time index columns?

woodwork/table_accessor.py
```diff
@@ -484,6 +484,25 @@ def drop(self, columns):
 
         return new_df
 
+    def rename(self, columns):
+        """Renames columns in a DataFrame, maintaining Woodwork typing information.
+
+        Args:
+            columns (dict[str -> str]): A dictionary mapping columns whose names
+                we'd like to change to the name to which we'd like to change them.
```
thehomebrewnerd, line 492: Comment in schema docstring about simplifying also applies here.

woodwork/table_accessor.py
```diff
@@ -484,6 +484,25 @@ def drop(self, columns):
...
+
+        Returns:
+            pd.DataFrame: DataFrame with the specified columns renamed, maintaining Woodwork typing information.
```
thehomebrewnerd, line 495: `pd.DataFrame` -> `DataFrame`

woodwork/schema.py
tamargrey: I'm not sure--I can't think of a good reason, though

thehomebrewnerd: Maybe it had something to do with the complications of how we managed the index and time index before, but this seems like something we could allow now.

tamargrey: I'm fine with allowing renaming index columns in the Accessor if you are!

woodwork/schema.py
tamargrey: This is much cleaner! Changing

woodwork/schema.py
thehomebrewnerd: I haven't been able to think of a reason why we shouldn't allow it. 

@gsheni Do you have any objections to allowing users to rename the index column or time index column?

gsheni: Nope, I don't have any objections to it, and seems like a useful feature to me.

woodwork/schema.py
tamargrey: Added a `ColumnNotPresentError` that's used by `rename`. If we want to use this error in the other methods that are throwing similar errors, then we should probably do that as a part of the improvements epic.

gsheni: Added: https://github.com/alteryx/woodwork/issues/649

tamargrey: Thanks!

woodwork/table_accessor.py
```diff
@@ -722,7 +718,7 @@ def _get_invalid_schema_message(dataframe, schema):
             return f'dtype mismatch for column {name} between DataFrame dtype, '\
                 f'{df_dtype}, and {schema.logical_types[name]} dtype, {schema_dtype}'
     if schema.index is not None:
-        if not all(dataframe.index == dataframe[schema.index]):
+        if not all(dataframe.index == dataframe[schema.index]):  # --> not rel for dask bc no underlying and isunique nto going to happen
```
thehomebrewnerd, line 721: Remove comment?

thehomebrewnerd: Or we can fix when things get merged with the Dask PR...

thehomebrewnerd (approved):
One comment about a comment, but otherwise looks good!

woodwork/table_accessor.py
tamargrey: oh haha I'll remove. Was something I was thinking about for your Dask PR, but I realized I liked the way you handled it better.

PR #648, opened 2021-03-01
Add Dask support to WoodworkTableAccessor

### LINKED ISSUE #610: Add Dask Table Accessor
Allows Dask DataFrames to use the ww namespace for accessors.

thehomebrewnerd:
- Add Dask support to WoodworkTableAccessor
- Closes #610

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -35,6 +35,7 @@ Release Notes
         * Add KoalasColumnAccessor (:pr:`634`)
         * Add ``pop`` to WoodworkTableAccessor (:pr:`636`)
         * Add ``drop`` to WoodworkTableAccessor (:pr:`640`)
+        * Add DaskTableAccessor (:pr:`648`)
     * Fixes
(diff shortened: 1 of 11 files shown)


tamargrey (commented):
Looking good! I think there's a few places in the table accessor documentation where we refer to returning a `pd.Series` (pop) or `pd.DataFrame` (drop), which we should probably be changing to include Dask at this point.

thehomebrewnerd:
> Looking good! I think there's a few places in the table accessor documentation where we refer to returning a `pd.Series` (pop) or `pd.DataFrame` (drop), which we should probably be changing to include Dask at this point.

@tamargrey Updated `pop` and `drop` docstrings. Also reviewed a few others and updated for consistency in the way we are formatting the return values for DataFrame. `describe` and `mutual_information` still list `pd.DataFrame` as the return type, which should be correct since we are bringing data into memory to do those calculations.

tamargrey (approved):
lgtm

PR #651, opened 2021-03-02
Add schema properties to WoodworkTableAccessor

### LINKED ISSUE #602: @property does not pass from Schema to Accessor

tamargrey:
- We weren't getting proper property behavior with attributes that are properties on the Schema but are getting pulled in the WoodworkTableAccessor's getattr, so by adding them as properties on the WoodworkTableAccessor, we get the benefit of not being able to set those properties directly unless we define a setter, which is important for things like `index` and `time_index`
- Closes #602

### DIFF
--- a/docs/source/api_reference.rst
+++ b/docs/source/api_reference.rst
@@ -19,6 +19,7 @@ WoodworkTableAccessor
     WoodworkTableAccessor.mutual_information_dict
     WoodworkTableAccessor.mutual_information
     WoodworkTableAccessor.pop
+    WoodworkTableAccessor.rename
     WoodworkTableAccessor.select
     WoodworkTableAccessor.set_index
     WoodworkTableAccessor.set_types
@@ -54,6 +55,7 @@ Schema
 
     Schema
     Schema.add_semantic_tags
+    Schema.rename
     Schema.remove_semantic_tags
     Schema.reset_semantic_tags
     Schema.set_index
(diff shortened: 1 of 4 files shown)


thehomebrewnerd (approved):
LGTM!

PR #664, opened 2021-03-04
Preserve pandas underlying index when not creating a Woodwork index

### LINKED ISSUE #597: Replicate preserve index behavior implemented on DataTable

tamargrey:
- Only updates underlying pandas index when a Woodwork index is set. Otherwise leaves underlying index alone.
- Closes #597

### DIFF
--- a/woodwork/tests/accessor/test_table_accessor.py
+++ b/woodwork/tests/accessor/test_table_accessor.py
@@ -1465,7 +1465,7 @@ def test_accessor_set_index(sample_df):
     assert sample_df.ww.index is None
     if isinstance(sample_df, pd.DataFrame):
         # underlying index not set for Dask/Koalas
-        # Check that underlying index doesn't get reset when Woodworki index is removed
+        # Check that underlying index doesn't get reset when Woodwork index is removed
         assert (sample_df.index == sample_df['full_name']).all()
 
 
--- a/woodwork/tests/accessor/test_table_accessor.py
+++ b/woodwork/tests/accessor/test_table_accessor.py
@@ -795,7 +795,7 @@ def test_underlying_index_reset(sample_df):
     assert type(sample_df.index) == unspecified_index
 
 
-def test_underlying_index_after_updates(sample_df):
+def test_underlying_index_unchanged_after_updates(sample_df):
     if dd and isinstance(sample_df, dd.DataFrame):
         pytest.xfail('Setting underlying index is not supported with Dask input')
     if ks and isinstance(sample_df, ks.DataFrame):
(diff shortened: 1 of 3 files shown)


woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -705,58 +705,142 @@ def test_make_index(sample_df):
...
+
+    sample_df.ww.set_index('full_name')
+    assert type(sample_df.index) == specified_index
+
+    sample_df.ww.set_index(None)
+    assert type(sample_df.index) == specified_index
+
+    # Use pandas operation to reset index
+    sample_df.ww.reset_index(drop=True, inplace=True)
+    assert type(sample_df.index) == unspecified_index
```
thehomebrewnerd, line 784: Do you think it would be worth confirming the behavior of `df.ww.reset_index` when we have an index set, or when we use `inplace=False`? Not sure it is necessary though...

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -1370,7 +1454,7 @@ def test_accessor_set_index(sample_df):
     assert sample_df.ww.index is None
     if isinstance(sample_df, pd.DataFrame):
         # underlying index not set for Dask/Koalas
-        assert (sample_df.index == range(4)).all()
+        assert (sample_df.index == sample_df['full_name']).all()
```
thehomebrewnerd, line 1457: Maybe add a comment here that we are checking to make sure that the underlying index doesn't change when setting the index to None? Might not be clear why we are testing this a few months in the future.

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -705,58 +705,142 @@ def test_make_index(sample_df):
...
+
+    sample_df.ww.set_index(None)
+    assert type(sample_df.index) == specified_index
+
+    # Use pandas operation to reset index
+    sample_df.ww.reset_index(drop=True, inplace=True)
+    assert type(sample_df.index) == unspecified_index
+
+
+def test_underlying_index_after_updates(sample_df):
```
thehomebrewnerd, line 787: Mabye call this `test_underlying_index_unchanged_after_updates` to make it a bit more clear what we are testing?

thehomebrewnerd (approved):
I think this looks pretty good - made a few suggestions, but I don't think any of them are critical, so up to you if you want to change or not.

woodwork/tests/accessor/test_table_accessor.py
tamargrey: yeah, happy to add both of those checks

woodwork/tests/accessor/test_table_accessor.py
thehomebrewnerd: Added comment has an extra `i` in it. `Woodworki` -> `Woodwork`

thehomebrewnerd (approved):
Looks good aside from the very minor typo in the added comment.

PR #666, opened 2021-03-05
Fix bug in _reset_semantic_tags causing columns to share the same semantic tags set

### LINKED ISSUE #662: Set time index adds tag to multiple columns after reset_semantic_tags call
If you try to set the time index after a call to reset the semantic tags on a dataframe, the `time_index` tag gets added to multiple columns and can return the wrong value for `df.ww.time_index`.

#### Code Sample

```python
>>> import woodwork as ww
>>> df = ww.demo.load_retail(nrows=100, return_dataframe=True)
df.ww.init(name="retail")
>>> df.ww.init(name="retail")
>>> df.ww.reset_semantic_tags()
>>> df.ww.set_time_index('order_date')
>>> df.ww.time_index
'description'
>>> df.ww
                   Physical Type     Logical Type Semantic Tag(s)
Column                                                           
order_product_id           Int64          Integer     ['numeric']
order_id                   Int64          Integer     ['numeric']
product_id              category      Categorical    ['category']
description               string  NaturalLanguage  ['time_index']
quantity                   Int64          Integer     ['numeric']
order_date        datetime64[ns]         Datetime  ['time_index']
unit_price               float64           Double     ['numeric']
customer_name             string  NaturalLanguage  ['time_index']
country                   string  NaturalLanguage  ['time_index']
total                    float64           Double     ['numeric']
cancelled                boolean          Boolean  ['time_index']
```

thehomebrewnerd:
- Fix bug in _reset_semantic_tags causing columns to share the same semantic tags set
- Closes #662 

This PR fixes a bug in `_reset_semantic_tags` that was causing multiple schema columns to share the same semantic tags set after reset. This was being cause by resetting the tags to `LogicalType.standard_tags` without creating a new set, so any columns that shared the same logical type were using the same set after the reset. This fix creates a new set, so that each column now has a unique set, and adds a test to verify this.

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -42,6 +42,7 @@ Release Notes
         * Adds ``__getitem__`` to WoodworkTableAccessor (:pr:`633`)
     * Fixes
         * Create new Schema object when performing pandas operation on Accessors (:pr:`595`)
+        * Fix bug in ``_reset_semantic_tags`` causing columns to share same semantic tags set (:pr:`666`)
     * Changes
         * Move mutual information logic to statistics utils file (:pr:`584`)
         * Bump min Koalas version to 1.4.0 (:pr:`638`)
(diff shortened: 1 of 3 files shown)


tamargrey (approved):
lgtm! glad it was a quick fix!

PR #668, opened 2021-03-05
Remove dtype from Schema

### LINKED ISSUE #653: Remove dtype from column dictionary
When Woodwork is initialized, the Schema object creates data dictionaries for each column and stores these in the Schema object. These column dictionaries include the dtype of the column, which is currently set to the value of the `LogicalType.pandas_dtype` on init. This is problematic for a few reasons:
- We have two sources of information for the dtype of the columns (the series dtype and the dtype stored in the schema)
- For Koalas dataframes, the dtype of the column might be changed to the value of the `LogicalType.backup_dtype` instead of `LogicalType.pandas_dtype` and as a result the value in the column dict could be incorrect
- When we initialize a new column using the column dictionary we need to drop the dtype from the dictionary as dtype is not a valid init parameter

We could drop the dtype from the column dictionary completely and instead pull the dtype from the series directly when it is needed. After init the dtype of the series should be set to the correct value - either `LogicalType.pandas_dtype` or `LogicalType.backup_dtype`. (Note one potential complication here is that the backup dtype for string is `str`. Running `.astype('str') results in an object dtype of `object` for a Koalas Series).

If we drop the dtype from the column dictionary, we would also need to update the Schmea and accessor repr methods appropriately - the schema has no access to the dataframe, so we would either need to display the pandas_dtype or backup_dtype values, or not display the dtype in the Schema repr and only include those in the accessor repr. Also, we would need to move the `physical_types` property from the Schema to the accessor and update the code accordingly.

If we make this change, we should also review the code when initializing a table with an existing schema to make sure the proper validation occurs to prevent initialization with dtypes that are incompatible between the series and the LogicalType, or to perform the proper type conversion during init if differences are detected.

### LINKED ISSUE #656: Repr physical type is not changed when updating the LogicalType of a column with accessor
When using `df.ww.set_types()` to change the logical type of a column, the repr does not properly display the updated dtype of the column. This is likely because the dtype item in the column dictionary is not getting updated, and this is the value that is used in the repr.

#### Code Sample

```python
>>> import pandas as pd
>>> import woodwork as ww
>>> 
>>> df = pd.DataFrame({
...     'integers': [-2, 30, 20],
...     'bools': [True, False, True],
...     'names': ["Jane Doe", "Bill Smith", "John Hancock"]
... })
>>> df
   integers  bools         names
0        -2   True      Jane Doe
1        30  False    Bill Smith
2        20   True  John Hancock
>>> df.ww.init()
>>> df.ww
         Physical Type Logical Type Semantic Tag(s)
Column                                             
integers         Int64      Integer     ['numeric']
bools          boolean      Boolean              []
names         category  Categorical    ['category']
>>> df.dtypes
integers       Int64
bools        boolean
names       category
dtype: object
>>> df.ww.set_types(logical_types={'names': 'FullName'})
>>> df.ww
         Physical Type Logical Type Semantic Tag(s)
Column                                             
integers         Int64      Integer     ['numeric']
bools          boolean      Boolean              []
names         category     FullName              []
>>> df.dtypes
integers      Int64
bools       boolean
names        string
dtype: object

```
After initialization, Woodwork infers the names column to be Categorical and the physical type shown in the repr is `category` as expected and is also the dtype of the column. However, after running `set_types` and changing the LogicalType to `FullName`, the repr still shows the physical type as `category` although the dataframe dtype is `string` which is what we expect to see in both places.

### LINKED ISSUE #658: Schema validation doesn't differentiate between `int64` and `Int64` dtypes
Our definition of what's a valid schema for a DataFrame involves that the dtypes of the DataFrame all match the valid dtype for that column's LogicalType. This means that an `int64` dtype DataFrame series should be invalid for a column with logical type `Integer`.

However, we do not catch that case currently:

 

This is how we attempt to catch that case:
```python
        df_dtype = dataframe[name].dtype
        valid_dtype = _get_valid_dtype(dataframe[name], schema.logical_types[name])

        if df_dtype != valid_dtype:
            ....
```
This is probably related to how pandas differentiates between the nullable and non-nullable dtypes, so we need to make sure we check for valid dtypes (here and across Woodwork) in ways that do differentiate between nullable and non-nullable dtypes.

tamargrey:
- Removes the dtype field from the Schema class so that we're not storing the dtype in multiple ways
    - updates Schema repr to no longer contain physical type since it would need to know the DataFrame type
    - Adds reprs that contain physical type to Table Accessor 
    - Adds `_get_valid_ltype_dtype` helper to get the user-facing dtype (as opposed to the actual dtype of the series)
- Closes #653 , closes #658, closes #656

### DIFF
--- a/woodwork/accessor_utils.py
+++ b/woodwork/accessor_utils.py
@@ -134,4 +134,3 @@ def _get_valid_ltype_dtype(series, logical_type):
     if ks and isinstance(series, ks.Series) and backup_dtype:
         dtype = backup_dtype
     return dtype
-# --> dont pass in whole series just type?? or a needs_backup
--- a/woodwork/accessor_utils.py
+++ b/woodwork/accessor_utils.py
@@ -65,7 +65,7 @@ def _update_column_dtype(series, logical_type):
             series = ks.from_pandas(formatted_series)
         else:
             series = series.apply(_reformat_to_latlong)
-    if logical_type.pandas_dtype != str(series.dtype):  # --> should this be cheking the valid ltype dtype or just pandas?
+    if logical_type.pandas_dtype != str(series.dtype):
         # Update the underlying series
         try:
             if _get_ltype_class(logical_type) == Datetime:
--- a/woodwork/accessor_utils.py
+++ b/woodwork/accessor_utils.py
@@ -80,7 +80,7 @@ def _update_column_dtype(series, logical_type):
                 else:
                     series = pd.to_datetime(series, format=logical_type.datetime_format)
             else:
-                new_dtype = _get_valid_ltype_dtype_str(series, logical_type)
+                new_dtype = _get_valid_ltype_dtype(series, logical_type)
                 series = series.astype(new_dtype)
         except (TypeError, ValueError):
             error_msg = f'Error converting datatype for {series.name} from type {str(series.dtype)} ' \
@@ -110,7 +110,7 @@ def _is_dataframe(data):
     return False
 
 
-def _get_valid_underlying_dtype_str(series, logical_type):
+def _get_valid_underlying_dtype(series, logical_type):
     """Return the dtype that is considered valid for a series
     with the given logical_type"""
     backup_dtype = logical_type.backup_dtype
@@ -126,13 +126,12 @@ def _get_valid_underlying_dtype_str(series, logical_type):
     return str(valid_dtype)
 
 
-# --> maybe pick a better name for these - also the str at the end might be redundant
-def _get_valid_ltype_dtype_str(series, logical_type):
+def _get_valid_ltype_dtype(series, logical_type):
     """Return the dtype that is considered valid for a given logical_type"""
     dtype = logical_type.pandas_dtype
 
     backup_dtype = logical_type.backup_dtype
     if ks and isinstance(series, ks.Series) and backup_dtype:
         dtype = backup_dtype
-    return str(dtype)
+    return dtype
 # --> dont pass in whole series just type?? or a needs_backup
--- a/woodwork/accessor_utils.py
+++ b/woodwork/accessor_utils.py
@@ -65,7 +65,7 @@ def _update_column_dtype(series, logical_type):
             series = ks.from_pandas(formatted_series)
         else:
             series = series.apply(_reformat_to_latlong)
-    if logical_type.pandas_dtype != str(series.dtype):
+    if logical_type.pandas_dtype != str(series.dtype):  # --> should this be cheking the valid ltype dtype or just pandas?
         # Update the underlying series
         try:
             if _get_ltype_class(logical_type) == Datetime:
@@ -80,10 +80,7 @@ def _update_column_dtype(series, logical_type):
                 else:
                     series = pd.to_datetime(series, format=logical_type.datetime_format)
             else:
-                if ks and isinstance(series, ks.Series) and logical_type.backup_dtype:
-                    new_dtype = logical_type.backup_dtype
-                else:
-                    new_dtype = logical_type.pandas_dtype
+                new_dtype = _get_valid_ltype_dtype_str(series, logical_type)
                 series = series.astype(new_dtype)
         except (TypeError, ValueError):
             error_msg = f'Error converting datatype for {series.name} from type {str(series.dtype)} ' \
@@ -113,7 +110,7 @@ def _is_dataframe(data):
     return False
 
 
-def _get_valid_dtype(series, logical_type):
+def _get_valid_underlying_dtype_str(series, logical_type):
     """Return the dtype that is considered valid for a series
     with the given logical_type"""
     backup_dtype = logical_type.backup_dtype
@@ -126,4 +123,16 @@ def _get_valid_dtype(series, logical_type):
     else:
         valid_dtype = logical_type.pandas_dtype
 
-    return valid_dtype
+    return str(valid_dtype)
+
+
+# --> maybe pick a better name for these - also the str at the end might be redundant
+def _get_valid_ltype_dtype_str(series, logical_type):
+    """Return the dtype that is considered valid for a given logical_type"""
+    dtype = logical_type.pandas_dtype
+
+    backup_dtype = logical_type.backup_dtype
+    if ks and isinstance(series, ks.Series) and backup_dtype:
+        dtype = backup_dtype
+    return str(dtype)
+# --> dont pass in whole series just type?? or a needs_backup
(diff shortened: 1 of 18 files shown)


woodwork/accessor_utils.py
```diff
@@ -65,7 +65,7 @@ def _update_column_dtype(series, logical_type):
             series = ks.from_pandas(formatted_series)
         else:
             series = series.apply(_reformat_to_latlong)
-    if logical_type.pandas_dtype != str(series.dtype):
+    if logical_type.pandas_dtype != str(series.dtype):  # --> should this be cheking the valid ltype dtype or just pandas?
```
tamargrey, line 68: I think we might want to switch to checking if `_get_valid_underlying_dtype(..) != str(series.dtype)` because a series that's using a backup dtype will always go into this code block even if it's the correct dtype

woodwork/accessor_utils.py
```diff
@@ -126,4 +123,14 @@ def _get_valid_dtype(series, logical_type):
     else:
         valid_dtype = logical_type.pandas_dtype
 
-    return valid_dtype
+    return str(valid_dtype)
+
+
+def _get_valid_ltype_dtype(series, logical_type):
```
tamargrey, line 129: Added this helper because it was a code pattern I was seeing a fair amount, and I think it helps us differentiate when we need to be dealing with the actual series dtype (which will be `'object'` for certain koalas logical types) and when we should be dealing with the user-facing dtype (which should be `'str'` for those same columns).

@thehomebrewnerd I recognize this kind of puts us in another place of storing dtypes in multiple ways, but the sense I got from the Koalas PRs is that we don't really want users to interface with the `'object'` dtype more than necessary. So any time we are showing something to the user, I've made the change to use the `'str'` dtype--this caused changes in a couple of places (how we store physical type in serialization, the Column Accessor's repr, the physical type we present in descrive).

woodwork/column_accessor.py
```diff
@@ -192,15 +197,14 @@ def wrapper(*args, **kwargs):
 
                 # Try to initialize Woodwork with the existing Schema
                 if _is_series(result):
-                    valid_dtype = _get_valid_dtype(result, self._schema['logical_type'])
-                    if result.dtype == valid_dtype:
+                    valid_dtype = _get_valid_underlying_dtype(result, self._schema['logical_type'])
+                    if str(result.dtype) == valid_dtype:
```
tamargrey, line 201: We need to make sure we only compare dtypes' strings, because if a column's dtype is `int64`, it'll return True if we check for equality with both the strings `'int64'` (expected) and `'Int64'` (which seems less expected to me--haven't been able to figure out if this is a bug)

woodwork/column_accessor.py
```diff
@@ -192,15 +197,14 @@ def wrapper(*args, **kwargs):
...
                         schema = copy.deepcopy(self._schema)
-                        # We don't need to pass dtype from the schema to init
-                        del schema['dtype']
                         result.ww.init(**schema, use_standard_tags=self.use_standard_tags)
                     else:
+                        user_dtype = _get_valid_ltype_dtype(result, self._schema['logical_type'])
```
tamargrey, line 205: felt this case was similar to how we tell users to convert to the `convert_dtype` dtype instead of the `valid_dtype` dtype, but I know it's a slightly different situation here, so happy to switch back.

woodwork/type_sys/utils.py
```diff
@@ -74,7 +74,7 @@ def list_logical_types():
         [{'name': ltype.__name__,
           'type_string': ltype.type_string,
           'description': ltype.__doc__,
-          'physical_type': ltype.pandas_dtype,
+          'physical_type': ltype.pandas_dtype,  # --> should we offer both pd dtype and backup?
```
tamargrey, line 77: Probably wouldn't be for this PR, and definitely low priority, but might be useful to add a 'backup_physical_type' section. (I can't remember if we've had this conversation, so maybe there's a good reason for not adding it in)

woodwork/statistics_utils.py
```diff
@@ -89,13 +90,14 @@ def _get_describe_dict(dataframe, include=None):
...
         mode = _get_mode(series)
+
         # The format of the mode should match its format in the DataFrame
         if ks and isinstance(dataframe, ks.DataFrame) and series.name in latlong_columns:
             mode = list(mode)
 
         values["nan_count"] = series.isna().sum()
         values["mode"] = mode
-        values["physical_type"] = column['dtype']
+        values["physical_type"] = _get_valid_ltype_dtype(dataframe[column_name], logical_type)
```
thehomebrewnerd, line 100: Same as other comments - what should this really be?

thehomebrewnerd (commented):
Leaving a few comments for now. I think we need to decide what we really want to present to the user. Once we have that settled I can do a more complete review.

Having said that, I'm starting to think maybe it is an issue that we use the term `backup_dtype`, when it doesn't really represent the real physical type for Koalas in the `object`/`str` situation.  Maybe `backup_dtype` should actually be the physical dtype (`object`) and when we need to tell the user to convert we use `str`. This would kind of flip our logic in terms of how we have handled this so far.

I'm concerned with having our user-presented dtype being `str`, but if they did `series.dtype` they would see `object` with Koalas. That seems like it could be confusing.

woodwork/accessor_utils.py
```diff
@@ -126,4 +123,14 @@ def _get_valid_dtype(series, logical_type):
     else:
         valid_dtype = logical_type.pandas_dtype
 
-    return valid_dtype
+    return str(valid_dtype)
```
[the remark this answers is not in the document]
tamargrey: yep--meant to remove that!

woodwork/column_accessor.py
```diff
@@ -174,7 +179,7 @@ def __repr__(self):
         if self._schema is None:
             _raise_init_error()
         msg = u"<Series: {} ".format(self._series.name)
-        msg += u"(Physical Type = {}) ".format(self._series.dtype)
+        msg += u"(Physical Type = {}) ".format(_get_valid_ltype_dtype(self._series, self.logical_type))
```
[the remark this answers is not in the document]
tamargrey: By changing the backup_dtype to be `object`, we won't have to make this change, since the series' dtype will be correct.

woodwork/column_accessor.py
tamargrey: Now that I think about it more, I agree that it makes the most sense that we show the underlying dtype here. It might make it confusing if a user wants to do the type conversion themself, but if they're using koalas, they likely are familiar enough with koalas to know that they'll need to convert with `str`

woodwork/statistics_utils.py
tamargrey: The other examples were already using the series dtype, so they shouldn't change. But this one is using the column dtype, so we should switch to `dataframe[column_name].dtype` for consistency.

I'm pretty sure that by switching `str` to `object` for the backup dtype, the series dtype should now always match the Logical type's dtype, but if there was some mismatch, we'd want to show the user what their data actually is, and it's cleaner to go through the series than to go through a helper.

tamargrey:
Closing because enough had changed from when I opened this PR that it made more sense to cherry-pick the commits to a new branch.

PR #669, opened 2021-03-05
Adds __setitem__ to WoodworkTableAccessor

### LINKED ISSUE #647: Add __setitem__ to WoodworkTableAccessor
For feature parity with DataTables, we should be able to use `df.ww['new_col'] = series` to add or update a column.

For exact feature parity with DataTables, it should be required that the series has Woodwork typing information initialized. However, it'd also be possible to pass in a series without Woodwork typing initialized and for Woodwork to perform type inference on the data.

jeff-hernandez:
Closes #647 by adding the method `__setitem__` to the WoodworkTableAccessor

### DIFF
--- a/woodwork/exceptions.py
+++ b/woodwork/exceptions.py
@@ -15,8 +15,8 @@ class StandardTagsRemovalWarning(UserWarning):
 
 class StandardTagsChangedWarning(UserWarning):
     def get_warning_message(self, use_standard_tags, col_name):
-        managed, changed = ('used', 'added to') if use_standard_tags else ('not used', 'removed from')
-        return f'Standard tags are {managed} by the data frame and have been {changed} "{col_name}"'
+        changed = 'added to' if use_standard_tags else 'removed from'
+        return f'Standard tags have been {changed} "{col_name}"'
 
 
 class UpgradeSchemaWarning(UserWarning):
--- a/woodwork/exceptions.py
+++ b/woodwork/exceptions.py
@@ -13,6 +13,12 @@ class StandardTagsRemovalWarning(UserWarning):
         return f"Removing standard semantic tag(s) '{', '.join(standard_tags_to_remove)}' from column '{name}'"
 
 
+class StandardTagsChangedWarning(UserWarning):
+    def get_warning_message(self, use_standard_tags, col_name):
+        managed, changed = ('used', 'added to') if use_standard_tags else ('not used', 'removed from')
+        return f'Standard tags are {managed} by the data frame and have been {changed} "{col_name}"'
+
+
 class UpgradeSchemaWarning(UserWarning):
     def get_warning_message(self, saved_version_str, current_schema_version):
         return ('The schema version of the saved Woodwork table '
--- a/woodwork/table_accessor.py
+++ b/woodwork/table_accessor.py
@@ -13,6 +13,7 @@ from woodwork.exceptions import (
     ColumnNameMismatchWarning,
     ColumnNotPresentError,
     ParametersIgnoredWarning,
+    StandardTagsChangedWarning,
     TypingInfoMismatchWarning
 )
 from woodwork.indexers import _iLocIndexerAccessor, _locIndexerAccessor
@@ -189,10 +190,12 @@ class WoodworkTableAccessor:
             column = init_series(column, use_standard_tags=self.use_standard_tags)
         elif self.use_standard_tags and not column.ww.use_standard_tags:
             column.ww._schema['semantic_tags'] |= column.ww.logical_type.standard_tags
-            warnings.warn(f'Standard tags are used by the data frame and have been added to "{col_name}"')
+            message = StandardTagsChangedWarning().get_warning_message(self.use_standard_tags, col_name)
+            warnings.warn(message, StandardTagsChangedWarning)
         elif not self.use_standard_tags and column.ww.use_standard_tags:
             column.ww._schema['semantic_tags'] -= column.ww.logical_type.standard_tags
-            warnings.warn(f'Standard tags are not used by the data frame and have been removed from "{col_name}"')
+            message = StandardTagsChangedWarning().get_warning_message(self.use_standard_tags, col_name)
+            warnings.warn(message, StandardTagsChangedWarning)
 
         self._dataframe[col_name] = column
         self._schema.columns[col_name] = column.ww._schema
--- a/woodwork/table_accessor.py
+++ b/woodwork/table_accessor.py
@@ -187,11 +187,10 @@ class WoodworkTableAccessor:
                           ColumnNameMismatchWarning)
 
         if column.ww._schema is None:
-            column.ww.init()
+            column.ww.init(use_standard_tags=self.use_standard_tags)
 
         self._dataframe[col_name] = column
         self._schema.columns[col_name] = column.ww._schema
-
 
     def __repr__(self):
         return repr(self._schema)
--- a/woodwork/table_accessor.py
+++ b/woodwork/table_accessor.py
@@ -9,6 +9,7 @@ from woodwork.accessor_utils import (
     _update_column_dtype
 )
 from woodwork.exceptions import (
+    ColumnNameMismatchWarning,
     ColumnNotPresentError,
     ParametersIgnoredWarning,
     TypingInfoMismatchWarning
--- a/woodwork/table_accessor.py
+++ b/woodwork/table_accessor.py
@@ -170,6 +170,28 @@ class WoodworkTableAccessor:
         series.ww.init(**column, use_standard_tags=self.use_standard_tags)
         return series
 
+    def __setitem__(self, col_name, column):
+        series = tuple(pkg.Series for pkg in (pd, dd, ks) if pkg)
+        if not isinstance(column, series):
+            raise ValueError('New column must be of Series type')
+
+        # Don't allow reassigning of index or time index with setitem
+        if self.index == col_name:
+            raise KeyError('Cannot reassign index. Change column name and then use dt.set_index to reassign index.')
+        if self.time_index == col_name:
+            raise KeyError('Cannot reassign time index. Change column name and then use dt.set_time_index to reassign time index.')
+
+        if column.name is not None and column.name != col_name:
+            warnings.warn(ColumnNameMismatchWarning().get_warning_message(column.name, col_name),
+                          ColumnNameMismatchWarning)
+
+        if column.ww._schema is None:
+            column.ww.init()
+
+        self._dataframe[col_name] = column
+        self._schema.columns[col_name] = column.ww._schema
+
+
     def __repr__(self):
         return repr(self._schema)
 
--- a/woodwork/table_accessor.py
+++ b/woodwork/table_accessor.py
@@ -178,9 +178,9 @@ class WoodworkTableAccessor:
 
         # Don't allow reassigning of index or time index with setitem
         if self.index == col_name:
-            raise KeyError('Cannot reassign index. Change column name and then use dt.set_index to reassign index.')
+            raise KeyError('Cannot reassign index. Change column name and then use df.ww.set_index to reassign index.')
         if self.time_index == col_name:
-            raise KeyError('Cannot reassign time index. Change column name and then use dt.set_time_index to reassign time index.')
+            raise KeyError('Cannot reassign time index. Change column name and then use df.ww.set_time_index to reassign time index.')
 
         if column.name is not None and column.name != col_name:
             warnings.warn(ColumnNameMismatchWarning().get_warning_message(column.name, col_name),
--- a/woodwork/table_accessor.py
+++ b/woodwork/table_accessor.py
@@ -187,6 +187,12 @@ class WoodworkTableAccessor:
 
         if column.ww._schema is None:
             column = init_series(column, use_standard_tags=self.use_standard_tags)
+        elif self.use_standard_tags and not column.ww.use_standard_tags:
+            column.ww._schema['semantic_tags'] |= column.ww.logical_type.standard_tags
+            warnings.warn(f'Standard tags are used by the data frame and have been added to "{col_name}"')
+        elif not self.use_standard_tags and column.ww.use_standard_tags:
+            column.ww._schema['semantic_tags'] -= column.ww.logical_type.standard_tags
+            warnings.warn(f'Standard tags are not used by the data frame and have been removed from "{col_name}"')
 
         self._dataframe[col_name] = column
         self._schema.columns[col_name] = column.ww._schema
--- a/woodwork/table_accessor.py
+++ b/woodwork/table_accessor.py
@@ -6,7 +6,8 @@ import woodwork.serialize_accessor as serialize
 from woodwork.accessor_utils import (
     _get_valid_dtype,
     _is_dataframe,
-    _update_column_dtype
+    _update_column_dtype,
+    init_series
 )
 from woodwork.exceptions import (
     ColumnNameMismatchWarning,
@@ -185,7 +186,7 @@ class WoodworkTableAccessor:
                           ColumnNameMismatchWarning)
 
         if column.ww._schema is None:
-            column.ww.init(use_standard_tags=self.use_standard_tags)
+            column = init_series(column, use_standard_tags=self.use_standard_tags)
 
         self._dataframe[col_name] = column
         self._schema.columns[col_name] = column.ww._schema
--- a/woodwork/tests/accessor/test_column_accessor.py
+++ b/woodwork/tests/accessor/test_column_accessor.py
@@ -344,7 +344,7 @@ def test_remove_standard_semantic_tag(sample_series):
     series = convert_series(sample_series, Categorical)
     # Check that warning is raised if use_standard_tags is True - tag should be removed
     series.ww.init(logical_type=Categorical, semantic_tags='tag1', use_standard_tags=True)
-    expected_message = 'Standard tags are used by the data frame and have been added to "sample_series"'
+    expected_message = 'Standard tags have been added to "sample_series"'
     with pytest.warns(UserWarning) as record:
         series.ww.remove_semantic_tags(['tag1', 'category'])
     assert len(record) == 1
--- a/woodwork/tests/accessor/test_column_accessor.py
+++ b/woodwork/tests/accessor/test_column_accessor.py
@@ -345,7 +345,7 @@ def test_remove_standard_semantic_tag(sample_series):
     series = convert_series(sample_series, Categorical)
     # Check that warning is raised if use_standard_tags is True - tag should be removed
     series.ww.init(logical_type=Categorical, semantic_tags='tag1', use_standard_tags=True)
-    expected_message = 'Standard tags have been added to "sample_series"'
+    expected_message = 'Standard tags have been removed from "sample_series"'
     with pytest.warns(StandardTagsChangedWarning) as record:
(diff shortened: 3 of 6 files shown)


woodwork/table_accessor.py
```diff
@@ -170,6 +171,27 @@ def __getitem__(self, key):
...
+    def __setitem__(self, col_name, column):
+        series = tuple(pkg.Series for pkg in (pd, dd, ks) if pkg)
+        if not isinstance(column, series):
+            raise ValueError('New column must be of Series type')
+
+        # Don't allow reassigning of index or time index with setitem
+        if self.index == col_name:
+            raise KeyError('Cannot reassign index. Change column name and then use dt.set_index to reassign index.')
+        if self.time_index == col_name:
+            raise KeyError('Cannot reassign time index. Change column name and then use dt.set_time_index to reassign time index.')
```
jeff-hernandez, line 183: How do we want to reassign the index and time index using the accessor?

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -1631,3 +1632,164 @@ def test_sets_koalas_option_on_init(sample_df_koalas):
...
+    df.ww['age'] = new_series
+
+    assert 'age' in df.columns
+    assert 'age' in df.ww.columns
+    assert df.ww['age'].ww.logical_type == original_col.ww.logical_type
+    assert df.ww['age'].ww.semantic_tags == original_col.ww.semantic_tags
+    assert df.ww['age'].dtype == dtype
+    assert original_col is not df.ww['age']
+
+    # Change dtype, logical types, and tags with conflicting use_standard_tags
```
thehomebrewnerd, line 1774: I guess I'm not clear on what we are testing with the conflicting use_standard_tags. I think we would get the same result here if there was no conflict, since the LogicalType Boolean doesn't have any standard tags.

thehomebrewnerd (commented):
Couple things to consider here. Let me know if you'd like to discuss.

woodwork/table_accessor.py
```diff
@@ -170,6 +171,27 @@ def __getitem__(self, key):
...
+            raise KeyError('Cannot reassign index. Change column name and then use df.ww.set_index to reassign index.')
+        if self.time_index == col_name:
+            raise KeyError('Cannot reassign time index. Change column name and then use df.ww.set_time_index to reassign time index.')
+
+        if column.name is not None and column.name != col_name:
+            warnings.warn(ColumnNameMismatchWarning().get_warning_message(column.name, col_name),
+                          ColumnNameMismatchWarning)
+
+        if column.ww._schema is None:
+            column.ww.init(use_standard_tags=self.use_standard_tags)
```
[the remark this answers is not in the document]
jeff-hernandez: Here, we want to initialize Woodwork for the column if it hasn't been already initialized. If a user passes a column that fails the init call, wouldn't the column-related error be raised from that init call and tell the user how to fix?

I'll switch to `init_series`. Is there a better way to check if a column has been initialized other than `column.ww._schema is None`?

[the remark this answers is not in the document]
thehomebrewnerd: I think the issue is that the error that is raised when a column fails init, is very specific to initialization: https://github.com/alteryx/woodwork/blob/42e57bc5afd971ff47e66f4abbb1065eb5d6c603/woodwork/column_accessor.py#L225

However, in this case the user isn't calling `init` we are doing it for them behind the scenes, so I was thinking it could be confusing for a user to see a message about initializing when they never called init explicitly. You are correct, that the message the would see does contain information on how to fix though.

Unless I'm overlooking something I think you should still be able to use `init_series` inside the current check of `if column.ww._schema is None:`.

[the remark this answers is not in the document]
thehomebrewnerd: Maybe we should consider making the error more generic so it can be used here as well and make sense in this context?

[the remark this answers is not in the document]
thehomebrewnerd: Another option would be to only allow users to assign a column that has already been initialized, which would force the user to handle initialization on their own, either with `df.ww.init` or `init_series`.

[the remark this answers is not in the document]
jeff-hernandez: It'd be worth noting in the docs

Is there another way to check if Woodwork has been initialized in the column? I used `column.ww._schema is None` to check but not sure there's anything more direct (`column.ww.is_initialized`)?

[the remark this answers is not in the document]
thehomebrewnerd: The schema check should be fine for initialization, and I don't think there is anything else to check at this point. If the schema exists Woodwork has been initialized successfully, otherwise the schema will be none, so I think this should be a pretty reliable check.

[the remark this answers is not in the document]
jeff-hernandez: I think the error is helpful in this context. I like the idea of making it more generic.

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -1631,3 +1632,164 @@ def test_sets_koalas_option_on_init(sample_df_koalas):
...
+    error_msg = 'Cannot reassign index. Change column name and then use df.ww.set_index to reassign index.'
+    with pytest.raises(KeyError, match=error_msg):
+        df.ww['id'] = df.id
+
+    error_msg = 'Cannot reassign time index. Change column name and then use df.ww.set_time_index to reassign time index.'
+    with pytest.raises(KeyError, match=error_msg):
+        df.ww['signup_date'] = df.signup_date
+
+
+def test_setitem_different_name(sample_df):
```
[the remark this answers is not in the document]
jeff-hernandez: ✅ 5decf3b

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -1631,3 +1632,164 @@ def test_sets_koalas_option_on_init(sample_df_koalas):
...
+        dtype = 'int64'
+        new_series = pd.Series([1, 2, 3], dtype=dtype)
+        new_series = ks.Series(new_series)
+    else:
+        dtype = 'Int64'
+        new_series = pd.Series([1, 2, 3], dtype=dtype)
+
+    df.ww['test_col2'] = new_series
+    assert 'test_col2' in df.columns
+    assert 'test_col2' in df.ww.columns
```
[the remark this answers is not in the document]
jeff-hernandez: ✅ 7e043ec

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -1631,3 +1632,164 @@ def test_sets_koalas_option_on_init(sample_df_koalas):
...
+    df = sample_df.copy()
+    df.ww.init(
+        index='id',
+        time_index='signup_date',
+        use_standard_tags=True
+    )
+
+    # Change to column no change in types
+    original_col = df['age']
+    original_col.ww.init()
```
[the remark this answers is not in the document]
jeff-hernandez: ✅ 6b6f303

woodwork/tests/accessor/test_table_accessor.py
jeff-hernandez: Is there anything missing in this replicate of the test? How do we want to update?

thehomebrewnerd: I think we should use a different LogicalType here. If we use a Double for the new column, we can do the following:
- Create the dataframe with use_standard_tags = True
- Create the series with use_standard_tags = False and add another custom tag like `test_tag` which is currently being done
- Assign the new column to the dataframe and make sure the semantic tags of the column in the dataframe contain the `numeric` tag as well as the custom tag `test_tag`

tamargrey: I agree that we want the table's `use_standard_tags` to take precedence here, but I think it'd be worth warning users if we've changed a column's tags in order to add it to a Woodwork table.

And the case nate outlined is one where Woodwork adds tags to a column, but we should also test a case where Woodwork removes tags from a column before adding it 

- Create the dataframe with use_standard_tags = False
- Create the series with use_standard_tags = True and add another custom tag like test_tag which is currently being done
- Assign the new column to the dataframe and make sure the semantic tags of the column in the dataframe contains only the custom tag test_tag

tamargrey (commented):
Looking good--excited about getting this in!

Also, there's a commented out section from `test_accessor_with_falsy_column_names` that we can add back in now that we can use setitem

woodwork/table_accessor.py
```diff
@@ -168,6 +169,27 @@ def __getitem__(self, key):
         series.ww.init(**column, use_standard_tags=self.use_standard_tags)
         return series
 
+    def __setitem__(self, col_name, column):
+        series = tuple(pkg.Series for pkg in (pd, dd, ks) if pkg)
+        if not isinstance(column, series):
```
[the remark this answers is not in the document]
thehomebrewnerd: @tamargrey I kinda had this thought at one point, but I don't think we were implementing any special checks in the DataTable code, so debated whether we should do it here or not. Eventually I settled on not doing anything differently for now, but I don't have a strong opinion on that.

Maybe this is something that we should implement in the improvements epic? Or maybe wait until it becomes an issue and then fix? Not sure...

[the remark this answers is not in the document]
tamargrey: I think it'd be fine to wait and see if this pops up as an issue

woodwork/table_accessor.py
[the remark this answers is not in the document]
jeff-hernandez: ✅  8f5317e

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -1727,3 +1728,158 @@ def test_sets_koalas_option_on_init(sample_df_koalas):
         ks.set_option('compute.ops_on_diff_frames', False)
         sample_df_koalas.ww.init()
         assert ks.get_option('compute.ops_on_diff_frames') is True
+
+
+def test_setitem_invalid_input(sample_df):
```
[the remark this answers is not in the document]
tamargrey: Gonna wait on worrying about this, so resolving.

woodwork/table_accessor.py
```diff
@@ -168,6 +170,33 @@ def __getitem__(self, key):
...
+                          ColumnNameMismatchWarning)
+
+        if column.ww._schema is None:
+            column = init_series(column, use_standard_tags=self.use_standard_tags)
+        elif self.use_standard_tags and not column.ww.use_standard_tags:
+            column.ww._schema['semantic_tags'] |= column.ww.logical_type.standard_tags
+            warnings.warn(f'Standard tags are used by the data frame and have been added to "{col_name}"')
+        elif not self.use_standard_tags and column.ww.use_standard_tags:
+            column.ww._schema['semantic_tags'] -= column.ww.logical_type.standard_tags
+            warnings.warn(f'Standard tags are not used by the data frame and have been removed from "{col_name}"')
```
tamargrey, line 195: We have a `StandardTagsRemovalWarning` in exceptions.py that should be useful here. Probably worth adding a `StandardTagsAddedWarning`. 

Though it might be worth standardizing it to a single use `StandardTagsChangedWarning` if we can find a good way to update the warning language to work for both.

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -1727,3 +1728,158 @@ def test_sets_koalas_option_on_init(sample_df_koalas):
...
+    new_series = pd.Series([1, 2, 3, 4], name='wrong', dtype='float')
+    if ks and isinstance(sample_df, ks.DataFrame):
+        new_series = ks.Series(new_series)
+
+    warning = 'Name mismatch between wrong and id. Series name is now id'
+    with pytest.warns(ColumnNameMismatchWarning, match=warning):
+        df.ww['id'] = new_series
+
+    assert df.ww['id'].name == 'id'
+    assert 'id' in df.ww._schema.columns.keys()
```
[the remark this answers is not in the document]
jeff-hernandez: ✅  34c0cb3

woodwork/table_accessor.py
jeff-hernandez: ✅ ca1b061

tamargrey: Can we also remove the usage of `StandardTagsRemovalWarning` in `schema.py` and use the new warning instead?

jeff-hernandez: Hmm, can't seem to find the usage in `schema.py`. Can you attach a link to the lines of code?

tamargrey: I meant `schema_column.py`! sorry!

tamargrey: https://github.com/alteryx/woodwork/pull/497/files#diff-2f8d2164896536382b984429f46712c1754ab4d361f34672ac3aa81443a2f5f6R123

jeff-hernandez: Oh okay, will update

jeff-hernandez: ✅ 60a98ba

woodwork/exceptions.py
```diff
@@ -13,6 +13,12 @@ def get_warning_message(self, standard_tags_to_remove, name):
         return f"Removing standard semantic tag(s) '{', '.join(standard_tags_to_remove)}' from column '{name}'"
 
 
+class StandardTagsChangedWarning(UserWarning):
+    def get_warning_message(self, use_standard_tags, col_name):
+        managed, changed = ('used', 'added to') if use_standard_tags else ('not used', 'removed from')
+        return f'Standard tags are {managed} by the data frame and have been {changed} "{col_name}"'
```
thehomebrewnerd, line 19: `data frame` -> `dataframe`

woodwork/tests/accessor/test_column_accessor.py
```diff
@@ -344,7 +344,7 @@ def test_remove_standard_semantic_tag(sample_series):
     series = convert_series(sample_series, Categorical)
     # Check that warning is raised if use_standard_tags is True - tag should be removed
     series.ww.init(logical_type=Categorical, semantic_tags='tag1', use_standard_tags=True)
-    expected_message = "Removing standard semantic tag(s) 'category' from column 'sample_series'"
+    expected_message = 'Standard tags are used by the data frame and have been added to "sample_series"'
```
thehomebrewnerd, line 347: This is not the right message for this scenario. We are are removing a standard tag (`category`) from the series, and we want to warn the user that this has been done. Also, we are doing this on a Series but the message references a dataframe.

woodwork/exceptions.py
```diff
@@ -13,6 +13,12 @@ def get_warning_message(self, standard_tags_to_remove, name):
         return f"Removing standard semantic tag(s) '{', '.join(standard_tags_to_remove)}' from column '{name}'"
 
 
+class StandardTagsChangedWarning(UserWarning):
```
thehomebrewnerd, line 16: Do we still need the `StandardTagsRemovalWarning` class above this now that we are using this new class or do we need to leave it until we delete DataColumn/DataTable?

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -1745,6 +1746,185 @@ def test_sets_koalas_option_on_init(sample_df_koalas):
...
+        new_series = ks.Series(new_series)
+
+    warning = 'Name mismatch between wrong and id. Series name is now id'
+    with pytest.warns(ColumnNameMismatchWarning, match=warning):
+        df.ww['id'] = new_series
+
+    assert df.ww['id'].name == 'id'
+    assert 'id' in df.columns
+    assert 'id' in df.ww._schema.columns.keys()
+    assert 'wrong' not in df.ww.columns
```
thehomebrewnerd, line 1781: I think I might have been a little confused in my prior comment here.  In any case, let's update these assertions:
`assert 'id' in df.ww.columns` which will check that 'id' is present in the schema dictionary
`asssert 'wrong' not in df.ww.columns` to check that 'wrong' is not present
add `assert 'wrong not in df.columns` just to make sure a column named 'wrong didn't end up in the dataframe somehow.

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -1745,6 +1746,185 @@ def test_sets_koalas_option_on_init(sample_df_koalas):
...
+        new_series2 = ks.Series(new_series2)
+
+    warning = 'Name mismatch between wrong2 and new_col. Series name is now new_col'
+    with pytest.warns(ColumnNameMismatchWarning, match=warning):
+        df.ww['new_col'] = new_series2
+
+    assert df.ww['new_col'].name == 'new_col'
+    assert 'new_col' in df.columns
+    assert 'new_col' in df.ww._schema.columns.keys()
+    assert 'wrong2' not in df.ww.columns
```
thehomebrewnerd, line 1794: Use same pattern as in comment above - update assertion that uses `.keys()` and add a new assertion to check 'wrong2' is not in the dataframe.

thehomebrewnerd (changes requested):
A few more things to clean up, but we are getting closer!

woodwork/tests/accessor/test_table_accessor.py
jeff-hernandez: ✅ 60e44f8

woodwork/tests/accessor/test_column_accessor.py
jeff-hernandez: ✅ 9bce6ab

woodwork/exceptions.py
jeff-hernandez: We do currently need it for DataColumn. I think it'd be better to leave it until it's no longer being used.

woodwork/tests/accessor/test_column_accessor.py
```diff
@@ -344,7 +344,7 @@ def test_remove_standard_semantic_tag(sample_series):
     series = convert_series(sample_series, Categorical)
     # Check that warning is raised if use_standard_tags is True - tag should be removed
     series.ww.init(logical_type=Categorical, semantic_tags='tag1', use_standard_tags=True)
-    expected_message = 'Standard tags are used by the data frame and have been added to "sample_series"'
+    expected_message = 'Standard tags have been added to "sample_series"'
```
thehomebrewnerd, line 347: This warning message still doesn't align with the behavior of this test. In line 349 we are running `remove_semantic_tags` which is removing the standard tag of `category` from the series. Then in line 352 we are checking that all the tags have been removed. This warning message should indicate we are removing the standard tag, not adding it.

woodwork/tests/accessor/test_column_accessor.py
```diff
@@ -344,7 +344,7 @@ def test_remove_standard_semantic_tag(sample_series):
...
     with pytest.warns(UserWarning) as record:
```
thehomebrewnerd, line 348: Also just noticed, I think this should be `StandardTagsChangedWarning` instead of `UserWarning`.

thehomebrewnerd (commented):
Still think we need to do a little clean up on the warning message.

woodwork/tests/accessor/test_column_accessor.py
jeff-hernandez: ✅ 7d50d0c

woodwork/tests/accessor/test_column_accessor.py
jeff-hernandez: ✅ 0d46d13

thehomebrewnerd (approved):
Looks good, but please wait to merge until @tamargrey has had a change to review and approve as well.

tamargrey (approved):
lgtm!

PR #672, opened 2021-03-05
Change backup dtype from str to object for relevant Logical Types

### LINKED ISSUE #670: Update backup_dtype to match actual series dtype
In Woodwork we have implemented the concept of a `backup_dtype` on a LogicalType, primarily for Koalas which does not support all of dtypes that pandas supports. However, when defining backup_dtypes for certain logical types, we have not used the actual dtype, but rather the dtype string that the user would have to call to convert the series.

For example the Categorical logical type has `pandas_dtype = 'category'` and `backup_dtype = 'str'`. With Koalas, if you run `series.astype('str')` you get a series with a dtype of `object`.  If you run `series.astype('object')` an error is raised. Since we have been using the backup_dtype value in Woodwork for performing type conversions we defined the `backup_dtype` to be the value that was valid for passing to `astype`. This has created unintended consequences that impact what a user sees in Error messages and `repr` output as well as what gets checked by Woodwork when determining if a schema is valid for a given dataframe.

For consistency and clarity, we should define the `backup_dtype` to be the actual dtype of the series. For the Categorical logical type this would mean `backup_dtype = 'object'`. In places where we need to use `str` either in calls to `astype` or in our error messages, we could use a helper function to return this value based on the series library type and the logical type.

We should also consider renaming `pandas_dtype` to something more generic such as `primary_dtype`. Since we also currently use the `pandas_dtype` when working with Dask, the naming is mismatched and could be confusing.

This issue is related to both #653 and #656.

tamargrey:
- Changes the backup dtype for certain LogicalTypes to be `object` so that a column's physical type will match its series' dtype.
- Updates `_get_valid_dtype` to take in a boolean `needs_backup` to make it clear that the function never looks at a series' dtype; it's determining a valid dtype for a Logical Type. 
- Closes #670 

With this change, we should be able to avoid having to use `backup_dtype` or `pandas_dtype` directly. Either we can use the helpers in `accessor_utils` or we can get the dtype directly from the series.

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -48,6 +48,7 @@ Release Notes
         * Bump min Koalas version to 1.4.0 (:pr:`638`)
         * Preserve pandas underlying index when not creating a Woodwork index (:pr:`664`)
         * Restrict Koalas version to ``<1.7.0`` due to breaking changes (:pr:`674`)
+        * Stop using ``str`` for backup dtype and use ``object`` instead (:pr:`672`)
     * Documentation Changes
         * Update docstrings and API Reference page (:pr:`660`)
     * Testing Changes
--- a/woodwork/accessor_utils.py
+++ b/woodwork/accessor_utils.py
@@ -80,10 +80,7 @@ def _update_column_dtype(series, logical_type):
                 else:
                     series = pd.to_datetime(series, format=logical_type.datetime_format)
             else:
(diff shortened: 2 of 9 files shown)


tamargrey:
Closing because of significant changes from Koalas 1.7.0 that mean we're going to handle all the dtype cleanup in #681

PR #673, opened 2021-03-08
Automated Latest Dependency Updates

machineFL:
This is an auto-generated PR with dependency updates. Please do not delete the dep-update branch because it's needed by the auto-dependency bot.


thehomebrewnerd:
Because of breaking changes with Koalas 1.7.0 we have restricted the version to `<1.7.0` in #674 , so this PR is no longer needed

PR #674, opened 2021-03-08
Limit koalas to <1.7.0

thehomebrewnerd:
- Limit koalas to <1.7.0

Restrict Koalas version to `<1.7.0` as a result of breaking changes with version `1.7.0`.


gsheni (approved):
lgtm

PR #677, opened 2021-03-08
Make sure column order is maintained when necessary on Woodwork Tables

### LINKED ISSUE #503: Look into maintaining column order on Schema
Previously, we'd made sure to only access the list of columns from the DataFrame when grabbing the set of columns in a DataTable, ensuring that the order of columns on the DataFrame was maintained when presenting typing information (in the repr, for example) or when creating a new DataTable from an existing one (_new_dt_from_cols).

The new Schema api doesn't store the DataFrame, so when presenting the repr it may be possible that the column order will not match that of the DataFrame. 

We should look into cases when the columns may become out of order on the Schema and determine if this is a possible issue with methods like `rename` or `pop` along with the repr.

tamargrey:
- Confirms that various Woodwork operations don't change the order of the DataFrame's columns or the types DataFrame that used as the repr
- Closes #503


woodwork/table_accessor.py
```diff
@@ -228,13 +228,17 @@ def loc(self):
...
     @property
     def types(self):
         """DataFrame containing the physical dtypes, logical types and semantic
         tags for the Schema."""
-        return self._schema.types
+        repr_df = self._schema._get_typing_info()
+        # Maintain the same column order used in the DataFrame
+        repr_df = repr_df.loc[list(self._dataframe.columns), :]
+
+        return repr_df
```
tamargrey, line 241: When we remove `dtype` from the Schema dict in #668 , this'll change again (we'll probably want to maintain collumn order on `WoodworkTableAccessor._get_typing_info`), but this works for the meantime to make sure that the Woodwork table repr follows the same column order as the DataFrame even if changes are made

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -1727,3 +1727,56 @@ def test_sets_koalas_option_on_init(sample_df_koalas):
         ks.set_option('compute.ops_on_diff_frames', False)
         sample_df_koalas.ww.init()
         assert ks.get_option('compute.ops_on_diff_frames') is True
+
+
+def test_maintain_column_order_on_type_changes(sample_df):
```
thehomebrewnerd, line 1732: Do we expect that any of these operations would change the order of either the schema columns or the dataframe columns? Doesn't seem like it to me, so I think this test would pass even without the changes to the accessor code. Having said that, I'm fine with leaving this as an additional validation check to make sure we don't break it in the future.

thehomebrewnerd (approved):
This looks pretty good to me. I think a couple of these tests would pass without any code changes, but it's good to have them for extra validation in the future.

One other thought I had - would it be worthwhile for us to have a test where we directly reorder the schema columns by manipulating the dictionary directly? We are doing it indirectly through various operations, but we could build a new schema columns dict with a new order and reassign it (I think at least) and then verify that the ordering still follows the dataframe.

thehomebrewnerd (approved):
LGTM!

PR #682, opened 2021-03-09
Cleanup dtype usage

### LINKED ISSUE #681: Cleanup dtype usage in Accessor
In order to remove the `dtype` field from the Schema dictionary (#653), we need to make sure we have consistent dtype usage across woodwork. There are several changes that should be made :
- replace `pandas_dtype` with `primary_dtype` since it also gets used for Dask
- Confirm we compare strings when looking at dtype equality (at Schema validation and at in _make_dataframe_call)
  - add tests that catch the bugs that happen when strings aren't being compared (see #668)
- pass `type(series)` to `_get_valid_dtype` so that it's clear we never look at the data when determine a logical types valid dtype
- make sure we never do the logic to get the backup dtype directly and either call `_get_valid_dtype` or look at the series' dtype

tamargrey:
- cleans up how we use dtypes across Woodwork
- Closes #681

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -50,6 +50,8 @@ Release Notes
         * Bump min Koalas version to 1.4.0 (:pr:`638`)
         * Preserve pandas underlying index when not creating a Woodwork index (:pr:`664`)
         * Restrict Koalas version to ``<1.7.0`` due to breaking changes (:pr:`674`)
+        * Replace ``pandas_dtype`` with ``primary_dtype`` on Logical Types (:pr:`682`)
+        * Clean up dtype usage across Woodwork (:pr:`682`)
     * Documentation Changes
         * Update docstrings and API Reference page (:pr:`660`)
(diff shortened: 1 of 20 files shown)


docs/source/release_notes.rst
```diff
@@ -50,6 +50,8 @@ Release Notes
         * Bump min Koalas version to 1.4.0 (:pr:`638`)
         * Preserve pandas underlying index when not creating a Woodwork index (:pr:`664`)
         * Restrict Koalas version to ``<1.7.0`` due to breaking changes (:pr:`674`)
+        * Replace ``pandas_dtype`` with ``primary_dtype`` on Logical Types (:pr:`682`)
+        * Clean up dtype usage across Woodwork (:pr:`682`)
```
thehomebrewnerd, line 54: I think we can use a single entry here (probably the second one).

thehomebrewnerd (approved):
Looks good - just one really small comment about the release notes.

PR #683, opened 2021-03-09
Improve error when calling accessor methods or properties before init

### LINKED ISSUE #667: Improve error when calling accessor methods
The new WoodworkTableAccessor methods that were added as part of #660 as well as the `select` method do not check that Woodwork is initialized before continuing. This result in an error message that is unhelpful if called on an uninitialized dataframe: `AttributeError: 'NoneType' object has no attribute 'reset_semantic_tags'`

These methods should be updated:
- select
- set_time_index
- add_semantic_tags
- remove_semantic_tags
- reset_semantic_tags

The following check should be added at the start of the methods:
```python
if self._schema is None:
    _raise_init_error()
```

thehomebrewnerd:
- Improve error when calling accessor methods or properties before init
- Closes #667 

Adds a check to numerous methods as well as corresponding tests to make sure that a helpful error is raised if a user attempts to access a Woodwork property or method before first initializing Woodwork.

### DIFF
--- a/docs/source/api_reference.rst
+++ b/docs/source/api_reference.rst
@@ -72,7 +72,6 @@ Schema
     Schema.add_semantic_tags
     Schema.index
     Schema.logical_types
-    Schema.physical_types
     Schema.rename
     Schema.remove_semantic_tags
     Schema.reset_semantic_tags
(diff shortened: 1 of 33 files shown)


thehomebrewnerd:
> We should probably also have an error raised at `WoodworkTableAccessor.__getitem__` as well as the setters for metadata and description on the Column Accessor since they both raise `TypeError: 'NoneType' object does not support item assignment` if the schema isn't initialized.
> 
> Plus I think the type setting methods on the column accessor is having a similar behavior where it gets passed to the getattr

@tamargrey Made some updates that I think address all these things, including automating the column test like was done for the table.

tamargrey (approved):
lgtm!

PR #685, opened 2021-03-09
Remove dtype from Schema dictionary

### LINKED ISSUE #653: Remove dtype from column dictionary

tamargrey:
- Removes the `dtype` field from the Schema columns dictionary, relying on Series dtype when checking the dtype of a column.
- Adds a separate repr for Schema and Table Accessor since only the Table Accessor has access to physical types
- Closes #653

### DIFF
--- a/woodwork/tests/accessor/test_table_accessor.py
+++ b/woodwork/tests/accessor/test_table_accessor.py
@@ -112,6 +112,21 @@ def test_accessor_schema_property(sample_df):
     assert sample_df.ww._schema == sample_df.ww.schema
 
 
+def test_accessor_physical_types_property(sample_df):
+    sample_df.ww.init()
+
+    assert isinstance(sample_df.ww.physical_types, dict)
+    assert set(sample_df.ww.physical_types.keys()) == set(sample_df.columns)
+    for k, v in sample_df.ww.physical_types.items():
+        logical_type = sample_df.ww.columns[k]['logical_type']
+        if ks and isinstance(sample_df, ks.DataFrame):
+            assert v == logical_type.backup_dtype
+        else:
+            assert v == logical_type.primary_dtype
+
+
+# --> add test for tyypes property
+
 def test_accessor_separation_of_params(sample_df):
     # mix up order of acccessor and schema params
     schema_df = sample_df.copy()
@@ -514,7 +529,6 @@ def test_sets_category_dtype_on_init():
             df = pd.DataFrame(series)
             df.ww.init(logical_types=ltypes)
             assert df.ww.columns[column_name]['logical_type'] == logical_type
-            assert df.ww.columns[column_name]['dtype'] == logical_type.primary_dtype
             assert df[column_name].dtype == logical_type.primary_dtype
 
 
@@ -526,10 +540,8 @@ def test_sets_object_dtype_on_init(latlong_df):
         df = latlong_df.loc[:, [column_name]]
         df.ww.init(logical_types=ltypes)
         assert df.ww.columns[column_name]['logical_type'] == LatLong
-        assert df.ww.columns[column_name]['dtype'] == LatLong.primary_dtype
         assert df[column_name].dtype == LatLong.primary_dtype
-        df_pandas = to_pandas(df[column_name])
-        expected_val = (3, 4)
+
         if ks and isinstance(latlong_df, ks.DataFrame):
             expected_val = [3, 4]
         assert df_pandas.iloc[-1] == expected_val
@@ -562,7 +574,6 @@ def test_sets_string_dtype_on_init():
             df = pd.DataFrame(series)
             df.ww.init(logical_types=ltypes)
             assert df.ww.columns[column_name]['logical_type'] == logical_type
-            assert df.ww.columns[column_name]['dtype'] == logical_type.primary_dtype
             assert df[column_name].dtype == logical_type.primary_dtype
 
 
@@ -584,7 +595,6 @@ def test_sets_boolean_dtype_on_init():
         df = pd.DataFrame(series)
         df.ww.init(logical_types=ltypes)
         assert df.ww.columns[column_name]['logical_type'] == logical_type
-        assert df.ww.columns[column_name]['dtype'] == logical_type.primary_dtype
         assert df[column_name].dtype == logical_type.primary_dtype
 
 
@@ -607,7 +617,6 @@ def test_sets_int64_dtype_on_init():
             df = pd.DataFrame(series)
             df.ww.init(logical_types=ltypes)
             assert df.ww.columns[column_name]['logical_type'] == logical_type
-            assert df.ww.columns[column_name]['dtype'] == logical_type.primary_dtype
             assert df[column_name].dtype == logical_type.primary_dtype
 
 
@@ -628,7 +637,6 @@ def test_sets_float64_dtype_on_init():
         df = pd.DataFrame(series)
         df.ww.init(logical_types=ltypes)
         assert df.ww.columns[column_name]['logical_type'] == logical_type
-        assert df.ww.columns[column_name]['dtype'] == logical_type.primary_dtype
         assert df[column_name].dtype == logical_type.primary_dtype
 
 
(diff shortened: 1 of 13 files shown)


woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -112,6 +112,19 @@ def test_accessor_schema_property(sample_df):
...
+    sample_df.ww.init(logical_types={'age': 'Categorical'})
+
+    assert isinstance(sample_df.ww.physical_types, dict)
+    assert set(sample_df.ww.physical_types.keys()) == set(sample_df.columns)
+    for k, v in sample_df.ww.physical_types.items():
+        logical_type = sample_df.ww.columns[k]['logical_type']
+        if ks and isinstance(sample_df, ks.DataFrame) and logical_type.backup_dtype is not None:
+            assert v == logical_type.backup_dtype
+        else:
+            assert v == logical_type.primary_dtype
```
thehomebrewnerd, line 125: Thoughts on using the `_get_valid_dtype` function here?

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -271,10 +284,6 @@ def test_accessor_equality_with_schema(sample_df, sample_column_names, sample_in
     # eq not implemented on Accessor class, so Schema's eq is called
     assert schema_df.ww.__eq__(comparison_schema)
```
thehomebrewnerd, line 285: Maybe something we should think about for the future, but is this adequate now given the changes we have made? If two accessors have different settings for `use_standard_tags` are they equal? If two have the same logical types and semantic tags but different data, are they equal?

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -651,13 +654,8 @@ def test_sets_datetime64_dtype_on_init():
         df = pd.DataFrame(series)
         df.ww.init(logical_types=ltypes)
         assert df.ww.columns[column_name]['logical_type'] == logical_type
-        assert df.ww.columns[column_name]['dtype'] == logical_type.primary_dtype
         assert df[column_name].dtype == logical_type.primary_dtype
 
-
-def test_invalid_dtype_casting():
-    column_name = 'test_series'
```
thehomebrewnerd, line 659: Is there a reason you removed this and combined with the test above?

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -1146,6 +1137,25 @@ def test_dataframe_methods_on_accessor_other_returns(sample_df):
...
+    if isinstance(sample_df, pd.DataFrame):
+        pytest.skip("No need to test converting pandas DataFrame to pandas")
+
+    sample_df.ww.init(name='woodwork', index='id')
+
+    if dd and isinstance(sample_df, dd.DataFrame):
+        pd_df = sample_df.ww.compute()
+        assert isinstance(pd_df, pd.DataFrame)
+        assert pd_df.ww.index == 'id'
+        assert pd_df.ww.name == 'woodwork'
```
thehomebrewnerd, line 1150: These assertions are the same as those below, so these can be deleted and the assertions below can be moved outside the if/else block.

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -1146,6 +1137,25 @@ def test_dataframe_methods_on_accessor_other_returns(sample_df):
         pd.testing.assert_index_equal(schema_df.ww.keys(), schema_df.keys())
 
 
+def test_dataframe_methods_on_accessor_to_pandas(sample_df):
```
thehomebrewnerd, line 1140: Nice test to add - should we also confirm that the dtypes for all the columns match the expected `primary_dtype` value after converting to pandas?

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -1814,3 +1820,80 @@ def test_maintain_column_order_disordered_schema(sample_df):
...
+
+
+def test_accessor_types(sample_df):
+    sample_df.ww.init()
+
+    returned_types = sample_df.ww.types
+    assert isinstance(returned_types, pd.DataFrame)
+    assert 'Physical Type' in returned_types.columns
+    assert 'Logical Type' in returned_types.columns
+    assert 'Semantic Tag(s)' in returned_types.columns
```
thehomebrewnerd, line 1832: If we want to make sure the column order is consistent, maybe we should check that the list of columns matches what we expect, instead of just checking that certain names are present.

thehomebrewnerd (commented):
Looking pretty good overall, just a few suggestions to review/consider.

woodwork/tests/accessor/test_table_accessor.py
tamargrey: definitely an accident--thanks for catching!

woodwork/tests/accessor/test_table_accessor.py
tamargrey: kind of stuck between wanting to clean this up and also wanting to be a bit more explicit here incase we ever change `_get_valid_dtype` or something to do with dtypes.

woodwork/tests/accessor/test_table_accessor.py
tamargrey: good call--moved

woodwork/tests/accessor/test_table_accessor.py
thehomebrewnerd: I don't see this as a big deal either way - we can leave it as it is for now.

woodwork/tests/accessor/test_table_accessor.py
tamargrey: Good question. I'm thinking that you're right and we should be checking, at the very least, the underlying data as well (probably only for pandas dataframes, though)

For the `use_standard_tags` question, I think two dataframes with opposite `use_standard_tags` values would be equal if no column's logical type had any standard tags.

I'll make an issue, and given that the current equality behavior in the Table Accessor is worse than that in DataTables, we should add it to the initial Schema/Accessor epic

tamargrey: Created #686

woodwork/tests/accessor/test_table_accessor.py
thehomebrewnerd: Ah, interesting. Makes sense.

If this becomes an issue in the future, I think we could probably write a custom `to_pandas` method on the KoalasTableAccesor class to handle this, making sure the dtypes are converted properly.

tamargrey: going to leave the test as is for now without the Categorical column for simplicity

thehomebrewnerd (approved):
LGTM!

PR #692, opened 2021-03-11
Add include_index parameter to DataTable mutual info

### LINKED ISSUE #671: Add avoid_index parameter to mutual info to allow for more control
- We should add the following parameters to the mutual info function and associated defaults.  
    -  `avoid_index=False`
- This way the end-user can have more control of the logic for calculating MI  
- This is for the current DataTable implementation

tamargrey:
- Allows users to include the index column in mutual info calculations, where the default is to not include the index column
- Closes #671

### DIFF
--- a/woodwork/datatable.py
+++ b/woodwork/datatable.py
@@ -1013,10 +1013,10 @@ class DataTable(object):
             nrows (int): The number of rows to sample for when determining mutual info.
                 If specified, samples the desired number of rows from the data.
                 Defaults to using all rows.
-            include_index (bool): If True, the column specified as the index will be included
-                in mutual information calculations. If False, the index column will not have
-                mutual information calculated for it. Defaults to False.
-
+            include_index (bool): If True, the column specified as the index will be
+                included as long as its LogicalType is valid for mutual information calculations.
+                If False, the index column will not have mutual information calculated for it.
+                Defaults to False.
         Returns:
             list(dict): A list containing dictionaries that have keys `column_1`,
             `column_2`, and `mutual_info` that is sorted in decending order by mutual info.
@@ -1081,9 +1081,10 @@ class DataTable(object):
             nrows (int): The number of rows to sample for when determining mutual info.
                 If specified, samples the desired number of rows from the data.
                 Defaults to using all rows.
-            include_index (bool): If True, the column specified as the index will be included
-                in mutual information calculations. If False, the index column will not have
-                mutual information calculated for it. Defaults to False.
+            include_index (bool): If True, the column specified as the index will be
+                included as long as its LogicalType is valid for mutual information calculations.
+                If False, the index column will not have mutual information calculated for it.
+                Defaults to False.
 
         Returns:
             pd.DataFrame: A Dataframe containing mutual information with columns `column_1`,
--- a/woodwork/datatable.py
+++ b/woodwork/datatable.py
@@ -1025,13 +1025,10 @@ class DataTable(object):
         """
         valid_types = get_valid_mi_types()
 
-        valid_columns = [col.name for col in self.columns.values() if (
-            col.name != self.index and _get_ltype_class(col.logical_type) in valid_types)]
+        valid_columns = [col.name for col in self.columns.values() if _get_ltype_class(col.logical_type) in valid_types]
 
-        if (include_index and
-            self.index is not None and
-                _get_ltype_class(self.columns[self.index].logical_type) in valid_types):
-            valid_columns.append(self.index)
+        if not include_index and self.index is not None:
+            valid_columns.remove(self.index)
 
         data = self._dataframe.loc[:, valid_columns]
         if dd and isinstance(data, dd.DataFrame):
--- a/woodwork/datatable.py
+++ b/woodwork/datatable.py
@@ -1000,7 +1000,7 @@ class DataTable(object):
             data[col_name] = new_col.cat.codes
         return data
 
-    def mutual_information_dict(self, num_bins=10, nrows=None):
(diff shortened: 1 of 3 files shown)


woodwork/datatable.py
```diff
@@ -1000,7 +1000,7 @@ def _make_categorical_for_mutual_info(self, data, num_bins):
             data[col_name] = new_col.cat.codes
         return data
 
-    def mutual_information_dict(self, num_bins=10, nrows=None):
+    def mutual_information_dict(self, num_bins=10, nrows=None, include_index=False):
```
tamargrey, line 1003: Assumed we wanted to default to not including the index, but happy to change

woodwork/datatable.py
```diff
@@ -1021,9 +1024,15 @@ def mutual_information_dict(self, num_bins=10, nrows=None):
...
         """
         valid_types = get_valid_mi_types()
+
         valid_columns = [col.name for col in self.columns.values() if (
             col.name != self.index and _get_ltype_class(col.logical_type) in valid_types)]
 
+        if (include_index and
+            self.index is not None and
+                _get_ltype_class(self.columns[self.index].logical_type) in valid_types):
+            valid_columns.append(self.index)
```
tamargrey, line 1034: Another assumption I made: The "index" in question was the DataTable/Woodwork index. 

But now I'm wondering if it might have been in reference to the pandas underlying index. With Woodwork's assumption that if an index is set, it'll match the underlying pandas index, we could change this so that we always add the pandas underlying index as a column for MI and, if Woodwork defined an index, it'd match that. Otherwise, it'd just provide the Pandas index (which could be of note to the user or a RangeIndex if it's not been defined)

woodwork/datatable.py
```diff
@@ -1013,6 +1013,9 @@ def mutual_information_dict(self, num_bins=10, nrows=None):
             nrows (int): The number of rows to sample for when determining mutual info.
                 If specified, samples the desired number of rows from the data.
                 Defaults to using all rows.
+            include_index (bool): If True, the column specified as the index will be included
+                in mutual information calculations. If False, the index column will not have
+                mutual information calculated for it. Defaults to False.
```
thehomebrewnerd, line 1018: Probably would be good to clarify here and in the other docstring that the index column will only be included if it is of a valid type for mutual information, since there are still circumstances where using `include_index=True` could result in the index not being present in the mutual information output.

woodwork/datatable.py
thehomebrewnerd: Would be good to clarify before we close, but I was thinking index referred to Woodwork index, not pandas index...

woodwork/datatable.py
thehomebrewnerd: I'll defer to @gsheni on this point.

thehomebrewnerd (commented):
Looks good overall. A little clarification in the docstrings might be useful.

woodwork/datatable.py
tamargrey: Good point. How about 

> If True, the column specified as the index will be included
                in mutual information calculations as long as its Logical Type supports mutual information.
                If False, the index column will not have mutual information calculated for it.
                Defaults to False.

thehomebrewnerd: I think I'd flip the wording just a bit - it's more about mutual information supporting the LogicalType than the LogicalType supporting mutual information. So maybe something like this:
```
If True, the column specified as the index will be included as long as its LogicalType
is valid for mutual information calculations.
If False, the index column will not have mutual information calculated for it.
Defaults to False.
```

woodwork/datatable.py
gsheni: We should default to False for now.

woodwork/datatable.py
```diff
@@ -1013,17 +1013,26 @@ def mutual_information_dict(self, num_bins=10, nrows=None):
...
         Returns:
             list(dict): A list containing dictionaries that have keys `column_1`,
             `column_2`, and `mutual_info` that is sorted in decending order by mutual info.
             Mutual information values are between 0 (no mutual information) and 1
             (perfect dependency).
         """
         valid_types = get_valid_mi_types()
+
         valid_columns = [col.name for col in self.columns.values() if (
             col.name != self.index and _get_ltype_class(col.logical_type) in valid_types)]
```
gsheni, line 1029: Should we remove the check here? `col.name != self.index`

tamargrey: If we make sure the index gets included here, then we'll need to remove it later if `include_index` is False or if its logical type is not valid for mutual info, so I guess it'd just be the reverse of the way it currently is. I'm fine with either one

woodwork/datatable.py
tamargrey: Created a separate issue #697 for using the underlying index if we decide to add this at a later date, but wanted to keep this PR simple, as this wasn't the intended behavior ("index" was, indeed, the Woodwork index)

woodwork/datatable.py
gsheni: I think if we do 1 less check to `_get_ltype_class` and 1 less check to see if its in `valid_types` its better. 
```python
        valid_types = get_valid_mi_types()

        valid_columns = [col.name for col in self.columns.values() if _get_ltype_class(col.logical_type) in valid_types)]

        if not include_index and self.index is not None:
            valid_columns.remove(self.index)
```

tamargrey: That makes sense to me. Made this change (using `remove` for dropping column names from a list, though, bc pop is for indices)

gsheni (approved):
lgtm

PR #700, opened 2021-03-12
Include DataFrame in accessor equality check

### LINKED ISSUE #686: Improve Table Accessor __eq__ check
Currently, WoodworkTableAccessor checks for equality by grabbing the Schema's `__eq__` method, but this means a few things:
- Won't check DataFrame equality - underlying data could be different but the Accessor's would still come back as equal - we should only check pandas DataFrames though
- Will say two DataFrames are equal even if their `use_standard_tags` values are different as long as the semantic tags values are the same for all columns - not sure if this is a desired behavior or not

We should add an `__eq__` method to `WoodworkTableAccessor` as well as relevant tests.

Note - DataTables does check for equality in pandas DataFrames, so this needs to happen before we remove the DataTables code

tamargrey:
- Adds an `__eq__` method to WoodworkTableAccessor that checks for equality in the underlying data as well as the Schema
- Closes #686

### DIFF
--- a/woodwork/tests/accessor/test_table_accessor.py
+++ b/woodwork/tests/accessor/test_table_accessor.py
@@ -338,34 +338,24 @@ def test_getitem_invalid_input(sample_df):
         df.ww['invalid_column']
 
 
-def test_accessor_equality_with_schema(sample_df, sample_column_names, sample_inferred_logical_types):
-    comparison_schema = Schema(sample_column_names, sample_inferred_logical_types)
-
+def test_accessor_equality(sample_df):
+    # Confirm equality with same schema and same data
     schema_df = sample_df.copy()
     schema_df.ww.init()
 
-    # eq not implemented on Accessor class, so Schema's eq is called
-    assert schema_df.ww.__eq__(comparison_schema)
+    copy_df = schema_df.ww.copy()
+    assert schema_df.ww == copy_df.ww
 
-    logical_types = {
-        'id': Double,
-        'full_name': FullName
-    }
-    semantic_tags = {
-        'email': 'test_tag',
-    }
-    comparison_schema = Schema(sample_column_names,
-                               logical_types={**sample_inferred_logical_types, **logical_types},
-                               index='id',
-                               time_index='signup_date',
-                               semantic_tags=semantic_tags)
-    schema_df = sample_df.copy()
-    schema_df.ww.init(logical_types=logical_types,
-                      index='id',
-                      time_index='signup_date',
-                      semantic_tags=semantic_tags,
-                      already_sorted=True)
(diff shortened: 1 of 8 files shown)


woodwork/tests/schema/test_schema.py
```diff
@@ -144,6 +144,24 @@ def test_schema_equality(sample_column_names, sample_inferred_logical_types):
...
+    assert schema.use_standard_tags != no_standard_tags_schema.use_standard_tags
+    assert schema != no_standard_tags_schema
+
+    # Different standard tags values - no logical types that have standard tags
+    nl_ltypes = {col_name: NaturalLanguage for col_name in sample_column_names}
+    schema.set_types(logical_types=nl_ltypes)
+    no_standard_tags_schema.set_types(logical_types=nl_ltypes)
+
+    assert schema.use_standard_tags != no_standard_tags_schema.use_standard_tags
+    assert schema == no_standard_tags_schema
```
tamargrey, line 162: Still not sure if different values for `use_standard_tags` should make two schemas (and therefore two table accessors) not equal or whether it should just depend on the actual semantic tag values.

I guess one problem with this current behavior is that you could then make the exact same changes to `schema` and `no_standard_tags_schema` (change a logical type to one with standard tags) and end up with schemas that are not equal.

tamargrey: Okay, changed so that we also make sure the `use_standard_tags` values are the same

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -338,34 +338,24 @@ def test_getitem_invalid_input(sample_df):
...
-                               semantic_tags=semantic_tags)
-    schema_df = sample_df.copy()
-    schema_df.ww.init(logical_types=logical_types,
-                      index='id',
-                      time_index='signup_date',
-                      semantic_tags=semantic_tags,
-                      already_sorted=True)
-    assert schema_df.ww == comparison_schema
+    # Confirm not equal with different schema but same data
+    copy_df.ww.set_index('id')
```
thehomebrewnerd, line 350: Should we do something like set a semantic tag or time index here instead since setting the index will also update the dataframe index for pandas at least? Just thinking we have both schema and dataframe differences here.

thehomebrewnerd (approved):
Looks pretty good. Just one minor item to consider.

woodwork/tests/accessor/test_table_accessor.py
tamargrey: oh yeah, good call. Changed to update the time index instead

thehomebrewnerd (approved):
LGTM!

PR #712, opened 2021-03-16
Add missing tests to Schema/Accessor

### LINKED ISSUE #708: Add missing tests to Schema/Accessor
There are some tests in the DataTables/DataColumn code that we don't seem to have corresponding tests in the Schema/Accessor code:
- Non string column names on table and column accessors - `test_datacolumn_init_with_falsy_name` in DataColumn tests has a numeric column name (this test isn't present in Accessor tests), and `test_datatable_init_with_numpy` in DataTable tests has numeric column names (we're not handling the numpy case yet). Probably best to have tests that check non string column names explicitly. 
- Move tests that explicitly check semantic tag update logic from `column_accessor` tests to `schema_column` tests so that they're more effectively testing for both column and table Accessors (since both use the schema_column logic).
  - duplicate tag logic
- Missed tests related to instantiated ltypes  - `test_select_instantiated`, `test_datetime_inference_with_format_param`
- missed statistical test - `test_datatable_describe_with_no_match`

tamargrey:
- Certain cases we were testing in DataTables weren't being tested in the Schema/Accessor approach, so this PR adds tests surrounding those cases.
- Closes #708

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -64,6 +64,7 @@ Release Notes
         * Update Customizing Type Inference guide to use accessor (:pr:`696`)
     * Documentation Changes
     * Testing Changes
+        * Add tests to Accessor/Schema that weren't previously covered (:pr:`712`)
 
     Thanks to the following people for contributing to this release:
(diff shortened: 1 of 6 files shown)


thehomebrewnerd (approved):
LGTM!

PR #716, opened 2021-03-16
Add accessor test that uses numeric_time_index_df

tamargrey:
- Now we can remove the DataTables test that uses this fixture without having to also remove the fixture, plus we add a useful test to Accessor


thehomebrewnerd (approved):
Looks good!

PR #717, opened 2021-03-17
Add Column Accessor Examples to Start Documentation

### LINKED ISSUE #702: Update docs to show usage of column accessor
The current documentation does not cover using the accessor on a series. We should update the documents to show how this works. Of particular importance is the use of the `init_series` function which needs to be used any time initialization would require changing the dtype of the series.

thehomebrewnerd:
- Add Column Accessor to Start Documentation
- Closes #702 

This PR adds a brief section to the Start documentation, reviewing how to initialize Woodwork on a Series. Also covered is how to use the `ww.init_series` function when a dtype change is required.


docs/source/start.ipynb
```diff
@@ -333,6 +333,67 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "As you can see from the output above, the specified tag has been added to the semantic tags for the series.\n",
+    "\n",
+    "In cases where the LogicalType would require the Series dtype to change, a helper function `ww.init_series` must be used. This function will return a new series object with Woodwork initialized and the dtype of the series changed to match the dtype of the LogicalType.\n",
+    "\n",
+    "To demonstrate this, first create a Series, with a `string` dtype. Then, initialize a Woodwork series with a `Categorical` logical type using the `init_series` function. The series that is returned will have Woodwork initialized with the LogicalType set to `Categorical` as expected."
```
tamargrey, line 383: might be good to remind users that the reason why init series is necessary is because `Categorical`'s physical type is `category`

docs/source/start.ipynb
```diff
@@ -333,6 +333,67 @@
     "df.ww"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Using Woodwork with a Series\n",
+    "\n",
+    "Woodwork also can be used to store typing information on a Series. There are two approaches for initializing Woodwork on a Series. If you have a Series that has a dtype that is compatible with the specified or inferred LogicalType, Woodwork can be initialized through the `ww` namespace, just as with DataFrames."
```
tamargrey, line 342: I wonder if we should be more explicit about what it means for a dtype to be "compatible" with a Logical Type.

docs/source/start.ipynb
```diff
@@ -333,6 +333,67 @@
...
+   "source": [
+    "series.ww.add_semantic_tags('new_tag')\n",
+    "series.ww"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "As you can see from the output above, the specified tag has been added to the semantic tags for the series.\n",
```
tamargrey, line 379: We should also mention that the pandas operations are accessible in the same way through the accessor with something like `series.ww.shape`

docs/source/start.ipynb
```diff
@@ -333,6 +333,67 @@
...
+    "\n",
+    "In cases where the LogicalType would require the Series dtype to change, a helper function `ww.init_series` must be used. This function will return a new series object with Woodwork initialized and the dtype of the series changed to match the dtype of the LogicalType.\n",
```
tamargrey, line 381: Similar to my comment on the word "compatible", I just want to make sure users will really understand what situations will require them to use this.

tamargrey (commented):
Looking good. Mostly had some thoughts on how clear it is to users that a dtype mismatch between their Series and its Logical Type is what will force them to use init_series.

docs/source/start.ipynb
thehomebrewnerd: I get your point here. Trying to think about how we can do that cleanly without launching into too much of a detailed explanation here, as that should be covered in the Understanding Types and Tags guide. Maybe it would be sufficient to refer users to that guide if they need more information?

tamargrey: Yeah, maybe a quick reference to the idea that the Series' dtype has to match the LogicalType's physical type, and if they need more information reference the types and tags guide

jeff-hernandez (commented):
Looking good! It could be helpful to cover how a column in a Woodwork initialized DataFrame can be set by a Woodwork initialized Series and how this affects the Series like the semantic tags.

thehomebrewnerd:
> Looking good! It could be helpful to cover how a column in a Woodwork initialized DataFrame can be set by a Woodwork initialized Series and how this affects the Series like the semantic tags.

I thought about this as well, but I think this guide is really intended to be a quick start guide, and I'm a little concerned adding that is maybe too much detail for this guide. I do agree that we should cover that somewhere though. I think I will create a new issue about reviewing our documentation to make sure we adequately cover using various parts of the API, and I can mention this there as well.

thehomebrewnerd:
> Looking good. Mostly had some thoughts on how clear it is to users that a dtype mismatch between their Series and its Logical Type is what will force them to use init_series.

@tamargrey Made several updates to the content and flow to incorporate your suggestions. Hopefully this all makes sense.

thehomebrewnerd:
> > Looking good! It could be helpful to cover how a column in a Woodwork initialized DataFrame can be set by a Woodwork initialized Series and how this affects the Series like the semantic tags.
> 
> I thought about this as well, but I think this guide is really intended to be a quick start guide, and I'm a little concerned adding that is maybe too much detail for this guide. I do agree that we should cover that somewhere though. I think I will create a new issue about reviewing our documentation to make sure we adequately cover using various parts of the API, and I can mention this there as well.

Added #721 for this

tamargrey (approved):
lgtm!

PR #719, opened 2021-03-17
Update release notes test to use correct branch name

gsheni:
- To be consistent across Featuretools, and Woodwork, the release branch name should be the same
  - https://github.com/alteryx/woodwork/blob/main/.github/workflows/release_notes_updated.yml#L18
  - https://github.com/alteryx/featuretools/blob/main/.github/workflows/release_notes_updated.yml#L18


rwedge (approved):
LGTM

PR #720, opened 2021-03-17
Change dtype of sample_series and remove convert_series util

### LINKED ISSUE #710: Update sample_series fixture to have proper dtype
Currently we call `convert_series(sample_series, Categorical)`any time we want to use the `sample_series` fixture because its dtype is `object`. If we change the dtype to be `categorical` for pandas and dask and `string` for koalas, we can greatly minimize `convert_dtype` usage (probably won't be able to fully remove because there are some places where we pass different logical types to convert_series)

tamargrey:
- Changes sample_series to have the dtype that's valid for the `Categorical` logical type, which it gets inferred as. This means we can remove the `convert_series` util as there were only 9 more uses of it, and 8 of them just required a simple `astype` call
- Closes #710


woodwork/tests/accessor/test_accessor_utils.py
```diff
@@ -21,14 +21,17 @@
 
 def test_init_series_valid_conversion_specified_ltype(sample_series):
     series = init_series(sample_series, logical_type='categorical')
-    assert series is not sample_series
+    assert series is sample_series
```
thehomebrewnerd, line 24: I think with the change to the dtype of `sample_series` we now never test converting a series dtype with Koalas like we did before. Maybe we should update this to convert sample series to have a different dtype at the start in this case, and leave the rest of the test the same?

I think this is one test where having sample_series not have a `category` dtype was useful.

woodwork/tests/accessor/test_accessor_utils.py
```diff
@@ -37,7 +40,7 @@ def test_init_series_valid_conversion_specified_ltype(sample_series):
 
 def test_init_series_valid_conversion_inferred_ltype(sample_series):
     series = init_series(sample_series)
-    assert series is not sample_series
+    assert series is sample_series
```
thehomebrewnerd, line 43: Same as previous comment - I think we want to verify that the dtype gets changed properly by `init_series` in this test.

woodwork/tests/accessor/test_accessor_utils.py
```diff
@@ -59,7 +62,7 @@ def test_init_series_all_parameters(sample_series):
                          metadata=metadata,
                          description=description,
                          use_standard_tags=False)
-    assert series is not sample_series
+    assert series is sample_series
```
thehomebrewnerd, line 65: Same as above

thehomebrewnerd (commented):
Looks pretty good, just a couple comments on the tests in `test_accessor_utils.py`

woodwork/tests/accessor/test_accessor_utils.py
tamargrey: ahh, got it. I can change the dtype back to `'object'` for this and the other init_series tests at the start to keep the test the same as it was before, if you think that makes sense

thehomebrewnerd: Yeah, I think that makes sense.

tamargrey: okay, changed!

thehomebrewnerd (approved):
LGTM!

PR #730, opened 2021-03-22
Use pd.read_csv for readme and get started

### LINKED ISSUE #727: Update README and Get Started to use pd.read_csv instead of ww.load_retail
In README.md and the Get Started guide, we should use `pd.read_csv` to get the initial dataframe rather than using `ww.load_retail`. This will make it more clear to users that we are starting from a regular pandas dataframe.

This will require the use of the `make_index` parameter when initializing Woodwork as the current `load_retail` function adds an index column to the DataFrame before returning it.

thehomebrewnerd:
- Use pd.read_csv for readme and get started
- Closes #727 

Updates Getting Started guide and REAME.md to use `pd.read_csv` instead of `ww.demo.load_retail` to better illustrate we are starting directly from a pandas DataFrame.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,10 +43,11 @@ conda install -c conda-forge woodwork
 Below is an example of using Woodwork. In this example, a sample dataset of order items is used to create a Woodwork `DataFrame`, specifying the `LogicalType` for four of the columns.
 
 ```python
+import pandas as pd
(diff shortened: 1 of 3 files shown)


gsheni (approved):
lgtm

PR #736, opened 2021-03-24
Add validation control for the table accessor

tamargrey:
- Adds a `validate` param to the WoodworkTableAccessor that disables all validation that occurs, passing `validate` down to the Schema class and `_get_column_schema`.

Note: Decided to open a PR after just the Table Accessor since it was touching a lot of areas of Woodwork already, and I want to make sure we give each component of the validation control changes enough review time. The Column Accessor validation control, removing validation from deserialization, and how we address this in documentation (if at all) can each be their own, separate PRs

### DIFF
--- a/woodwork/schema.py
+++ b/woodwork/schema.py
@@ -53,7 +53,7 @@ class Schema(object):
             use_standard_tags (bool, optional): If True, will add standard semantic tags to columns based
                 specified logical type for the column. Defaults to True.
             column_descriptions (dict[str -> str], optional): Dictionary mapping column names to column descriptions.
-            validate (bool, optional): Whether parameter validateion should occur. Defaults to True. Warning:
+            validate (bool, optional): Whether parameter validation should occur. Defaults to True. Warning:
                 Should be set to False only when parameters and data are known to be valid.
                 Any errors resulting from skipping validation with invalid inputs may not be easily understood.
         """
--- a/woodwork/schema.py
+++ b/woodwork/schema.py
@@ -26,7 +26,8 @@ class Schema(object):
                  table_metadata=None,
                  column_metadata=None,
                  use_standard_tags=True,
-                 column_descriptions=None):
+                 column_descriptions=None,
+                 validate=True):
         """Create Schema
 
         Args:
@@ -52,10 +53,14 @@ class Schema(object):
             use_standard_tags (bool, optional): If True, will add standard semantic tags to columns based
                 specified logical type for the column. Defaults to True.
             column_descriptions (dict[str -> str], optional): Dictionary mapping column names to column descriptions.
+            validate (bool, optional): Whether parameter validateion should occur. Defaults to True. Warning:
+                Should only be False when confident that parameters valid since any resulting errors may
+                not be easily understood.
         """
-        # Check that inputs are valid
-        _validate_params(column_names, name, index, time_index, logical_types,
-                         table_metadata, column_metadata, semantic_tags, column_descriptions)
+        if validate:
+            # Check that inputs are valid
+            _validate_params(column_names, name, index, time_index, logical_types,
+                             table_metadata, column_metadata, semantic_tags, column_descriptions)
 
         self.name = name
         self.use_standard_tags = use_standard_tags
@@ -66,12 +71,13 @@ class Schema(object):
                                             semantic_tags,
                                             use_standard_tags,
                                             column_descriptions,
-                                            column_metadata)
+                                            column_metadata,
+                                            validate)
         if index is not None:
-            self.set_index(index)
+            self.set_index(index, validate=validate)
 
         if time_index is not None:
-            self.set_time_index(time_index)
+            self.set_time_index(time_index, validate=validate)
 
         self.metadata = table_metadata or {}
 
@@ -273,14 +279,16 @@ class Schema(object):
                         semantic_tags,
                         use_standard_tags,
                         column_descriptions,
-                        column_metadata):
+                        column_metadata,
+                        validate):
         """Create a dictionary with column names as keys and new column dictionaries holding
         each column's typing information as values."""
         columns = {}
         for name in column_names:
             semantic_tags_for_col = _convert_input_to_set((semantic_tags or {}).get(name),
-                                                          error_language=f'semantic_tags for {name}')
-            _validate_not_setting_index_tags(semantic_tags_for_col, name)
+                                                          error_language=f'semantic_tags for {name}', validate=validate)
+            if validate:
+                _validate_not_setting_index_tags(semantic_tags_for_col, name)
             description = (column_descriptions or {}).get(name)
             metadata_for_col = (column_metadata or {}).get(name)
 
@@ -289,10 +297,11 @@ class Schema(object):
                                              semantic_tags=semantic_tags_for_col,
                                              use_standard_tags=use_standard_tags,
                                              description=description,
-                                             metadata=metadata_for_col)
+                                             metadata=metadata_for_col,
+                                             validate=validate)
         return columns
 
-    def set_index(self, new_index):
+    def set_index(self, new_index, validate=True):
         """Sets the index. Handles setting a new index, updating the index, or removing the index.
 
         Args:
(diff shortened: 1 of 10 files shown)


woodwork/table_accessor.py
```diff
@@ -88,8 +89,12 @@ def init(self,
                 Any other arguments provided will be ignored. Note that any changes made to the schema object after
                 initialization will propagate to the DataFrame. Similarly, to avoid unintended typing information changes,
                 the same schema object should not be shared between DataFrames.
+            validate (bool, optional): Whether parameter and data validation should occur. Defaults to True. Warning:
+                Should only be False when confident that parameters and data are valid since any resulting errors may
+                not be easily understood.
```
thehomebrewnerd, line 94: Wording suggestion: `Should be set to False only when parameters and data are known to be valid. Any errors resulting from skipping validation with invalid inputs may not be easily understood.`

woodwork/schema.py
```diff
@@ -52,10 +53,14 @@ def __init__(self, column_names,
             use_standard_tags (bool, optional): If True, will add standard semantic tags to columns based
                 specified logical type for the column. Defaults to True.
             column_descriptions (dict[str -> str], optional): Dictionary mapping column names to column descriptions.
+            validate (bool, optional): Whether parameter validateion should occur. Defaults to True. Warning:
+                Should only be False when confident that parameters valid since any resulting errors may
+                not be easily understood.
```
thehomebrewnerd, line 58: If accessor docstring is update, be sure to update here as well.

woodwork/tests/schema/test_schema_init.py
```diff
@@ -355,3 +355,27 @@ def test_schema_init_with_column_metadata(sample_column_names, sample_inferred_l
...
+
+def test_no_schema_validation_invalid_input(sample_column_names, sample_inferred_logical_types):
+    error = 'semantic_tags for id must be a string, set or list'
+    with pytest.raises(TypeError, match=error):
+        Schema(sample_column_names, sample_inferred_logical_types,
+               semantic_tags={'id': 4}, validate=True)
+    error = "'int' object has no attribute 'union'"
+    with pytest.raises(AttributeError, match=error):
+        Schema(sample_column_names, sample_inferred_logical_types,
+               semantic_tags={'id': 4}, validate=False)
```
thehomebrewnerd, line 381: I'm wondering if there is a better way to test these invalid input conditions. We don't really care so much about what error/message gets raised here...we only care that it is not coming from our Woodwork validation code. If the Python message were changed, this test would break, but everything would still be ok from our perspective.

One idea would be to use mocks for our validation functions (where possible) and use the mocks to check that our validation functions never get called instead of checking that some other non-Woodwork error is getting raised. Might want to think through this a little more to make sure we have a robust test.

thehomebrewnerd (commented):
Looking pretty good. I think we should think through how we test the validation skip condition a little more though. See code comment for more detail on my thoughts.

woodwork/tests/schema/test_schema_init.py
tamargrey: Totally agree. Was also feeling like it was a bit of a cop out to just check that one or two validation calls weren't getting called. Would be great to get something with better coverage. 

I need to read up more on mocks to understand what's possible, but I wonder if we could define a list of our validation functions (and make sure that validation is only ever happening with functions and not just inline) and confirm none of them get called with various input parameters being used.

Ideally we'd also have some way of raising an error if some new validation is added but it's not under validation control , we raise an error. I'll spend some time thinking about it.

tamargrey: Here's one option of how we could test whether specific validation functions are getting called:

```python
@patch("woodwork.schema._validate_params")
def test_validation_method_called(mock_validation, sample_column_names, sample_inferred_logical_types):
    assert not mock_validation.called

    not_validated_schema = Schema(sample_column_names, sample_inferred_logical_types, validate=False)
    assert not mock_validation.called

    validated_schema = Schema(sample_column_names, sample_inferred_logical_types, validate=True)
    assert mock_validation.called

    assert validated_schema == not_validated_schema
```

I think if we add this kind of test for every validation function that's getting skipped when `validate=False`, that should make for good coverage.

thehomebrewnerd: This approach seems solid to me and will allow us to verify that the validate parameter controls the function calls as we expect.

woodwork/schema.py
```diff
@@ -303,10 +312,11 @@ def set_index(self, new_index):
         if old_index is not None:
             self.remove_semantic_tags({old_index: 'index'})
```
tamargrey, line 313: There are some validation checks that happen in `remove_semantic_tags`, but as `old_index` will always be None at `init`, we can't reach this section of `set_index` (and the corresponding section in `set_time_index`) in Schema init. 

However, we're now exposing `validate` in `set_index` and `set_time_index` so, in theory, users could pass `validate=False` while setting a Schema's index directly, and we'd miss these validation checks.

However, since the Accessor `set_index` and `set_time_index` methods don't have to expose `validate`, I don't expect these methods to get called often, if at all, so it didn't seem worth the additional complexity.

thehomebrewnerd: I agree. I think the main idea here is to have a way to skip the computationally expensive validation checks, so if a few simple validation checks remain, I don't perceive that as a big issue. If we identify these remaining checks as performance bottlenecks later in the year when we do profiling work, we can work to improve/resolve then.

woodwork/utils.py
```diff
@@ -27,14 +27,14 @@ def camel_to_snake(s):
...
+def _convert_input_to_set(semantic_tags, error_language='semantic_tags', validate=True):
     """Takes input as a single string, a list of strings, or a set of strings
         and returns a set with the supplied values. If no values are supplied,
         an empty set will be returned."""
     if not semantic_tags:
         return set()
 
-    if type(semantic_tags) not in [list, set, str]:
+    if validate and type(semantic_tags) not in [list, set, str]:
         raise TypeError(f"{error_language} must be a string, set or list")
```
tamargrey, line 38: If we want to be able to use mocks to test whether this has been called, we'll want to pull this and the check below into their own function. 

It might be a bit excessive to pull these single lines into their own functions, and as we're not actually catching any errors when we don't validate and we just get unexpected results from this function, it might be so bad to not do that.

Anyway, I'd gone ahead to see what that would look like and was running into some unexpected behavior that I'll need to take another look at tomorrow to see if I can figure out if we do want to use mocks here as well.

woodwork/table_accessor.py
tamargrey: Made these changes here and in Schema

woodwork/utils.py
tamargrey: Turns out that in order to test exceptions that are raised by a function you've mocked, you need to specify the exceptions as a side effect. In the other tests where mocks have been added, we just test the non-error cases (for when we are validating), so instead of adding the side effect to a mocked `test_convert_input_to_set`, I just added a separate test to check validation control on `_convert_input_to_set`.

woodwork/schema.py
```diff
@@ -52,10 +53,14 @@ def __init__(self, column_names,
             use_standard_tags (bool, optional): If True, will add standard semantic tags to columns based
                 specified logical type for the column. Defaults to True.
             column_descriptions (dict[str -> str], optional): Dictionary mapping column names to column descriptions.
+            validate (bool, optional): Whether parameter validateion should occur. Defaults to True. Warning:
```
thehomebrewnerd, line 56: `validateion` -> `validation`

thehomebrewnerd (approved):
One minor spelling error and a conflict with release notes to resolve, but otherwise looks good to me!

woodwork/schema.py
tamargrey: thanks for catching!

woodwork/utils.py
```diff
@@ -27,28 +27,38 @@ def camel_to_snake(s):
...
 
-    if not all([isinstance(tag, str) for tag in semantic_tags]):
-        raise TypeError(f"{error_language} must contain only strings")
+    if validate:
+        _validate_string_tags(semantic_tags, error_language)
 
     return semantic_tags
 
 
+def _validate_tags_input(semantic_tags, error_language):
```
thehomebrewnerd, line 52: Maybe change this to `_validate_tags_input_type` to be a little more clear what we are checking here?

thehomebrewnerd (approved):
Looks good, just one suggestion about function name, but it's not a big deal.

woodwork/utils.py
tamargrey: changed!

PR #741, opened 2021-03-25
Rename ZIPCode to PostalCode

### LINKED ISSUE #164: Rename ZIPCode logical type to PostalCode
`ZIPCode` has limited meaning outside of the United States, so this logical type should be renamed to something more generic such as `PostalCode` to have more global applicability.

thehomebrewnerd:
- Rename ZIPCode to PostalCode
- Closes #164 

Renames ZIPCode logical type to PostalCode. Also had to update one test because the renaming changed the order or output from the `ww.list_logical_types()` function.

### DIFF
--- a/woodwork/statistics_utils.py
+++ b/woodwork/statistics_utils.py
@@ -168,7 +168,7 @@ def _get_mutual_information_dict(dataframe, num_bins=10, nrows=None, include_ind
     """Calculates mutual information between all pairs of columns in the DataFrame that
     support mutual information. Logical Types that support mutual information are as
     follows:  Boolean, Categorical, CountryCode, Datetime, Double, Integer, Ordinal,
-    SubRegionCode, and ZIPCode
+    SubRegionCode, and PostalCode
 
     Args:
         dataframe (pd.DataFrame): Data containing Woodwork typing information
--- a/woodwork/statistics_utils.py
+++ b/woodwork/statistics_utils.py
@@ -168,7 +168,7 @@ def _get_mutual_information_dict(dataframe, num_bins=10, nrows=None, include_ind
     """Calculates mutual information between all pairs of columns in the DataFrame that
     support mutual information. Logical Types that support mutual information are as
     follows:  Boolean, Categorical, CountryCode, Datetime, Double, Integer, Ordinal,
-    SubRegionCode, and PostalCode
+    PostalCode, and SubRegionCode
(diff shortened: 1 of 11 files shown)


woodwork/statistics_utils.py
```diff
@@ -168,7 +168,7 @@ def _get_mutual_information_dict(dataframe, num_bins=10, nrows=None, include_ind
     """Calculates mutual information between all pairs of columns in the DataFrame that
     support mutual information. Logical Types that support mutual information are as
     follows:  Boolean, Categorical, CountryCode, Datetime, Double, Integer, Ordinal,
-    SubRegionCode, and ZIPCode
+    SubRegionCode, and PostalCode
```
tamargrey, line 171: I think we're trying to do alphabetical order here, so we should probably switch `SubRegionCode` and `PostalCode`

woodwork/table_accessor.py
```diff
@@ -681,7 +681,7 @@ def mutual_information_dict(self, num_bins=10, nrows=None, include_index=False):
         Calculates mutual information between all pairs of columns in the DataFrame that
         support mutual information. Logical Types that support mutual information are as
         follows:  Boolean, Categorical, CountryCode, Datetime, Double, Integer, Ordinal,
-        SubRegionCode, and ZIPCode
+        SubRegionCode, and PostalCode
```
tamargrey, line 684: here too

woodwork/table_accessor.py
```diff
@@ -708,7 +708,7 @@ def mutual_information(self, num_bins=10, nrows=None, include_index=False):
         """Calculates mutual information between all pairs of columns in the DataFrame that
         support mutual information. Logical Types that support mutual information are as
         follows:  Boolean, Categorical, CountryCode, Datetime, Double, Integer, Ordinal,
-        SubRegionCode, and ZIPCode
+        SubRegionCode, and PostalCode
```
tamargrey, line 711: here too (maybe it's a little silly that we have this docstring in three separate places)

tamargrey (approved):
small comment about alphabetical order, but lgmt other than that!

woodwork/statistics_utils.py
thehomebrewnerd: Yep, I rearranged other places to keep the order, but missed this. I'll update these before merging.

PR #768, opened 2021-04-01
Drop python 3.6 support

### LINKED ISSUE #530: Drop Python 3.6 Support
- As numpy 1.20.0 drops support for python 3.6, we will be dropping support for it in an upcoming release.

tamargrey:
- Closes #530


docs/source/release_notes.rst
```diff
@@ -11,6 +11,7 @@ Release Notes
         * Rename ``ZIPCode`` logical type to ``PostalCode`` (:pr:`741`)
         * Fix issue with smart-open version 5.0.0 (:pr:`750`, :pr:`758`)
         * Update minimum scikit-learn version to 0.22 (:pr:`763`)
+        * Drop support for Python version 3.6 (:pr:`768`)
```
gsheni, line 14: Let's add something like this below Future Release. We did this [for Featuretools](https://github.com/alteryx/featuretools/blob/main/docs/source/release_notes.rst)
```
    .. warning::
        The next non-bugfix release of Featuretools will not support Python 3.6
```

tamargrey: Does that mean that dropping 3.6 support shouldn't go into the upcoming release, then?

gsheni: Sorry, I mean we should put something like this
```
    .. warning::
        This release Woodwork does not support Python 3.6
```

gsheni (approved):

PR #777, opened 2021-04-02
Remove ColumnNameMismatch warning

### LINKED ISSUE #757: Remove ColumnNameMismatchWarning
- The warning raised in ColumnNameMismatchWarning when assigning/create a Series that does not have the same name. However, this does not match pandas's behavior, and thus should be removed (since we now using the accessor approach)
- There is only one place where this warning is [used](https://github.com/alteryx/woodwork/blob/36e68ed25119cf92d161612f1704c325631f4369/woodwork/table_accessor.py#L203).

thehomebrewnerd:
- Remove ColumnNameMismatch warning
- Closes #757 

Previously Woodwork would raise a warning if assigning a named series to a dataframe column of a different name through the Woodwork accessor. Since the user is explicitly supplying the name they want during the column assignment, this warning is not necessary and can be removed.


jeff-hernandez (approved):
Looks good

gsheni (approved):

PR #780, opened 2021-04-05
Store make index on WoodworkTableAccessor

### LINKED ISSUE #744: Store value of `make_index` on WoodworkTableAccessor
- As a Featuretools user, I wish I could use access the value of `make_index` that was used when initializing Woodwork. This would allow for some performance improvement in the Featuretools `EntitySet.concat` method.

Note: This is not a blocking issue for Featuretools, but would be nice to have.

tamargrey:
- adds a `WoodworkTableAccessor.make_index` attribute
- Also includes the `make_index` value in the equality check
- Closes #744

### DIFF
--- a/woodwork/table_accessor.py
+++ b/woodwork/table_accessor.py
@@ -94,6 +94,10 @@ class WoodworkTableAccessor:
         """
         if validate:
             _validate_accessor_params(self._dataframe, index, make_index, time_index, logical_types, schema)
+
+        # --> Set the made_index value here
+        self.made_index = make_index
+
         if schema is not None:
             self._schema = schema
             extra_params = []
@@ -146,6 +150,10 @@ class WoodworkTableAccessor:
 
     def __eq__(self, other):
         if self._schema != other.ww._schema:
+            return False
+
+        # --> make sure to thest this in equality test
+        if self.made_index != other.ww.made_index:
             return False
 
         # Only check pandas DataFrames for equality
@@ -580,6 +588,9 @@ class WoodworkTableAccessor:
                     else:
                         copied_schema = self._schema._get_subset_schema(list(self._dataframe.columns))
                         result.ww.init(schema=copied_schema)
+                        print('og made', self.made_index)
+                        result.ww.made_index = self.made_index
+                        print('new made', result.ww.made_index)
                 else:
                     # Confirm that the Schema is still valid on original DataFrame
                     # Important for inplace operations
--- a/woodwork/table_accessor.py
+++ b/woodwork/table_accessor.py
@@ -96,7 +96,7 @@ class WoodworkTableAccessor:
             _validate_accessor_params(self._dataframe, index, make_index, time_index, logical_types, schema)
 
         # --> Set the made_index value here
-        self.made_index = make_index
+        self.make_index = make_index
 
         if schema is not None:
             self._schema = schema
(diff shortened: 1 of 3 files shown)


woodwork/table_accessor.py
```diff
@@ -144,7 +145,12 @@ def init(self,
             if self._schema.time_index is not None:
                 self._sort_columns(already_sorted)
 
+        self.make_index = make_index
```
tamargrey, line 148: Put this at the end of the `init` method in order to also have a value for when a schema is passed in. (I'm assuming we want to be able to have this value stored even when initializing with a schema)

It's a bit of a weird situation because we warn users against passing additional parameters like `make_index` when also passing a `schema`, and say that we ignore them, but now we're technically not ignoring `make_index` (though passing `make_index=True, schema=schema` won't impact the underlying index at all--in fact it'd error in parameter validation).

However, it wouldn't really be useful for users to be able to pass a `make_index` value when also passing a `schema` because we won't also let them set a new index value.

woodwork/table_accessor.py
```diff
@@ -580,6 +586,7 @@ def wrapper(*args, **kwargs):
                     else:
                         copied_schema = self._schema._get_subset_schema(list(self._dataframe.columns))
                         result.ww.init(schema=copied_schema)
+                        result.ww.make_index = self.make_index
```
tamargrey, line 589: Another slightly odd situation here. 

Since I don't want to pass `make_index=True` to the `init` above this, I figure it's better to set the value after since I might expect a user to want that value propagated to the new DataFrame.

woodwork/table_accessor.py
```diff
@@ -144,7 +144,13 @@ def init(self,
             if self._schema.time_index is not None:
                 self._sort_columns(already_sorted)
 
+        # Record the value of
```
thehomebrewnerd, line 147: Comment needed?

woodwork/table_accessor.py
thehomebrewnerd: Do you think it would make more sense to always set `self.make_index` to `False` (the default) inside the `if schema is not None` block, and then set it to the user supplied value in the `else` block? That way I think the behavior would be more consistent with our warning.

If that change is made, might be good to add a comment explaining why it is done that way.

woodwork/table_accessor.py
thehomebrewnerd: Makes sense to me.

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -358,6 +358,19 @@ def test_accessor_equality(sample_df):
...
+    made_index_df = sample_df_pandas.copy()
+    made_index_df.insert(0, 'new_index', range(len(made_index_df)))
+    made_index_df.ww.init(index='new_index', make_index=False)
+
+    schema_df = sample_df_pandas.copy()
+    schema_df.ww.init(index='new_index', make_index=True)
+
+    pd.testing.assert_frame_equal(schema_df, made_index_df)
+
+    assert made_index_df.ww != schema_df
```
thehomebrewnerd, line 371: I think there is a missing `ww` on `schema_df`:  `assert made_index_df.ww != schema_df.ww`

woodwork/table_accessor.py
tamargrey: that seems like a good idea.... definitely more consistent with our warning.

woodwork/tests/accessor/test_table_accessor.py
tamargrey: yep!

woodwork/table_accessor.py
tamargrey: yeah, moved the place where we set `make_index` at your suggestion and added a more descriptive comment there

thehomebrewnerd (approved):
Looks good!

PR #782, opened 2021-04-05
Set _get_column_dict parameter use_standard_types to False by default

### LINKED ISSUE #776: Change default use_standard_tags for _get_column_dict to False
- This will allow the future ColumnSchema to work as primitive input and return types without adding unnecessary standard tags 
- Table/Column Accessors should still be `True` for use_standard_tags

rwedge:
Fixes #776 

`_get_column_dict` will have `use_standard_types=False` by default
`WoodworkColumnAccessor.__init__` will still have `use_standard_types=True` by default

swap two test cases based on this change

### DIFF
--- a/woodwork/schema_column.py
+++ b/woodwork/schema_column.py
@@ -23,7 +23,7 @@ def _get_column_dict(name,
         logical_type (str, LogicalType): The column's LogicalType.
         semantic_tags (str, list, set, optional): The semantic tag(s) specified for the column.
         use_standard_tags (boolean, optional): If True, will add standard semantic tags to the column based
-                specified logical type. Defaults to True.
+                on the specified logical type. Defaults to False.
         description (str, optional): User description of the column.
         metadata (dict[str -> json serializable], optional): Extra metadata provided by the user.
         validate (bool, optional): Whether to perform parameter validation. Defaults to True.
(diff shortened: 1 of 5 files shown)


woodwork/schema_column.py
```diff
@@ -13,7 +13,7 @@
 def _get_column_dict(name,
                      logical_type,
                      semantic_tags=None,
-                     use_standard_tags=True,
+                     use_standard_tags=False,
```
tamargrey, line 16: We should also change the docstring which has `If True, will add standard semantic tags to the column based
                specified logical type. Defaults to True.`

rwedge: > If True, will add standard semantic tags to the column based specified logical type

@tamargrey should this be "to the column based on the specified logical type"?

tamargrey: yes that definitely is a lot more clear! (The same typo is on the table Schema docstring and the TableAccessor docstring as well, I just noticed 🤦‍♀️ )

woodwork/table_accessor.py
```diff
@@ -82,7 +82,7 @@ def init(self,
             column_metadata (dict[str -> dict[str -> json serializable]], optional): Dictionary mapping column names
                 to that column's metadata dictionary.
             use_standard_tags (bool, optional): If True, will add standard semantic tags to columns based
-                specified logical type for the column. Defaults to True.
+                on the specified logical type for the column. Defaults to True.
```
tamargrey, line 85: Can we also change the `__init__` docstring in `schema.py` with this typo?

rwedge: done

tamargrey (approved):
lgtm!

PR #783, opened 2021-04-05
Add exclude parameter to select method

### LINKED ISSUE #465: Add `exclude` parameter for `select`
- As a user, I wish I could use Woodwork to selectively exclude specific types of data.

#### Code Example

```python

# get all non float columns
X.ww.select(exclude=['double'])

```

thehomebrewnerd:
- Add exclude parameter to WoodworkTableAccessor select method
- Closes #465 

Adds the ability to specify column types/tags to exclude when calling `df.ww.select`. Adds checks that one, but not both, of `include` and `exclude` is specified. Updated terminology in `select` to align with both options sharing much of the same code.

Note: Because of the large amount of shared code when using `include` or `exclude` some of the tests for `include` were not replicated for `exclude`.

### DIFF
--- a/woodwork/schema.py
+++ b/woodwork/schema.py
@@ -396,12 +396,12 @@ class Schema(object):
         assert not (include and exclude), "Cannot specify both include and exclude"
         if include and not isinstance(include, list):
             include = [include]
-        if exclude and not isinstance(exclude, list):
+        elif exclude and not isinstance(exclude, list):
             exclude = [exclude]
 
         if include is not None:
             selectors = include
-        else:
+        elif exclude is not None:
             selectors = exclude
 
         ltypes_used = set()
@@ -410,7 +410,7 @@ class Schema(object):
         tags_used = set()
         tags_in_schema = {tag for col in self.columns.values() for tag in col['semantic_tags']}
 
-        match_cols = set()
+        col_name_matches = set()
 
         for selector in selectors:
             # Determine if the selector is a registered, uninstantiated LogicalType
@@ -438,16 +438,19 @@ class Schema(object):
                 tags_used.add(selector)
             # Determine if the selector is a column name
             if col_names and selector in self.columns:
-                match_cols.add(selector)
+                col_name_matches.add(selector)
 
+        cols_to_return = set()
         for col_name, col in self.columns.items():
-            if _get_ltype_class(col['logical_type']) in ltypes_used or col['semantic_tags'].intersection(tags_used):
-                match_cols.add(col_name)
+            is_match = (_get_ltype_class(col['logical_type']) in ltypes_used or
+                        col['semantic_tags'].intersection(tags_used) or
+                        col_name in col_name_matches)
+            if include is not None and is_match:
+                cols_to_return.add(col_name)
+            elif exclude is not None and not is_match:
+                cols_to_return.add(col_name)
 
-        if include is not None:
-            return list(match_cols)
-        else:
-            return [col for col in self.columns.keys() if col not in match_cols]
+        return list(cols_to_return)
 
     def _get_subset_schema(self, subset_cols):
         """Creates a new Schema with specified columns, retaining typing information.
--- a/woodwork/schema.py
+++ b/woodwork/schema.py
@@ -377,20 +377,32 @@ class Schema(object):
     def _set_time_index_tags(self, time_index):
         self.columns[time_index]['semantic_tags'].add('time_index')
 
-    def _filter_cols(self, include, col_names=False):
+    def _filter_cols(self, include=None, exclude=None, col_names=False):
         """Return list of columns filtered with any of: semantic tags, LogicalTypes, column names
 
         Args:
             include (str or LogicalType or list[str or LogicalType]): parameter or list of parameters to
-                filter columns by. Can be Logical Types or Semantic Tags.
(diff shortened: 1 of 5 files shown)


woodwork/schema.py
```diff
@@ -377,30 +377,42 @@ def _set_index_tags(self, index):
...
+                filter columns by. Can be Logical Types or Semantic Tags. Columns that match will be
+                excluded from the returned list of columns.
 
             col_names (bool): Specifies whether to filter columns by name. Defaults to False.
 
         Returns:
             List[str] of column names that fit into filter.
         """
-        if not isinstance(include, list):
+        assert not (include and exclude), "Cannot specify both include and exclude"
```
tamargrey, line 396: I like the assertion here that they're not both present. Would make sense to also have the assertion that they're not both absent?

woodwork/schema.py
```diff
@@ -426,13 +438,16 @@ def _filter_cols(self, include, col_names=False):
...
         for col_name, col in self.columns.items():
             if _get_ltype_class(col['logical_type']) in ltypes_used or col['semantic_tags'].intersection(tags_used):
-                cols_to_include.add(col_name)
+                match_cols.add(col_name)
 
-        return list(cols_to_include)
+        if include is not None:
+            return list(match_cols)
+        else:
+            return [col for col in self.columns.keys() if col not in match_cols]
```
tamargrey, line 450: I know the opposite of the `include` block is going to be the `exclude` logic and that we've already confirmed that there will be either `include` or `exclude` present, but I think it might be nice to do `elif exclude is not None:` to remind anyone reading the code here that exclude is the other behavior here.

woodwork/schema.py
```diff
@@ -426,13 +438,16 @@ def _filter_cols(self, include, col_names=False):
                 tags_used.add(selector)
             # Determine if the selector is a column name
             if col_names and selector in self.columns:
-                cols_to_include.add(selector)
+                match_cols.add(selector)
 
         for col_name, col in self.columns.items():
             if _get_ltype_class(col['logical_type']) in ltypes_used or col['semantic_tags'].intersection(tags_used):
-                cols_to_include.add(col_name)
+                match_cols.add(col_name)
```
tamargrey, line 445: What would you think of sticking with a `cols_to_include` list of sorts (though maybe need a better name) and just determine what cols get added to it by whether or not we're excluding or including:

```python
for col_name, col in self.columns.items():
    is_match = get_ltype_class(col['logical_type']) in ltypes_used or col['semantic_tags'].intersection(tags_used)
    if include and is_match:
         cols_to_include.add(col_name)
    elif exclude and not is_match:
         cols_to_include.add(col_name)         
```

Really the only benefit of this way is that it lets us avoid iterating over the columns again for `exclude` below (which is only a problem if there's a lot of columns), so I don't know that there's a huge benefit to this way, but I thought I'd bring it up!

woodwork/tests/schema/test_schema.py
```diff
@@ -226,6 +253,9 @@ def test_filter_schema_cols_no_matches(sample_column_names, sample_inferred_logi
     filter_non_string = schema._filter_cols(include=1)
     assert filter_non_string == []
 
+    filter_exclude_all = schema._filter_cols(exclude=list(sample_inferred_logical_types.values()))
+    assert filter_exclude_all == []
```
tamargrey, line 257: I think this might be equivalent to the check in the table accessor tests where we check 
`assert len(schema_df.ww.select(exclude=all_types).columns) == 0`

In this case "no matches" might be the `schema._filter_cols(exclude=[])` return the full list of columns as opposed to "everything matches" returning no columns

tamargrey (commented):
looking good overall! Just a few suggestions

woodwork/schema.py
thehomebrewnerd: My reasoning for putting this assertion here is help guard against an unintended behavior and to make sure the logic that follows is valid. If both are specified, there will not be any errors raised without this assertion, so the assertion is really a protection for us as developers to highlight we did something wrong.

If both are `None` an error will result - albeit not a helpful error, but since it is a private method I didn't think that was a big deal.

With all that, I was thinking the second assertion wasn't needed, but there is no harm in adding it (aside from one more line of code) if you think it would be helpful.

tamargrey: ahh I see--I think that raising an error when there would otherwise be none is a good reason for the Assertion, and I see your point of not adding the second one if it'll raise an error on its own. Fine not adding!

woodwork/schema.py
thehomebrewnerd: Sure - make the code a little more explicit which is good. Made that change here and above as well.

woodwork/tests/schema/test_schema.py
thehomebrewnerd: Ah, I see. I was reading this initially as testing for an empty list being returned instead of "no columns found matching the specified criteria" which I now see is the intent. I'll update.

woodwork/schema.py
thehomebrewnerd: Good thought. Had to make some other updates to account for the logic around column name matches, but should now  iterate through the columns only once.

tamargrey (approved):
lgtm!

PR #786, opened 2021-04-06
Make `logical_type` and `name` params to `_get_column_dict` optional

### LINKED ISSUE #775: Allow null `name` and `logical_type` values for `_get_column_dict`
- In order for featuretools' Primitives to use the future `ColumnSchema` for input and return types, we need to handle a column schema with no `name` or `logical_type` attribute.
- However we still want these values when they're passed in from the column and table Accessors, so we should make sure this doesn't introduce undexpected behavior

tamargrey:
- Will allow ColumnSchema to be useful for Featuretools by allowing either/or logical type and semantic tag definitions.
- Closes #775

### DIFF
--- a/woodwork/schema_column.py
+++ b/woodwork/schema_column.py
@@ -19,11 +19,11 @@ def _get_column_dict(name=None,
                      validate=True):
     """Creates a dictionary that contains the typing information for a Schema column
     Args:
-        name (str): The name of the column.
-        logical_type (str, LogicalType): The column's LogicalType.
+        name (str, optional): The name of the column.
+        logical_type (str, LogicalType, optional): The column's LogicalType.
         semantic_tags (str, list, set, optional): The semantic tag(s) specified for the column.
         use_standard_tags (boolean, optional): If True, will add standard semantic tags to the column based
-                on the specified logical type. Defaults to False.
+                on the specified logical type if a logical type is defined for the column. Defaults to False.
         description (str, optional): User description of the column.
         metadata (dict[str -> json serializable], optional): Extra metadata provided by the user.
         validate (bool, optional): Whether to perform parameter validation. Defaults to True.
@@ -31,7 +31,8 @@ def _get_column_dict(name=None,
     if metadata is None:
         metadata = {}
     if validate:
-        _validate_logical_type(logical_type)
+        if logical_type is not None:
+            _validate_logical_type(logical_type)
         _validate_description(description)
         _validate_metadata(metadata)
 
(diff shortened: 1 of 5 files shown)


woodwork/schema_column.py
```diff
@@ -46,12 +46,13 @@ def _get_column_dict(name,
 
 
 def _validate_logical_type(logical_type):
-    ltype_class = _get_ltype_class(logical_type)
+    if logical_type is not None:
+        ltype_class = _get_ltype_class(logical_type)
```
thehomebrewnerd, line 50: Wondering if it would be better to only call this function if there is a non-`None` logical type supplied in `_get_column_dict` instead of updating this to handle the `None` case? Is there an advantage to this approach?

woodwork/tests/schema/test_schema_column.py
```diff
@@ -27,14 +27,12 @@
 
 
 def test_validate_logical_type_errors():
+    assert _validate_logical_type(None) is None
+
```
thehomebrewnerd, line 31: If we only call this function with non-`None` inputs we can remove this assertion and leave in the removed block below.

woodwork/tests/schema/test_schema_column.py
```diff
@@ -109,6 +107,32 @@ def test_get_column_dict_params():
...
+        'description': None,
+        'metadata': {}
+    }
+    assert _get_column_dict() == empty_col
+
+    just_tags = _get_column_dict(semantic_tags={'numeric', 'time_index'})
+    assert just_tags.get('logical_type') is None
+    assert just_tags.get('semantic_tags') == {'numeric', 'time_index'}
+
+    just_tags = _get_column_dict(logical_type=Integer)
```
thehomebrewnerd, line 123: should this be called `just_ltype` instead?

woodwork/tests/schema/test_schema_column.py
```diff
@@ -109,6 +107,32 @@ def test_get_column_dict_params():
...
+    assert just_tags.get('logical_type') == Integer
+    assert just_tags.get('semantic_tags') == set()
+
+    error = "Cannot use standard tags when logical_type is None"
+    with pytest.raises(ValueError, match=error):
+        _get_column_dict(semantic_tags='categorical', use_standard_tags=True)
+
+    error = "semantic_tags for None must be a string, set or list"
```
thehomebrewnerd, line 131: Think we should update this error message to only include the "for None" part if a name has been defined, otherwise just "semantic_tags must be a string, set or list"?

woodwork/schema_column.py
```diff
@@ -10,8 +10,8 @@
 from woodwork.utils import _convert_input_to_set
 
 
-def _get_column_dict(name,
-                     logical_type,
+def _get_column_dict(name=None,
+                     logical_type=None,
```
thehomebrewnerd, line 14: Not a big deal since this is a private function, but we should probably also update the docstring to call out `name` and `logical_type` as optional inputs now.

thehomebrewnerd (commented):
Looks pretty good...just a few things to consider.

woodwork/schema_column.py
tamargrey: Hmm I guess I was thinking that we send null values to `_validate_accessor_params`, but even there we don't pass possibly null values to the actual `_check_time_index` type functions, just the larger validate function. I'm fine with pulling the check outside of the validate_logical_type function

woodwork/tests/schema/test_schema_column.py
tamargrey: yep, changing

woodwork/tests/schema/test_schema_column.py
tamargrey: Yes, that seems like a better way to handle this error message

woodwork/schema_column.py
tamargrey: also added a line to the `use_standard_tags` docstring about needing a logical type

thehomebrewnerd (approved):
Looks good!

PR #788, opened 2021-04-06
Add deserialization validation control

### LINKED ISSUE #739: Avoid validating Woodwork Table at deserialization
Validation can be computationally expensive, and when we're deserializing a Woodwork Table, we know that it was once a valid Woodwork Table, so validation should not be necessary. 

We should confirm that validation is, indeed, not necessary, and then set `validate=False` upon initializing at deserialization.

thehomebrewnerd:
- Add deserialization validation control
- Closes #739 

Adds validation control to `deserialize.read_woodwork_table`, which defaults to `False`. Also adds validation control to `ww.read_csv` which defaults to True.

### DIFF
--- a/woodwork/deserialize.py
+++ b/woodwork/deserialize.py
@@ -136,7 +136,8 @@ def read_woodwork_table(path, profile_name=None, validate=False, **kwargs):
             profile_name (str, bool): The AWS profile specified to write to S3. Will default to None and search for AWS credentials.
                 Set to False to use an anonymous profile.
             validate (bool, optional): Whether parameter and data validation should occur when initializing Woodwork dataframe
-                during deserialization. Defaults to False.
+                during deserialization. Defaults to False. Note: If serialized data was modified outside of Woodwork and you
+                are unsure of the validity of the data or typing information, `validate` should be set to True.
             kwargs (keywords): Additional keyword arguments to pass as keyword arguments to the underlying deserialization method.
 
         Returns:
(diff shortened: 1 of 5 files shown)


woodwork/deserialize.py
```diff
@@ -121,18 +122,21 @@ def _typing_information_to_woodwork_table(table_typing_info, **kwargs):
...
-def read_woodwork_table(path, profile_name=None, **kwargs):
+def read_woodwork_table(path, profile_name=None, validate=False, **kwargs):
     """Read Woodwork table from disk, S3 path, or URL.
 
         Args:
             path (str): Directory on disk, S3 path, or URL to read `woodwork_typing_info.json`.
             profile_name (str, bool): The AWS profile specified to write to S3. Will default to None and search for AWS credentials.
                 Set to False to use an anonymous profile.
+            validate (bool, optional): Whether parameter and data validation should occur when initializing Woodwork dataframe
+                during deserialization. Defaults to False.
```
tamargrey, line 139: I'm not sure when a user might run into a situation where they'd have changed the serialized data in a way where they'd want to validate, but might be worth a warning/message to the user that if they're not sure that their data is still valid, they should set `validate=True`

thehomebrewnerd: Yeah, that's a fair point. I was not thinking that users would change the serialized data, but I'm sure that's something that could happen. I'll add that to the docstring.

thehomebrewnerd: Updated!

tamargrey (approved):
lgtm!

gsheni (approved):
lgtm

PR #791, opened 2021-04-07
Add ColumnSchema object to Woodwork

### LINKED ISSUE #770: Add `ColumnSchema` class to be used by WoodworkColumnAccessor
The contents of `schema_column.py` should be moved from independent functions into a pandas object called `ColumnSchema` that will be used by both the WoodworkColumnAccessor and the WoodworkTableAccessor. Before this issue gets handled, there are a number of changes that will be made to accommodate this change.

- Add ColumnSchema object
- use `_get_column_dict` logic in the `__init__` of ColumnSchema
- Move any schema-specific logic to the ColumnSchema class 
    - (doesn't need to happen for functionality but should happen for consistency with `TableAccessor`) + 
    - pull relevant testing logic to col schema tests as well
- Add a `WoodworkColumnAccessor.schema` property that returns a copy of the schema object
- When creating columns, create ColumnSchema objects instead of dictionaries 
    - Instead of `df.ww.columns` being a dictionary of str → dict it’ll be str → `ColumnSchema`
- Update everywhere that we were accessing column elements with dictionary get item logic
  - ['logical_type'] --> .logical_type
- This issue handle the separation between WoodworkColumnAccessor and ColumnSchema. 
  - Currently, all the logic lives on WoodworkColumnAccessor. 
  - With this issue, some of the logic should move from the WoodworkColumnAccessor to the new ColumnSchema. 
  - There may some methods that just call the ColumnSchema (from the WoodworkColumnAccessor).

tamargrey:
- Creates a ColumnSchema object that's used by the `WoodworkColumnAccessor` and the `TableSchema` to store typing information at the column level
- Closes #770

### DIFF
--- a/woodwork/column_accessor.py
+++ b/woodwork/column_accessor.py
@@ -63,6 +63,7 @@ class WoodworkColumnAccessor:
                                     description=description,
                                     metadata=metadata)
 
+    # --> should be on ColumnSchema object???
     @property
     def description(self):
         """The description of the series"""
@@ -137,6 +138,7 @@ class WoodworkColumnAccessor:
             _raise_init_error()
         return self._schema.logical_type
 
+    # --> should be on column schema object??
     @property
     def metadata(self):
         """The metadata of the series"""
--- a/woodwork/column_accessor.py
+++ b/woodwork/column_accessor.py
@@ -55,8 +55,7 @@ class WoodworkColumnAccessor:
         self._validate_logical_type(logical_type)
         self.use_standard_tags = use_standard_tags
 
-        self._schema = ColumnSchema(name=self._series.name,
-                                    logical_type=logical_type,
+        self._schema = ColumnSchema(logical_type=logical_type,
                                     semantic_tags=semantic_tags,
                                     use_standard_tags=self.use_standard_tags,
                                     description=description,
--- a/woodwork/column_accessor.py
+++ b/woodwork/column_accessor.py
@@ -55,7 +55,6 @@ class WoodworkColumnAccessor:
         self._validate_logical_type(logical_type)
         self.use_standard_tags = use_standard_tags
 
-# --> add a schema property below this that returns a copy of the schema
         self._schema = ColumnSchema(name=self._series.name,
                                     logical_type=logical_type,
                                     semantic_tags=semantic_tags,
@@ -63,7 +62,6 @@ class WoodworkColumnAccessor:
                                     description=description,
                                     metadata=metadata)
 
-    # --> should be on ColumnSchema object???
     @property
     def description(self):
         """The description of the series"""
@@ -138,7 +136,6 @@ class WoodworkColumnAccessor:
             _raise_init_error()
         return self._schema.logical_type
 
-    # --> should be on column schema object??
     @property
     def metadata(self):
         """The metadata of the series"""
@@ -201,7 +198,6 @@ class WoodworkColumnAccessor:
                 if _is_series(result):
                     valid_dtype = _get_valid_dtype(type(result), self._schema.logical_type)
                     if str(result.dtype) == valid_dtype:
-                        # --> add a way of initializing with a schema??
                         result.ww.init(logical_type=self._schema.logical_type,
                                        semantic_tags=copy.deepcopy(self._schema.semantic_tags),
                                        description=self._schema.description,
--- a/woodwork/column_accessor.py
+++ b/woodwork/column_accessor.py
@@ -5,8 +5,8 @@ import pandas as pd
 
 from woodwork.accessor_utils import _get_valid_dtype, _is_series, init_series
 from woodwork.column_schema import (
+    ColumnSchema,
     _add_semantic_tags,
-    _get_column_dict,
     _remove_semantic_tags,
     _reset_semantic_tags,
     _set_semantic_tags,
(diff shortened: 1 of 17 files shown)


tamargrey:
Note: At this point, this PR contains the minimal amount of changes to include `ColumnSchema` across Woodwork. There are further changes that should be made in order to have the `ColumnSchema` properly integrated as an object that follows the Accessor-Schema paradigm. They can either go into this PR or be separate issues:

  - Should we store `use_standard_tags` as a value on the `ColumnSchema` instead of `WoodworkColumnAccessor`?
      - If we do, remove it from the ColumnAccessor
      - A reason to not add it is that then we’d be storing `use_standard_tags` values twice on the `WoodworkTableAccessor`
      - If we do, make sure it gets included in the equality check (It’s not currently a part of the column accessor equality check - this is a problem we should fix no matter what!)
- How should we handle initializing Woodwork with an existing `ColumnSchema`?
  - Currently: passing each param to the `init` and deep copying the relevant attributes - we do this in four different places!!
  - Need a way of getting a deep copy of a `ColumnSchema`
  - Need a way of initializing with a schema?
- Add the other non-validation methods in `column_schema.py` as methods on `ColumnSchema` and update downstream
  - For tag updates this means passing the object in - this is where it might be useful to store the use_standard_tags value
- Move any other `WoodworkColumnAccessor` method and property logic to the `ColumnSchema` - doesn't seem like there's too much of this!

woodwork/column_accessor.py
```diff
@@ -196,10 +196,13 @@ def wrapper(*args, **kwargs):
 
                 # Try to initialize Woodwork with the existing schema
                 if _is_series(result):
-                    valid_dtype = _get_valid_dtype(type(result), self._schema['logical_type'])
+                    valid_dtype = _get_valid_dtype(type(result), self._schema.logical_type)
                     if str(result.dtype) == valid_dtype:
-                        schema = copy.deepcopy(self._schema)
-                        result.ww.init(**schema, use_standard_tags=self.use_standard_tags)
+                        result.ww.init(logical_type=self._schema.logical_type,
```
tamargrey, line 201: This is the first place where we could benefit from being able to do `new_series.ww.init(schema=series.ww.schema, use_standard_tags=series.ww.use_standard_tags)` like how we allow it in the table accessor. 

Since there's only four init params that we pass in, I might say we can get by without it for simplicity's sake, but we do this call four times by my count, and we also need to remember to copy the elements each time, and I think we'll be less likely to add a bug in the future if we are able to simplify this type of call.

woodwork/table_accessor.py
```diff
@@ -208,13 +215,11 @@ def __setitem__(self, col_name, column):
...
         elif self.use_standard_tags and not column.ww.use_standard_tags:
-            column.ww._schema['semantic_tags'] |= column.ww.logical_type.standard_tags
+            column.ww.add_semantic_tags(column.ww.logical_type.standard_tags)
             message = StandardTagsChangedWarning().get_warning_message(self.use_standard_tags, col_name)
             warnings.warn(message, StandardTagsChangedWarning)
         elif not self.use_standard_tags and column.ww.use_standard_tags:
-            column.ww._schema['semantic_tags'] -= column.ww.logical_type.standard_tags
-            message = StandardTagsChangedWarning().get_warning_message(self.use_standard_tags, col_name)
-            warnings.warn(message, StandardTagsChangedWarning)
+            column.ww.remove_semantic_tags(column.ww.logical_type.standard_tags)
```
tamargrey, line 222: The augmented assignment statements here and in the block above were causing issues now that we have semantic tags as an actual attribute instead of a dictionary element: `column.ww._schema.semantic_tags`. 

I figured we could make use of our WoodworkColumnAccessor operations here, and because `remove_semantic_tags` issues a `StandardTagsChangedWarning` when removing standard tags, it let us remove those lines here.

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -2138,7 +2138,7 @@ def test_setitem_overwrite_column(sample_df):
     df.ww.init(use_standard_tags=False)
 
     original_col = df['full_name']
-    new_series = pd.Series([0, 1, 2], dtype='float')
+    new_series = pd.Series([0, 1, 2], dtype='float', name='full_name')
```
thehomebrewnerd, line 2141: Is it necessary to name the series here for the error message to match? If so, I suspect a common occurrence would be for unnamed Series to be assigned. Does the error make sense in that case?

thehomebrewnerd (commented):
Looks good. Just a couple small comments.

woodwork/tests/accessor/test_table_accessor.py
tamargrey: Without a name it probably doesn't make sense--`Standard tags have been added to None`. I'm thinking maybe we make the name parameter optional in in `StandardTagsChangedWarning` and just say `Standard tags have been added to your column` if there's no name.

woodwork/column_schema.py
```diff
@@ -10,40 +10,53 @@
...
+                 logical_type=None,
+                 semantic_tags=None,
+                 use_standard_tags=False,
+                 description=None,
+                 metadata=None,
+                 validate=True):
+        """Create ColumnSchema
+
+        Args:
+            name (str, optional): The name of the column.
```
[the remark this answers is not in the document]
tamargrey: Yeah, I see your point here. I'd be fine removing it--won't be a breaking change, I think, because users didn't have access to `_get_column_dict` before, and we already handle the null name value in `_convert_input_to_set`. 

Plus, we will catch this error in the TableSchema because we convert the inputs to set before initializing individual `ColumnSchema`s so if someone had this issue in a table, they'd still get the good error with the column name. I see no problem in removing it!

thehomebrewnerd (approved):
Looks good!

PR #799, opened 2021-04-08
Add WoodworkColumnAccessor.schema and handle schema copying better

### LINKED ISSUE #792: Add WoodworkColumnAccessor.schema proprty
Now that the `WoodworkColumnAccessor` uses `SchemaColumn`, it will be useful to to have a `schema` property that can be referenced that returns a deep copied `SchemaColumn` object.

### LINKED ISSUE #793: Allow copying of `ColumnSchema`
Should return a copy of a ColumnSchema whose elements like semantic tags and metadata are deep copied as well. 

Useful for `WoodworkColumnAccessor.schema` object and initializing Woodwork with a schema

tamargrey:
- Adds a `schema` property to the Column accessor that uses `copy.deepcopy` because the Schema is simple enough to allow it (so I think we don't need a copy method)
- Avoids validating when re-initializing Woodwork after copying a Schema
- Closes #792 , closes #793

### DIFF
--- a/woodwork/column_accessor.py
+++ b/woodwork/column_accessor.py
@@ -63,6 +63,8 @@ class WoodworkColumnAccessor:
 
     @property
     def schema(self):
+        if self._schema is None:
+            _raise_init_error()
         return copy.deepcopy(self._schema)
 
     @property
--- a/woodwork/column_accessor.py
+++ b/woodwork/column_accessor.py
@@ -62,6 +62,10 @@ class WoodworkColumnAccessor:
                                     metadata=metadata)
 
     @property
+    def schema(self):
+        return copy.deepcopy(self._schema)
+
+    @property
     def description(self):
         """The description of the series"""
         if self._schema is None:
--- a/woodwork/table_accessor.py
+++ b/woodwork/table_accessor.py
@@ -600,7 +600,7 @@ class WoodworkTableAccessor:
                                       TypingInfoMismatchWarning)
                     else:
                         copied_schema = self._schema._get_subset_schema(list(self._dataframe.columns))
-                        result.ww.init(schema=copied_schema)
+                        result.ww.init(schema=copied_schema, validate=False)
                         result.ww.make_index = self.make_index
                 else:
                     # Confirm that the schema is still valid on original DataFrame
--- a/woodwork/table_accessor.py
+++ b/woodwork/table_accessor.py
@@ -190,7 +190,7 @@ class WoodworkTableAccessor:
             raise ColumnNotPresentError(key)
 
         series = self._dataframe[key]
-        column = self._schema.columns[key]
+        column = copy.deepcopy(self._schema.columns[key])
         column.semantic_tags -= {'index', 'time_index'}
 
         series.ww.init(logical_type=column.logical_type,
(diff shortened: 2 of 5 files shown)


woodwork/column_accessor.py
```diff
@@ -61,6 +61,10 @@ def init(self, logical_type=None, semantic_tags=None,
                                     description=description,
                                     metadata=metadata)
 
+    @property
+    def schema(self):
+        return copy.deepcopy(self._schema)
```
tamargrey, line 66: I _think_ `copy.deepcopy` does exactly what we need it to--copying for the purposes of the schema property and copying prior to initializing with a schema. So unless I'm missing something here, I'm happy to not implement a deep copy of our own and close the related issue

woodwork/table_accessor.py
```diff
@@ -307,7 +307,7 @@ def physical_types(self):
         """A dictionary containing physical types for each column"""
         if self._schema is None:
             _raise_init_error()
-        return {col_name: _get_valid_dtype(type(self._dataframe[col_name]), self.schema.logical_types[col_name]) for col_name in self._dataframe.columns}
+        return {col_name: _get_valid_dtype(type(self._dataframe[col_name]), self._schema.logical_types[col_name]) for col_name in self._dataframe.columns}
```
tamargrey, line 310: Here and one place above we were using `self.schema` instead of `self._schema`, which is preferable.

woodwork/table_accessor.py
```diff
@@ -629,7 +629,7 @@ def _get_subset_df_with_schema(self, cols_to_include, use_dataframe_order=True):
 
         new_schema = self._schema._get_subset_schema(cols_to_include)
         new_df = self._dataframe[cols_to_include]
-        new_df.ww.init(schema=new_schema)
+        new_df.ww.init(schema=new_schema, validate=False)
```
tamargrey, line 632: We shouldn't have to validate a subset schema

woodwork/table_accessor.py
```diff
@@ -600,7 +600,7 @@ def wrapper(*args, **kwargs):
                                       TypingInfoMismatchWarning)
                     else:
                         copied_schema = self._schema._get_subset_schema(list(self._dataframe.columns))
```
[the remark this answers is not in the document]
tamargrey: yes, but I think the idea is that we wanted to be explicit about making a copy here (though maybe the variable name does a good enough job of that)

woodwork/column_accessor.py
tamargrey: I'd be fine using `copy.deepcopy` for the table as well. That way we can use `_get_subset_schema` for getting actual subsets!

thehomebrewnerd: Yeah, that seems cleaner!

woodwork/table_accessor.py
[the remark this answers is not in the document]
thehomebrewnerd: No strong preference on my end, but I'm leaning toward `copied_schema = self.schema` as a slightly better option. As you say the variable name provides a clue that this is a copy, and the schema property code and docstring clearly say it returns a copy. That update would also mean we aren't using `_get_subset_schema` to get a full schema instead of a subset.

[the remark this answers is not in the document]
tamargrey: ok switched to using `self.schema`!

thehomebrewnerd (approved):
LGTM!

PR #801, opened 2021-04-08
Move spark config to a visible cell

tamargrey:
- Seeing the docs build fail on the `copy-schema` PR, so hoping to recreate that here and find a fix.
- The fix involves making the spark config visible, and this isn't ideal, but it seems like it's not getting executed in a raw rst cell in the same way it does in a regular cell.

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -26,6 +26,7 @@ Release Notes
         * Store column typing information in a ``ColumnSchema`` object instead of a dictionary (:pr:`791`)
     * Documentation Changes
         * Update Pygments version requirement (:pr:`751`)
+        * Move spark config to a visible cell (:pr:`801`)
     * Testing Changes
         * Add unit tests against minimum dependencies for python 3.6 on PRs and main (:pr:`743`, :pr:`753`, :pr:`763`)
         * Update spark config for test fixtures and docs build (:pr:`787`)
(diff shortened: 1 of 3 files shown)


docs/source/release_notes.rst
```diff
@@ -26,6 +26,7 @@ Release Notes
         * Store column typing information in a ``ColumnSchema`` object instead of a dictionary (:pr:`791`)
     * Documentation Changes
         * Update Pygments version requirement (:pr:`751`)
+        * Move spark config to a visible cell (:pr:`801`)
```
thehomebrewnerd, line 29: What if we updated this line to say `Update spark config for docs build` and mention this pr and 787 from below, and then update the testing changes entry to just say `Update spark config for test fixtures` to separate these out better?

thehomebrewnerd (commented):
Do you think we should add a comment somewhere to let readers know that this spark config cell is likely not something they will need to execute? I'm not sure how to best convey that, but I think it would be nice to somehow let them know this isn't something that is required for them to use if they already have spark configured properly.

tamargrey:
> Do you think we should add a comment somewhere to let readers know that this spark config cell is likely not something they will need to execute? I'm not sure how to best convey that, but I think it would be nice to somehow let them know this isn't something that is required for them to use if they already have spark configured properly.

Maybe a comment above indicating that and also provide a link to some sort of resource that explains what it is so that if they did happen to need to run it they can figure that out for themselves.

docs/source/release_notes.rst
tamargrey: that makes sense to me!

thehomebrewnerd (approved):
LGTM!

PR #807, opened 2021-04-09
Add config to switch between nullable and non-nullable dtypes

### LINKED ISSUE #802: Add config setting to allow users to switch between nullable and non-nullable dtypes
Several libraries are not yet fully supporting the new nullable pandas dtypes. Since Woodwork uses these new types by default, this presents problems for users who wish to use Woodwork in conjunction with libraries that do not yet support these types. Users could benefit from having an easy way to configure Woodwork to switch between using the nullable and non-nullable dtypes.

This could be done via a global configuration option such as this:
```python
# Update Woodwork to use nullable dtypes
ww.config.set_option('use_nullable_dtypes', True)

# Update Woodwork to use non-nullable dtypes
ww.config.set_option('use_nullable_dtypes', False)
```

This would allow users to switch between dtypes using an approach similar to this:
```python
import pandas as pd
import woodwork as ww
from woodwork.logical_types import Boolean

df = pd.DataFrame({'id': [0, 1, 2], 'bool_col': [True, False, True]})

# Use non-nullable dtypes
ww.config.set_option('use_nullable_dtypes', False)

df.ww.init(logical_types={'bool_col': 'Boolean'})

assert Boolean.primary_dtype == 'bool'
assert df.ww['bool_col'].logical_type == Boolean
assert df['bool_col'].dtype == 'bool'

# Update Woodwork to use nullable dtypes
ww.config.set_option('use_nullable_dtypes', True)

df.ww # Accessing schema will result in invalid schema warning after config change

# Reinitialize to fix invalid schema
df.ww.init(logical_types={'bool_col': 'Boolean'})) 

assert Boolean.primary_dtype == 'boolean'
assert df.ww['bool_col'].logical_type == Boolean
assert df['bool_col'].dtype == 'boolean'
```

thehomebrewnerd:
- Add config to switch between nullable and non-nullable dtypes
- Closes #802 

This PR adds a global configuration setting `use_nullalbe_dtypes` that allows users to easily switch between using the newer nullable pandas dtypes and the corresponding older non-nullable dtypes.

To Do:
- [x] Update default setting for `use_nullable_types` to False
- [x] Fix failing tests after changing default
- [x] Add new tests including null values setting `use_nullable_types` to True
- [x] Fix documentation build failures
- [ ] Review and update all documentation for accurate type information
- [x] Merge in latest changes from main


woodwork/logical_types.py
```diff
@@ -30,6 +51,14 @@ def __eq__(self, other, deep=False):
     def __str__(self):
         return str(self.__class__)
 
+    @property
+    def primary_dtype(self):
+        return type(self).primary_dtype
```
tamargrey, line 56: I'm not sure that it's immediately obvious why we need to do this (unless you're more familiar with python meta classes than I originally was!). Maybe it's just as simple as adding a couple of comments here or at the property definition on the metaclass indicating that this lets us include logic that gets executed whether or not it's being referenced from an instance object or just the class object itself.

When I was reading up on this stuff, I found this example really helpful:

```python
class MetaClass(type):
    @property
    def meta_prop(cls):
        return '(MetaClass.meta_prop) ' + cls.class_attr

class Class(object, metaclass=MetaClass):
    class_attr = ':D'
    
    @property
    def class_prop(self):
        return ('(Class.class_prop) ' + self.class_attr)
    
    @property
    def meta_prop(self):
        # For when we want to perform some operation beyond its definition at creation
        return ('(Class.meta_prop) ' + type(self).meta_prop)

# Access class attr
assert Class().class_attr == ':D'
assert Class.class_attr == ':D'

# Access class property
assert Class().class_prop == '(Class.class_prop) :D'
assert isinstance(Class.class_prop, property) # looks like '<property object at 0x16617f6d0>'

#  Access meta property
# This is the goal! Being able to reference a property on an uninstantiated class object
assert Class().meta_prop == '(Class.meta_prop) (MetaClass.meta_prop) :D'
assert Class.meta_prop == '(MetaClass.meta_prop) :D'
```

freddyaboulton:
@thehomebrewnerd This is great! Thanks for putting this together. I looked for three things when integrating this into evalml:

1. mutual information works for non-nullable bool (this wasn't the case when I defined my own type system)
2. ww raises an error when user passes data created with nullable types
3. Any unintended changes in evalml behavior. 

Our data checks unit tests passed without modifying anything so I think we're good on 1!

I manually tested 2 and we get an error! Maybe we can make the error message more actionable in the future but overall this is great. Repro:

image

For point 3. I didn't find unexpected changes in behavior compared to what I had listed in [points 4-6](https://alteryx.quip.com/JxZNAddlnAbb/Using-the-Woodwork-Accessor-in-EvalML#RCZACA1DgJz) of the original integration notes but two weird things came up though. 

If a user creates a column with [True, False, None] and manually types it as `category`, ww will infer as a boolean. I expected woodwork to infer as category given the limitations of the non-nullable type:

Repro:
```python
import pandas as pd
import woodwork as ww

df = pd.DataFrame({"should_be_categorical": pd.Series([True, None, False, None, True], dtype='category')})
df.ww.init()
assert df.ww.logical_types["should_be_categorical"] != ww.logical_types.Categorical
```

The other weird thing is that users can create a column with [True, False] and type it as object and woodwork will not infer it as boolean. 

Repro:
```python
 import pandas as pd
import woodwork as ww

df = pd.DataFrame({"should_be_boolean": pd.Series([True, True, False, True, True], dtype=object)})
df.ww.init()
assert df.ww.logical_types["should_be_boolean"] != ww.logical_types.Boolean
```

I don't think these two points are problems for evalml now, but they might be in the future. Or they might be problems for other users so I want to discuss what the expected behavior should be in these cases!

FYI @dsherry

thehomebrewnerd:
@freddyaboulton Thanks for testing this out and providing your feedback. Here is a little more info that explains the behavior you are seeing with the Boolean inference.

Our boolean inference function is currently very simple, and we use the pandas function `pdtypes.is_bool_dtype` to determine if a series is boolean or not.  This is our inference function.

```python
def boolean_func(series):
    if pdtypes.is_bool_dtype(series.dtype):
        return True
    return False
```

In your first case, what I found interesting is that `pdtypes.is_bool_dtype` returns true for a category dtype, which I did not expect.

```python
import pandas as pd
import pandas.api.types as pdtypes
series = pd.Series([True, None, False, None, True], dtype='category')
assert pdtypes.is_bool_dtype(series.dtype)
assert series.dtype == 'category'
```

I think we can fix this by simply returning False from our boolean inference function if the dtype is a categorical dtype.

As for the second case where a series with a dtype of `object` is not inferred as boolean, this is also consistent with our simple inference function. Certainly room for improvement, but as of now the only columns that will get inferred as boolean are those that have a boolean dtype according to `pdtypes.is_bool_dtype`.

woodwork/logical_types.py
thehomebrewnerd: @tamargrey Added a few comments to the property definitions. Do you think this is sufficient or is more needed?

tamargrey: Yeah, I think they do a good job of clarifying what's going on here!

thehomebrewnerd:
Closing this in favor of PR #830 which fixes the problems with a less complicated implementation.

PR #809, opened 2021-04-12
Store use_standard_tags only on ColumnSchema

### LINKED ISSUE #794: Determine if `use_standard_tags` should live on the ColumnSchema or WoodworkColumnAccessor
It's currently on the WoodworkColumnAccessor, but if we want to follow the TableSchema does, it would live there.

Reason to store on ColumnAccessor:
- The TableSchema contains ColumSchemas, so we'd end up storing `use_standard_tags` in multiple places on a TableAccessor (note - this value doesn't change, so it's unlikely to get out of sync)

Reasons to store on ColumnSchema:
- For the Table schema/accessor, `use_standard_tags` lives on the schema
- A reason to store the value on the Schema is that if a user were to perform some operation on a ColumnSchema (say resetting tags), the tags should be able to follow `use_standard_tags`. 
- When copying a schema, the value of `use_standard_tags` should be able to be retained

Either way, `use_standard_tags` is not getting included in the equality check, so if we move it to the ColumnSchema, it should get added to its equality check. Otherwise, we should add it to the WoodworkColumnAccessor equality check.

tamargrey:
- This PR stops storing `use_standard_tags` on the `TableSchema` and stores it only on `ColumnSchema`. This causes a few changes:
  - Different columns in Woodwork can have different use_standard_tags values within the same table
  - At init, use_standard_tags can either be a bool or a dict[str -> bool] . Any columns not included in the dictionary will default to the default use_standard_tags value
  - If a user is setting a new series without Woodwork initialized with df.ww['new_col'] = pd.Series(), the use_standard_tags value will be the default `True` instead of following the wider Table's setting
  - Serialization has changed
- Closes #794

### DIFF
--- a/woodwork/column_accessor.py
+++ b/woodwork/column_accessor.py
@@ -53,11 +53,10 @@ class WoodworkColumnAccessor:
         logical_type = _get_column_logical_type(self._series, logical_type, self._series.name)
 
         self._validate_logical_type(logical_type)
-        self.use_standard_tags = use_standard_tags
 
         self._schema = ColumnSchema(logical_type=logical_type,
                                     semantic_tags=semantic_tags,
-                                    use_standard_tags=self.use_standard_tags,
+                                    use_standard_tags=use_standard_tags,
                                     description=description,
                                     metadata=metadata)
 
@@ -165,6 +164,8 @@ class WoodworkColumnAccessor:
     def __eq__(self, other):
         if self._schema != other._schema:
             return False
+        if self._series.name != other._series.name:
+            return False
         if isinstance(self._series, pd.Series):
             return self._series.equals(other._series)
         return True
@@ -207,7 +208,7 @@ class WoodworkColumnAccessor:
                                        semantic_tags=copy.deepcopy(self._schema.semantic_tags),
                                        description=self._schema.description,
                                        metadata=copy.deepcopy(self._schema.metadata),
-                                       use_standard_tags=self.use_standard_tags)
+                                       use_standard_tags=self._schema.use_standard_tags)
                     else:
                         invalid_schema_message = 'dtype mismatch between original dtype, ' \
                             f'{valid_dtype}, and returned dtype, {result.dtype}'
@@ -263,7 +264,7 @@ class WoodworkColumnAccessor:
                                                            self.semantic_tags,
                                                            self._series.name,
                                                            self.logical_type.standard_tags,
-                                                           self.use_standard_tags)
+                                                           self._schema.use_standard_tags)
 
     def reset_semantic_tags(self):
         """Reset the semantic tags to the default values. The default values
@@ -276,7 +277,7 @@ class WoodworkColumnAccessor:
         if self._schema is None:
             _raise_init_error()
         self._schema.semantic_tags = _reset_semantic_tags(self.logical_type.standard_tags,
-                                                          self.use_standard_tags)
+                                                          self._schema.use_standard_tags)
 
     def set_logical_type(self, logical_type):
         """Update the logical type for the series, clearing any previously set semantic tags,
@@ -297,7 +298,7 @@ class WoodworkColumnAccessor:
         return init_series(new_series,
                            logical_type=logical_type,
                            semantic_tags=None,
-                           use_standard_tags=self.use_standard_tags,
+                           use_standard_tags=self._schema.use_standard_tags,
                            description=self.description,
                            metadata=copy.deepcopy(self.metadata))
 
@@ -313,7 +314,7 @@ class WoodworkColumnAccessor:
             _raise_init_error()
         self._schema.semantic_tags = _set_semantic_tags(semantic_tags,
                                                         self.logical_type.standard_tags,
-                                                        self.use_standard_tags)
+                                                        self._schema.use_standard_tags)
 
 
 def _raise_init_error():
--- a/woodwork/column_accessor.py
+++ b/woodwork/column_accessor.py
@@ -161,6 +161,12 @@ class WoodworkColumnAccessor:
             _raise_init_error()
         return self._schema.semantic_tags
 
+    @property
+    def use_standard_tags(self):
+        if self._schema is None:
+            _raise_init_error()
+        return self._schema.use_standard_tags
+
     def __eq__(self, other):
         if self._schema != other._schema:
             return False
@@ -172,12 +178,9 @@ class WoodworkColumnAccessor:
 
     def __getattr__(self, attr):
         # If the method is present on the Accessor, uses that method.
-        # If the method is present on TableSchema, uses that method.
         # If the method is present on Series, uses that method.
         if self._schema is None:
             _raise_init_error()
-        if hasattr(self._schema, attr):
-            return self._make_schema_call(attr)
         if hasattr(self._series, attr):
             return self._make_series_call(attr)
         else:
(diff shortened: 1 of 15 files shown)


woodwork/column_accessor.py
```diff
@@ -165,15 +164,20 @@ def semantic_tags(self):
...
         return True
 
     def __getattr__(self, attr):
         # If the method is present on the Accessor, uses that method.
+        # If the method is present on TableSchema, uses that method.
         # If the method is present on Series, uses that method.
         if self._schema is None:
             _raise_init_error()
+        if hasattr(self._schema, attr):
+            return self._make_schema_call(attr)
```
tamargrey, line 180: We didn't previously have a need for this, but unless we make `use_standard_tags` a property on the `WoodworkColumnAccessor`, we will still need a `_make_schema_call` method.

woodwork/column_accessor.py
```diff
@@ -165,15 +164,20 @@ def semantic_tags(self):
     def __eq__(self, other):
         if self._schema != other._schema:
             return False
+        if self._series.name != other._series.name:
+            return False
```
tamargrey, line 168: Our test for this was comparing a datetime series to a categorical series, and wasn't actually checking whether the names being the only difference would cause two column's to not be equal, and it turns out it wasn't (I guess `series.equals(other_series)` doesn't catch different names?)

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -2044,7 +2050,8 @@ def test_setitem_new_column(sample_df):
     assert 'test_col2' in df.columns
     assert 'test_col2' in df.ww._schema.columns.keys()
     assert df.ww['test_col2'].ww.logical_type == Integer
-    assert df.ww['test_col2'].ww.semantic_tags == set()
+    assert df.ww['test_col2'].ww.use_standard_tags is True
+    assert df.ww['test_col2'].ww.semantic_tags == {'numeric'}
```
tamargrey, line 2054: Note the change in behavior here:
The dataframe was initialized with `use_standard_tags=Falst`, but when we use `setitem` with a pandas series, we have to init Woodwork, so we default to `use_standard_tags=True` since we have no overall table value.

Users will have to init woodwork on the series to get `use_standard_tags` to be something other than the default.

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -2123,14 +2132,13 @@ def test_setitem_overwrite_column(sample_df):
...
-    with pytest.warns(StandardTagsChangedWarning, match=match):
-        df.ww['full_name'] = new_series
+    df.ww['full_name'] = new_series
 
     assert 'full_name' in df.columns
     assert 'full_name' in df.ww._schema.columns.keys()
     assert df.ww['full_name'].ww.logical_type == Double
-    assert df.ww['full_name'].ww.semantic_tags == {'test_tag', 'numeric'}
+    assert df.ww['full_name'].ww.use_standard_tags is False
+    assert df.ww['full_name'].ww.semantic_tags == {'test_tag'}
```
tamargrey, line 2141: Note the change in behavior here:

The dataframe was initialized with `use_standard_tags=True`, but when we use `setitem` with a Woodwork column where we set `use_standard_tags=False`, we're no longer updating the column to match the rest of the table because each column is independent, so the semantic tags don't change.

test-requirements.txt
```diff
@@ -5,4 +5,5 @@ pytest-xdist==2.1.0
 boto3>=1.10.45
 moto[all]>=1.3.14
 smart-open>=5.0.0
-pyarrow>=2.0.0
\ No newline at end of file
+pyarrow>=2.0.0
+mock==4.0.3
```
tamargrey, line 9: Not sure why this didn't come up sooner, but without adding `mock` to the test requirements we started getting `ModuleNotFoundError No module named mock`

woodwork/column_accessor.py
tamargrey: there's a code coverage issue below because we don't have any callable attrs... I'd rather not update the `make_schema_call` to not have the callable case, so I'm now thinking it's better to remove this and add a property on the column accessor instead... @thehomebrewnerd

woodwork/table_accessor.py
```diff
@@ -215,13 +224,7 @@ def __setitem__(self, col_name, column):
...
 
         if column.ww._schema is None:
-            column = init_series(column, use_standard_tags=self.use_standard_tags)
-        elif self.use_standard_tags and not column.ww.use_standard_tags:
-            column.ww.add_semantic_tags(column.ww.logical_type.standard_tags)
-            message = StandardTagsChangedWarning().get_warning_message(self.use_standard_tags, col_name)
-            warnings.warn(message, StandardTagsChangedWarning)
-        elif not self.use_standard_tags and column.ww.use_standard_tags:
-            column.ww.remove_semantic_tags(column.ww.logical_type.standard_tags)
+            column = init_series(column, use_standard_tags=True)
```
thehomebrewnerd, line 227: This is much cleaner! Nice!

woodwork/table_accessor.py
```diff
@@ -347,6 +350,13 @@ def time_index(self):
             _raise_init_error()
         return self._schema.time_index
 
+    @property
+    def use_standard_tags(self):
+        """Whether standard tags are used for each column in the table"""
```
thehomebrewnerd, line 355: When I first read this my thought was that this was returning a boolean indicating whether or not all columns had the same value. What if we said this instead? "A dictionary containing the `use_standard_tags` setting for each column in the table"

woodwork/tests/accessor/test_table_accessor.py
thehomebrewnerd: I could potentially see this as a point of confusion, but I think this behavior makes sense. If users want to control what Woodwork does, they should initialize WW first and specify what they want, otherwise they get the default WW behavior.

woodwork/tests/schema/test_table_schema_init.py
```diff
@@ -397,3 +413,49 @@ def test_validation_methods_called(mock_validate_params, mock_check_index,
...
+
+    no_standard_tags_schema = TableSchema(sample_column_names, sample_inferred_logical_types,
+                                          use_standard_tags=False)
+    assert set(no_standard_tags_schema.columns.keys()) == set(no_standard_tags_schema.use_standard_tags.keys())
+    assert not any([*no_standard_tags_schema.use_standard_tags.values()])
+    assert no_standard_tags_schema.semantic_tags['id'] == set()
+
+
+def test_use_standard_tags_from_dict(sample_column_names, sample_inferred_logical_types):
+    default_schema = TableSchema(sample_column_names, sample_inferred_logical_types, use_standard_tags={col_name: False for col_name in sample_column_names})
```
thehomebrewnerd, line 433: Line is a little long, maybe break up.

thehomebrewnerd (commented):
Looking good, just a few relatively minor comments.

woodwork/table_accessor.py
tamargrey: Good point! Changing

woodwork/column_accessor.py
tamargrey: Added `use_standard_tags` as a property on the Column Accessor

woodwork/column_accessor.py
```diff
@@ -165,15 +164,20 @@ def semantic_tags(self):
...
         if isinstance(self._series, pd.Series):
             return self._series.equals(other._series)
         return True
 
     def __getattr__(self, attr):
         # If the method is present on the Accessor, uses that method.
+        # If the method is present on TableSchema, uses that method.
```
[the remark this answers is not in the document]
tamargrey: ended up removing this line now that we don't have `_make_schema_call`

thehomebrewnerd (approved):
LGTM!

PR #810, opened 2021-04-12
Fix spelling in Dask/Koalas Guide

gsheni:
- Fixes Thew --> The


tamargrey:
The ModuleNotFoundError just started showing up for me as well. We can fix by adding `mock==4.0.3` (or another version if it'd be better) to our test requirements (not sure why it wasn't relevant before, though) to fix it

tamargrey (approved):

PR #814, opened 2021-04-13
Allow initializing a WoodworkColumnAccessor with a ColumnSchema

### LINKED ISSUE #795: Allow initializing Woodwork on a series with a ColumnSchema
Now that we have a schema object, we can't rely on `**dict` to pass the parameters to `WoodworkColumnAccessor.init`. Right now we're explicitly passing all the parameters in after copying them in four different places:

- `ColumnAccessor._make_series_call`
- `indexers._process_selection`
- `TableAccessor.__getitem__`
- `TableAccessor.pop`

This isn't ideal and could be cleaned up by allowing initializing with a ColumnSchema and passing a copied ColumnSchema.

Note: To add this, we might want to pull the "valid schema" logic into a helper function and make sure it's well-defined.

tamargrey:
- This PR allows `series.ww.init(schema=schema)` and updates relevant locations in Woodwork where this is useful
- Closes #795

### DIFF
--- a/woodwork/column_accessor.py
+++ b/woodwork/column_accessor.py
@@ -32,7 +32,8 @@ class WoodworkColumnAccessor:
         self._schema = None
 
     def init(self, logical_type=None, semantic_tags=None,
-             use_standard_tags=True, description=None, metadata=None):
+             use_standard_tags=True, description=None, metadata=None,
+             schema=None):
         """Initializes Woodwork typing information for a Series.
 
         Args:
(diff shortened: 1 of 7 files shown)


woodwork/table_accessor.py
```diff
@@ -111,6 +111,8 @@ def init(self,
                 extra_params.append('logical_types')
             if already_sorted:
                 extra_params.append('already_sorted')
+            if not use_standard_tags or isinstance(use_standard_tags, dict):
+                extra_params.append('use_standard_tags')
```
tamargrey, line 115: now that `use_standard_tags` is no longer a part of the kwargs for the TableAccessor (bc it's default is different than the TableSchema's), we need to add it to this check explicitly

woodwork/indexers.py
```diff
@@ -42,16 +42,14 @@ def _process_selection(selection, original_data):
         elif _is_dataframe(original_data):
             # Selecting a single column from a DataFrame
             schema = original_data.ww.schema.columns[selection.name]
-            schema.semantic_tags = schema.semantic_tags - {'index'} - {'time_index'}
+            schema.semantic_tags = (schema.semantic_tags - {'index'} - {'time_index'})
+            if schema.use_standard_tags:
+                schema.semantic_tags |= schema.logical_type.standard_tags
```
thehomebrewnerd, line 47: Don't think I'm following this. Wouldn't the schema for the column already contain the standard tags?

woodwork/table_accessor.py
```diff
@@ -203,12 +205,10 @@ def __getitem__(self, key):
         series = self._dataframe[key]
         column = copy.deepcopy(self._schema.columns[key])
         column.semantic_tags -= {'index', 'time_index'}
+        if column.use_standard_tags:
+            column.semantic_tags |= column.logical_type.standard_tags
```
thehomebrewnerd, line 209: Same question as above.

thehomebrewnerd (commented):
Looks pretty good - just one minor comment and a couple questions.

woodwork/indexers.py
tamargrey: Not if it was previously an index column!

We make a point on line 45 to remove index tags, so if we don't manually add the standard tags back in, we lose them. Previously, we were going through the Column Accessor init, so the standard tags were getting added back in through that and we lose that when we init with a ColumnSchema

thehomebrewnerd: Got it! Thanks for clarifying!

woodwork/column_accessor.py
```diff
@@ -49,21 +53,43 @@ def init(self, logical_type=None, semantic_tags=None,
                 based on the inferred or specified logical type of the series. Defaults to True.
             description (str, optional): Optional text describing the contents of the series.
             metadata (dict[str -> json serializable], optional): Metadata associated with the series.
+            schema (Woodwork.TableSchema, optional): Typing information to use for the Series instead of performing inference.
```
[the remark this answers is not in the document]
tamargrey: whoopse--fixed!

thehomebrewnerd (approved):
Looks good!

PR #817, opened 2021-04-14
Add ColumnSchema repr

### LINKED ISSUE #798: Add ColumnSchema repr
Currently `series.ww._schema`'s repr is `<woodwork.column_schema.ColumnSchema at 0x16591c0a0>`, which is not ideal.

The Column Accessor has a repr, and the schema repr should be similar, but it won't contain a name or physcial type:

```python
<Column: (Logical Type = Integer) (Semantic Tags = {'numeric'})>
```

tamargrey:
- The repr only show logical type and semantic tags, and if either is not present, doesn't include it in the repr (treats empty semantic tags list as not present, which is good for Featuretools' use but not for the ColumnAccessor, so the Column Accessor will not inherit this behavior)
- Closes #798 

Note: This repr will be used to look at a Featuretools primitive's input and return types, so it's important that this is functional for that purpose.

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -13,6 +13,7 @@ Release Notes
         * Add validation control to ``deserialize.read_woodwork_table`` and ``ww.read_csv`` (:pr:`788`)
         * Add ``WoodworkColumnAccessor.schema`` and handle copying column schema (:pr:`799`)
         * Allow initializing a ``WoodworkColumnAccessor`` with a ``ColumnSchema`` (:pr:`814`)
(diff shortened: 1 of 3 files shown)


thehomebrewnerd (approved):
Looks good to me!

rwedge:
Unrelated, the ColumnSchema docstring references a "name" parameter which doesn't exist

jeff-hernandez (approved):

tamargrey:
> Unrelated, the ColumnSchema docstring references a "name" parameter which doesn't exist

Removed--thanks for pointing out @rwedge!

gsheni (approved):

PR #825, opened 2021-04-14
Use virtualenv with latest dep checker, remove latest text files

gsheni:
- The latest dependency checker is create MRs and then closing them with no changes.
- This might be because the latest dependencies text files were not generated from machineFL.
- This also could be because there is a race condition.
- This MR attempts to fix this by deleting the files, and the latest dependencies checker will re-generate them. It also uses virtualenv and has removes the matrix configuration.

### DIFF
--- a/.github/workflows/latest_dependency_check.yml
+++ b/.github/workflows/latest_dependency_check.yml
@@ -14,7 +14,7 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: '3.8.x'
-    - name: Install woodwork dependencies and test dependencies
+    - name: Install pip and virtualenv
       run: |
         python -m pip install --upgrade pip
         python -m pip install virtualenv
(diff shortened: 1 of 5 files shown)


.github/workflows/latest_dependency_check.yml
```diff
@@ -8,34 +8,39 @@ on:
...
         python-version: '3.8.x'
-    - name: Install woodwork dependencies and test dependencies
+    - name: Install pip and virtualenv
       run: |
         python -m pip install --upgrade pip
+        python -m pip install virtualenv
+    - name: Update latest core dependencies
+      run: |
+        python -m virtualenv venv_core
+        source venv_core/bin/activate
```
rwedge, line 24: I'd add a `pip install --upgrade pip` after activating each virtual env, to ensure the pip inside the virtual env is also up to date

jeff-hernandez (approved):

.github/workflows/latest_dependency_check.yml
gsheni: Fixed:
https://github.com/alteryx/woodwork/pull/825/commits/2aad95d21414e4413118e7eebc354c8a9e1ad0bf

rwedge (approved):
LGTM

PR #829, opened 2021-04-15
Move column_schema.py functions to be ColumnSchema methods

### LINKED ISSUE #796: Move functions in `column_schema.py` to be methods on `ColumnSchema`
In order to simplify the integration of `ColumnSchema` the functions that were previously helper functions to the column dict were left as functions.

But as some of them take in a column object (`_is_col_numeric` for example) or are relevant to a specific column object (`_get_column_tags`, `_reset_semantic_tags`), we should make them `ColumnSchema` methods. 

Every function except for the validation functions should live on `ColumnSchema`, though we should confirm that this will not break any of the semantic tag functions, which get used by both the table and column accessors.

We should consider making the semantic tag functions into "public" methods like they are on the TableSchema.

tamargrey:
- These functions should live on the ColumnSchema object,  but wanted to wait to make the change until other things were dealt with
- Closes #796

### DIFF
--- a/woodwork/column_schema.py
+++ b/woodwork/column_schema.py
@@ -130,6 +130,20 @@ class ColumnSchema(object):
                           StandardTagsChangedWarning)
         self.semantic_tags = self.semantic_tags.difference(tags_to_remove)
 
+    def _reset_semantic_tags(self):
+        """Reset the set of semantic tags to the default values. The default values
+        will be either an empty set or the standard tags, controlled by the
+        use_standard_tags boolean.
+
+        Args:
+            standard_tags (set): Set of standard tags for the column logical type
+            use_standard_tags (bool): If True, retain standard tags after reset
+        """
+        new_tags = set()
+        if self.use_standard_tags:
+            new_tags = set(self.logical_type.standard_tags)
+        self.semantic_tags = new_tags
+
 
 def _validate_logical_type(logical_type):
     ltype_class = _get_ltype_class(logical_type)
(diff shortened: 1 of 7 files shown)


woodwork/column_schema.py
```diff
@@ -69,118 +69,100 @@ def __repr__(self):
...
+        if self.use_standard_tags:
+            if self.logical_type is None:
+                raise ValueError("Cannot use standard tags when logical_type is None")
+            semantic_tags = semantic_tags.union(self.logical_type.standard_tags)
 
-def _is_col_numeric(col):
-    return 'numeric' in col.logical_type.standard_tags
+        return semantic_tags
 
+    def _is_numeric(self):
```
thehomebrewnerd, line 83: Should these `_is_*` methods be properties instead since they never take arguments?

Also, maybe this is a separate issue/PR, but should these be public?

thehomebrewnerd (approved):
One small comment/question, but otherwise looks good.

woodwork/column_schema.py
tamargrey: I like the idea of them being properties! 

I've also been considering making them public. Should be pretty quick, so I'm fine adding that in here (I think EvalML requested it at some point?)

tamargrey: @thehomebrewnerd made both those changes--can revert/change it if you're not a fan

thehomebrewnerd (approved):
I like these changes - cleans things up just a little bit.

Issue #153, opened 2020-09-25
Improve inference of booleans to handle string representations


gsheni:
- To improve the boolean inference, we can say that if rows value counts fall into:
    - `[1, True, "true", "True", "yes", "t", "T"]`
    - `[0, False, "false", "False", "no", "f", "F"]`
  - --> we should infer the Logical Type to be Boolean. 
- This would prevent the following weird inference, where both columns are inferred to be categorical.

thehomebrewnerd:
If we update the inference of booleans to identify series such as `[1, 0, 1, 1]` a `Boolean` logical type, this series will also be a match for the `Integer` logical type. Both of these inference functions will call ` series.isnull().any()`  separately, which could be inefficient for large datasets. As part of the implementation, it would be good to update so this call only needs to happen once, if possible.

See PR #830 for additional context.

PR #830, opened 2021-04-15
Add BooleanNullable and IntegerNullable logical types

### LINKED ISSUE #802: Add config setting to allow users to switch between nullable and non-nullable dtypes

thehomebrewnerd:
- Add BooleanNullable and IntegerNullable logical types
- Closes #802 

This PR is submitted as an alternative to PR #807. In this PR the `Boolean` logical type is updated to use the `bool` dtype and the `Integer` logical type uses the `int64` dtype. Two new nullable dtypes were added `BooleanNullable` and `IntegerNullable` using the `boolean` and `Int64` dtypes, respectively.

Type inference was updated so that the nullable logical types will be inferred only if the data contains missing values.


woodwork/type_sys/inference_functions.py
```diff
@@ -29,6 +29,12 @@ def categorical_func(series):
 
 
 def integer_func(series):
+    if integer_nullable_func(series) and series.notnull().all():
```
gsheni, line 32: Can we verify that `notnull().all()` is the fastest method we can use? We can use something like this:
- https://stackoverflow.com/a/29530601/2512385

woodwork/type_sys/inference_functions.py
```diff
@@ -45,7 +51,13 @@ def double_func(series):
 
 
 def boolean_func(series):
-    if pdtypes.is_bool_dtype(series.dtype):
+    if boolean_nullable_func(series) and series.notnull().all():
```
gsheni, line 54: Can we verify that this fast? I think it would be faster to say that if there is Any NaN value in the Boolean Column, then we use **BooleanNullable**. Otherwise use **Boolean** (Rather than checking all values are NaN)

Maybe we check performance with something like this: 
- https://stackoverflow.com/a/29530601/2512385

thehomebrewnerd: How about `if boolean_nullable_func(series) and not series.isnull().any()` here instead?

gsheni: Yah that works

woodwork/type_sys/inference_functions.py
```diff
@@ -51,7 +51,7 @@ def double_func(series):
 
 
 def boolean_func(series):
-    if boolean_nullable_func(series) and series.notnull().all():
+    if boolean_nullable_func(series) and not series.isnull().any():
```
tamargrey, line 54: We'll end up doing `not series.isnull().any()` twice for this and the integer inference function. Would it be worth calculating it once in `infer_logical_type` and passing it in as a boolean? (I guess we'd maybe have to deal with kwargs or something in order to not have to change the rest of the inference functions to take it in as well).

Plus there might be a way to combine it with the `dropped_na = col.dropna()` that will also happen in the inference functions called in `infer_logical_type`.

gsheni: Agreed. Less repeated calls, the better our inference will be.

thehomebrewnerd: I think this will only get called once? A series can't be both `BooleanNullable` and `IntegerNullable`, based on the current inference functions at least, so either `boolean_nullable_func()` or `integer_nullable_func()` should return False and short circuit the `not series.isnull().any()` call.

thehomebrewnerd: My explanation above might miss the mark a bit. `boolean_func` will generally only get called if `boolean_nullable_func` returns `True`. This could change is someone modifies the parent-child relationship, making `Boolean` a root-level logical type instead a child of `BooleanNullable`, but that's probably unlikely. Same story for the integer functions.

So since `boolean_nullable_func` and `integer_nullable_func` can't both return `True` for a single series, I think this means that only one of `boolean_func` and `integer_func` will get called, meaning this `not series.isnull().any()` call will only happen once.

I think...

tamargrey: Got it! That makes sense to me.

Maybe we just make a note of this somewhere so that if we decide to update inference so if `[1,0,1,1]` could be a valid boolean column we're adding this change

thehomebrewnerd: Yeah, if we go with this approach and merge in this PR we can update #153 to include some info about this.

freddyaboulton:
I think this looks great from evalml's perspective! I tested it out and it's the same behavior as #807. The only thing that really changed is that I had to add some logic in evalml for raising an error when nullable types are passed in rather than relying on `ww.init` to raise an error due to different config settings. That's ok with me!

tamargrey:
Should we be updating `ColumnSchema.is_boolean` to include both `Boolean` and `BooleanNullable` dtypes and `get_valid_mi_types` to do so as well? That way our statistics functions will include both ltypes as well.

thehomebrewnerd:
> Should we be updating `ColumnSchema.is_boolean` to include both `Boolean` and `BooleanNullable` dtypes and `get_valid_mi_types` to do so as well? That way our statistics functions will include both ltypes as well.

Yes - I made that change and also updated the tests for both `is_booelan` and `is_numeric` to check the nullable types give the expected result.

woodwork/tests/accessor/test_statistics.py
```diff
@@ -271,15 +273,12 @@ def test_describe_accessor_method(describe_df):
         assert isinstance(stats_df, pd.DataFrame)
         assert set(stats_df.columns) == {'category_col'}
         assert stats_df.index.tolist() == expected_index
-        expected_vals.equals(stats_df['category_col'].dropna())
+        assert expected_vals.equals(stats_df['category_col'].dropna())
```
thehomebrewnerd, line 276: Note - these lines were missing the `assert` so were never being evaluated in our tests before.

woodwork/tests/accessor/test_statistics.py
```diff
@@ -390,14 +389,36 @@ def test_describe_accessor_method(describe_df):
...
+        assert stats_df.index.tolist() == expected_index
+        assert expected_vals.equals(stats_df['numeric_col'].dropna())
 
     # Test natural language columns
     natural_language_data = describe_df[['natural_language_col']]
-    if ks and isinstance(category_data, ks.DataFrame):
-        expected_dtype = 'object'
-    else:
-        expected_dtype = 'string'
+    expected_dtype = 'string'
```
thehomebrewnerd, line 421: Now that Koalas supports `string` we can update this.

woodwork/tests/type_system/test_ltype_inference.py
```diff
@@ -2,32 +2,30 @@
...
     'uint64',
     'uintp',
-    'float32',  # compatible starting with 1.4.0
     'float_',
     'object',
-    'datetime64[ns]',  # compatible starting with 1.4.0
     'category',
-    'string',
-    'boolean',
 ]
```
thehomebrewnerd, line 29: This never got updated when Koalas added support for these dtypes and WW changing the limit on the min version of Koalas.

woodwork/tests/accessor/test_statistics.py
```diff
@@ -228,7 +229,8 @@ def test_describe_accessor_method(describe_df):
     datetime_ltypes = [Datetime]
     formatted_datetime_ltypes = [Datetime(datetime_format='%Y~%m~%d')]
     timedelta_ltypes = [Timedelta]
-    numeric_ltypes = [Double, Integer]
+    nullable_numeric_ltypes = [Double, IntegerNullable]
```
gsheni, line 232: Do we need to test Mutual Info with `IntegerNullable` and `Integer`?

gsheni: Same question for Boolean and BooleanNullable

thehomebrewnerd: I added a couple more tests cases for `make_categorical_for_mutual_info` and `replace_nans_for_mutual_info` to make sure they work with the different pandas dtypes. Our test fixture `df_same_mi` uses the `IntegerNullable` logical type, so I _think_ we are covered here.

@tamargrey Any thoughts since you are probably more familiar with the mutual info code than I am?

woodwork/type_sys/inference_functions.py
thehomebrewnerd: Updated #153 other issue to mention this.

rwedge:
Will this update invalidate the Schema version due to the definition of Boolean and Integer changing?

woodwork/tests/accessor/test_statistics.py
tamargrey: I think the mutual info tests look good. (I guess we will only ever have to replace nans for BooleanNullable and IntegerNullable, so that's interesting), but I am wondering if we should have `boolean_ltypes = [Boolean, BooleanNullable]` or maybe `BooleanNullable` in its own list for the describe tests.

thehomebrewnerd: Updated `describe_df` to include a missing value in the boolean data and added separate assertions for `Boolean` and `BooleanNullable` as with `Integer` and `IntegerNullable`.

thehomebrewnerd:
> Will this update invalidate the serialization schema version due to the definition of Boolean and Integer changing?

@rwedge Yeah, it should. Looks like our serialization/deserialization tests don't use any null values in the boolean column though, so they didn't break.

gsheni:
@thehomebrewnerd let's update our serialization tests to have nulls in Integer and Boolean column. Then we should also update the schema version.

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -1761,7 +1764,7 @@ def test_set_types(sample_df):
     pd.testing.assert_frame_equal(to_pandas(original_df), to_pandas(sample_df))
 
     sample_df.ww.set_types(logical_types={'is_registered': 'Integer'})
-    assert sample_df['is_registered'].dtype == 'Int64'
+    assert sample_df['is_registered'].dtype == 'int64'
```
tamargrey, line 1767: Thinking it might be useful to have a set of `set_types` checks that go from `Integer` to `IntegerNullable` and visa versa both when it's a valid conversion and when it's not. 

The stack trace for when it's not is this, and at first I was thinking that the final error we throw isn't super helpful for understanding why it's not allowed unless you have an in-depth understanding of pandas nullable dtypes, but I guess the `ValueError` that we catch is a part of the error, so maybe it's okay, but it might still be worth having an error message that further indicates it's a nullable dtype issue.

```python
~/.pyenv/versions/3.8.2/envs/woodwork/lib/python3.8/site-packages/pandas/core/arrays/masked.py in to_numpy(self, dtype, copy, na_value)
    219             ):
--> 220                 raise ValueError(
    221                     f"cannot convert to '{dtype}'-dtype NumPy array "

ValueError: cannot convert to 'int64'-dtype NumPy array with missing values. Specify an appropriate 'na_value' for this dtype.

During handling of the above exception, another exception occurred:

TypeConversionError                       Traceback (most recent call last)
<ipython-input-46-2489f63a00c9> in <module>
----> 1 sample_df.ww.init(logical_types={'age':'Integer'})
      2 display(sample_df.ww)
      3 sample_df.ww.set_types(logical_types={'is_registered':'Boolean'})

~/repositories/woodwork/woodwork/table_accessor.py in init(self, index, time_index, logical_types, make_index, already_sorted, schema, validate, use_standard_tags, **kwargs)
    139                 parsed_logical_types[name] = logical_type
    140 
--> 141                 updated_series = _update_column_dtype(series, logical_type)
    142                 if updated_series is not series:
    143                     self._dataframe[name] = updated_series

~/repositories/woodwork/woodwork/accessor_utils.py in _update_column_dtype(series, logical_type)
     87                 f'to type {new_dtype}. Please confirm the underlying data is consistent with ' \
     88                 f'logical type {logical_type}.'
---> 89             raise TypeConversionError(error_msg)
     90     return series
     91 

TypeConversionError: Error converting datatype for age from type Int64 to type int64. Please confirm the underlying data is consistent with logical type Integer.
```

tamargrey (commented):
Looking good! A few suggestions of nullable/non nullable tests to add, but some of them might definitely be overkill or already being tested implicitly, so feel free to pick and choose if you want.

Also, it could be useful to add a `select` test that checks different combinations of `numeric` tag and nullble/non nullable logical types to confirm we're correctly differentiating between them

woodwork/deserialize.py
```diff
@@ -57,6 +59,43 @@ def _typing_information_to_woodwork_table(table_typing_info, validate, **kwargs)
...
+        column_metadata[col_name] = col['metadata']
+        use_standard_tags[col_name] = col['use_standard_tags']
+
+        if col['physical_type']['type'] == 'category':
+            # Make sure categories are recreated properly
+            cat_values = col['physical_type']['cat_values']
+            if table_type == 'pandas' and pd.__version__ > '1.1.5':
+                cat_object = pd.CategoricalDtype(pd.Index(cat_values, dtype='object'))
+            else:
+                cat_object = pd.CategoricalDtype(pd.Series(cat_values))
```
thehomebrewnerd, line 93: Had to do this messy logic because of a difference in behavior between pandas 1.1.5 and 1.2.0 and greater. Since Koalas won't work with >1.1.5 yet, this was the only solution I found to allow tests to pass with Koalas using pandas <1.2.0 and other installs using pandas >=1.2.0.

Once Koalas removes their restriction, we can update our pandas min version and clean this up.

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -1189,7 +1192,7 @@ def test_dataframe_methods_on_accessor(sample_df):
 
     pd.testing.assert_frame_equal(to_pandas(schema_df), to_pandas(copied_df))
 
-    ltype_dtype = 'Int64'
+    ltype_dtype = 'int64'
```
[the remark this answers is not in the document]
thehomebrewnerd: Added this. Incidentally, Koalas changes the series dtype when doing fillna, which invalidates the schema:
```
>>> ks.Series([1, 2, pd.NA], dtype='Int64').fillna(0)
0    1
1    2
2    0
dtype: int64
```

woodwork/tests/accessor/test_table_accessor.py
thehomebrewnerd: I added a test to test the valid and invalid conversions.

I understand your concern about the error message clarity. Since this is the same error that gets raised for any other invalid type conversion, I'm a little hesitant to add special logic to catch invalid conversion for the nullable/non-nullable dtypes and change the message based on that. I'm not sure it's worth the extra complexity in our error handling, but happy to discuss more if needed.

tamargrey: yeah, as long as users are willing/able to look further into the stack trace, they'll see the pandas error thrown, so I probably agree that it's not worth the extra complexity

woodwork/deserialize.py
tamargrey: Ah, well that's another reason to look forward to koalas fixing that restriction. Should we make an issue for this?

woodwork/deserialize.py
```diff
@@ -57,6 +59,41 @@ def _typing_information_to_woodwork_table(table_typing_info, validate, **kwargs)
...
+
+        if col['physical_type']['type'] == 'category':
+            # Make sure categories are recreated properly
+            cat_values = col['physical_type']['cat_values']
+            cat_dtype = col['physical_type']['cat_dtype']
+            cat_object = pd.CategoricalDtype(pd.Series(cat_values, dtype=cat_dtype))
+            dtypes[col_name] = cat_object
+        elif not (ks and col['physical_type']['type'] == 'object'):
+            # Can't specify `object` for koalas
+            dtypes[col_name] = col['physical_type']['type']
```
tamargrey, line 97: Why are we ignoring columns that indicate they've been stored with `'object'` dtype? This isn't possible in koalas, right? How will LatLong columns get handled? A user could have koalas installed but be using a Pandas DataFrame with LatLong columns, so maybe this needs to be looking at `table_type`?

I guess I'm not understanding what's happening here.

thehomebrewnerd: Let me take another look at this. This might have been a change that became obsolete after later changes, and never got removed.

thehomebrewnerd: Yeah, this can be removed. At one point I was setting the dtype for all columns in the dtypes dict, but then changed to doing it only for the category columns. This was necessary to handle the Koalas issues around the `object` dtype, but it no longer needed.

It has been removed!

tamargrey: awesome!

woodwork/deserialize.py
```diff
@@ -57,6 +59,43 @@ def _typing_information_to_woodwork_table(table_typing_info, validate, **kwargs)
     kwargs = loading_info.get('params', {})
     table_type = loading_info.get('table_type', 'pandas')
 
+    logical_types = {}
+    semantic_tags = {}
+    column_descriptions = {}
+    column_metadata = {}
+    use_standard_tags = {}
+    dtypes = {}
```
tamargrey, line 67: We only add categorical dtypes to this, right? Can we call this something like `dtype_categories` or `category_cols`?

woodwork/deserialize.py
```diff
@@ -57,6 +59,43 @@ def _typing_information_to_woodwork_table(table_typing_info, validate, **kwargs)
...
+            # Make sure categories are recreated properly
+            cat_values = col['physical_type']['cat_values']
+            if table_type == 'pandas' and pd.__version__ > '1.1.5':
+                cat_object = pd.CategoricalDtype(pd.Index(cat_values, dtype='object'))
+            else:
+                cat_object = pd.CategoricalDtype(pd.Series(cat_values))
+            dtypes[col_name] = cat_object
+        # elif not (ks and col['physical_type']['type'] == 'object'):
+        #     # Can't specify `object` for koalas
+        #     dtypes[col_name] = col['physical_type']['type']
```
tamargrey, line 97: comments

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -2250,18 +2248,14 @@ def test_accessor_types(sample_df):
...
-        'age': int_dtype,
-        'signup_date': 'datetime64[ns]',
-        'is_registered': boolean_dtype,
+        'id': Integer.primary_dtype,
+        'full_name': NaturalLanguage.primary_dtype,
+        'email': NaturalLanguage.primary_dtype,
+        'phone_number': NaturalLanguage.primary_dtype,
+        'age': IntegerNullable.primary_dtype,
+        'signup_date': Datetime.primary_dtype,
+        'is_registered': BooleanNullable.primary_dtype,
```
tamargrey, line 2258: Nice! this makes a lot of sense!!

woodwork/tests/test_utils.py
```diff
@@ -204,7 +210,7 @@ def test_read_csv_with_pandas_params(sample_df_pandas, tmpdir):
     sample_df_pandas.to_csv(filepath, index=False)
     nrows = 2
 
-    df_from_csv = ww.read_csv(filepath=filepath, nrows=nrows)
+    df_from_csv = ww.read_csv(filepath=filepath, nrows=nrows, dtype={'age': 'Int64', 'is_registered': 'boolean'})
```
tamargrey, line 213: yay kwargs letting us get this behavior without making any changes to `read_csv`!

woodwork/serialize.py
```diff
@@ -42,7 +46,8 @@ def typing_info_to_dict(dataframe):
              'type': str(_get_ltype_class(col.logical_type))
          },
          'physical_type': {
-             'type': str(dataframe[col_name].dtype)
+             'type': str(dataframe[col_name].dtype),
+             'cat_values': dataframe[col_name].dtype.categories.to_list() if str(dataframe[col_name].dtype) == 'category' else None
```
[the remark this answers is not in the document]
thehomebrewnerd: Added comment

woodwork/deserialize.py
```diff
@@ -12,7 +12,9 @@
 from woodwork.exceptions import OutdatedSchemaWarning, UpgradeSchemaWarning
 from woodwork.s3_utils import get_transport_params, use_smartopen
 from woodwork.serialize import FORMATS, SCHEMA_VERSION
-from woodwork.utils import _is_s3, _is_url, import_or_raise
+from woodwork.utils import _is_s3, _is_url, import_or_none, import_or_raise
+
+ks = import_or_none('databricks.koalas')
```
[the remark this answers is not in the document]
thehomebrewnerd: Nope - good catch! Removed.

woodwork/deserialize.py
thehomebrewnerd: Updated name

tamargrey (approved):
Maybe worth waiting till someone else is also able to take another look, but LGTM!

jeff-hernandez (approved):

PR #832, opened 2021-04-16
Expand ColumnSchema Tests

### LINKED ISSUE #797: Move schema-specific logic from WoodworkColumnAccessor to ColumnSchema
When we didn't have a `ColumnSchema` class for woodwork logic to live, we added relevant methods to the WoodworkColumnAccessor, and while we shouldn't lose these methods from the api reference, if any of them have logic that doesn't need the actual `series`, then that code should be moved to a `ColumnSchema` method.

This follows how the Table Schema/Accessor share logic--we keep the dataframe-reliant logic on the Accessor and everything else on the Schema.

tamargrey:
- There was a lot of testing logic for semantic tag updates that only existed on the `WoodworkColumnAccessor`, which meant that when making changes to any of the semantic tag update ColumnSchema methods you then had to run the column accessor tests.
- In moving over these tests, added extra checks for the `logical_type=None` case and discovered a few bugs in `is_numeric` `is_categorical` and `_remove_semantic_tags`, so added tests and fixed
- Closes #797


thehomebrewnerd (approved):
Looks good!

PR #833, opened 2021-04-16
Add validation control to column accessor init

### LINKED ISSUE #737: Add init parameter to disable validation for WoodworkColumnAccessor
During initialization, Woodwork performs numerous validation checks on the inputs and at times, the underlying data. These checks are currently performed every time Woodwork is initialized, and may be costly. At times users may know their data is valid and might want to skip the validation checks to improve performance. 

We should add a new parameter to the `init` methods for `WoodworkColumnAccessor` allow users to optionally skip validation during initialization. This value will also need to be propagated to some other methods that are called during the initialization process.

Validation control in Woodwork would work as follows:
```python
series.ww.init() # init series with validation
series.ww.init(validate=True) # init series with validation
series.ww.init(validate=False) #init series without validation
```

Initializing Woodwork with `validate=False` will cause all validation to be skipped including validation of the input arguments as well as the underlying data. The docstring for this parameter should include a warning that users should only use this parameter if they are certain they can safely skip validation, otherwise difficult to diagnose errors may occur.

thehomebrewnerd:
- Add validation control to column accessor init
- Closes #737 

This PR adds a validate parameter to the WoodworkColumnAccessor `init` method to allow users to optionally skip validation checks during initialization. Default behavior is for validation to happen.

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -14,6 +14,7 @@ Release Notes
         * Add ``WoodworkColumnAccessor.schema`` and handle copying column schema (:pr:`799`)
         * Allow initializing a ``WoodworkColumnAccessor`` with a ``ColumnSchema`` (:pr:`814`)
         * Add ``__repr__`` to ``ColumnSchema`` (:pr:`817`)
+        * Add validation control to WoodworkColumnAccessor (:pr:`833`)
     * Fixes
     * Changes
         * Rename ``FullName`` logical type to ``PersonFullName`` (:pr:`740`)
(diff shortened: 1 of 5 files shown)


tamargrey:
We can probably set `validate=False` for a couple of the places where we initialize Woodwork on a series that we're grabbing from a table (pop, getitem, indexers, _make_series_call) if we want.

thehomebrewnerd:
> We can probably set `validate=False` for a couple of the places where we initialize Woodwork on a series that we're grabbing from a table (pop, getitem, indexers, _make_series_call) if we want.

Ah, yeah. Good point. I'll review and update.

gsheni (approved):
lgtm

thehomebrewnerd:
> We can probably set `validate=False` for a couple of the places where we initialize Woodwork on a series that we're grabbing from a table (pop, getitem, indexers, _make_series_call) if we want.

@tamargrey Added `validate=False` in a few locations that should be safe (I think).

tamargrey (approved):
lgtm!

PR #837, opened 2021-04-21
Use ColumnNotPresentError for dataframe/input mismatches

### LINKED ISSUE #649: Use ColumnNotPresentError when raising an error for column not found
- There is multiple places we check whether a column exists or not before doing some sort of action. 
  - rename, drop, pop, reset_semantic_tags, _check_logical_types 
- For rename, we ColumnNotPresentError error. We should do the same in other locations. 
- Currently, we are raising LookUpError in the majority of places where the checking is being done.

thehomebrewnerd:
- Use ColumnNotPresentError for dataframe/input mismatches
- Closes #649 

Updates code in multiple places to use a ColumnNotPresentError in place of the previous LookupError when there is a mismatch between columns specified by users in input and the columns present in a dataframe/schema.

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -7,6 +7,7 @@ Release Notes
         * Add ``is_schema_valid`` and ``get_invalid_schema_message`` functions for checking schema validity (:pr:`834`)
     * Fixes
     * Changes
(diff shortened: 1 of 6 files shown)


tamargrey:
I think one additional place where this error would be useful is in `drop` where we currently have `ValueError(f'{not_present} not found in DataFrame')`

tamargrey (approved):
lgtm!

PR #838, opened 2021-04-21
Create WoodworkNotInitError for use in accessing woodwork attributes before initialization

### LINKED ISSUE #621: Add custom error when accessing Woodwork properties/methods before initilization
Currently when a user attempts to access a Woodwork property or method prior to initializing Woodwork with the new accessor approach, or uses `loc` or `iloc` before init, a `AttributeError` is raised. In many other cases we have created custom Woodwork error and warning classes. We should consider replacing the `AttributeError` with a custom `WoodworkNotInitError` or something comparable to be consistent with other error/warning approaches.

This should be done on both WoodworkColumnAccessor and WoodworkTableAccessor.

thehomebrewnerd:
- Create WoodworkNotInitError for use in accessing woodwork attributes before initialization
- Closes #621 

This PR adds a new custom `WoodworkNotInitError` that is used when users attempt to access attributes through the Woodwork accessor before initialization, replacing the generic `AttributeError` that was used previously.

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -7,6 +7,7 @@ Release Notes
         * Add ``is_schema_valid`` and ``get_invalid_schema_message`` functions for checking schema validity (:pr:`834`)
     * Fixes
     * Changes
+        * Raise custom ``WoodworkNotInitError`` when accessing Woodwork attributes before initialization (:pr:`838`)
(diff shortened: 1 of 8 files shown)


tamargrey (approved):
Nice that the changes are so clean! Looks good to me!

PR #849, opened 2021-04-22
Add Logical Type for Age and AgeNullable

### LINKED ISSUE #778: Add Age Logical Type
- There are many datasets where you have the age of a person listed (such as titanic).

```python
class Age(LogicalType):
    primary_dtype = 'Int64'
    standard_tags = {'numeric'}
```
- Only 1 - 2 tests needed to be added with this Logical Type
- The test fixture using the `age` column, can be renamed. We do not need to use the `Age` Logical Type in the test fixture.

jeff-hernandez:
Closes #778 by adding a logical type for age.

### DIFF
--- a/woodwork/logical_types.py
+++ b/woodwork/logical_types.py
@@ -31,6 +31,19 @@ class LogicalType(object, metaclass=LogicalTypeMetaClass):
         return str(self.__class__)
 
 
+class Age(LogicalType):
+    """Represents Logical Types that contain positive numbers. Has 'numeric' as a standard tag.
+
+    Examples:
+        .. code-block:: python
+
+            [15, 22, 45]
+            [30, 62, 87]
+    """
+    primary_dtype = 'Int64'
+    standard_tags = {'numeric'}
+
+
 class Boolean(LogicalType):
     """Represents Logical Types that contain binary values indicating true/false.
 
--- a/woodwork/logical_types.py
+++ b/woodwork/logical_types.py
@@ -32,7 +32,7 @@ class LogicalType(object, metaclass=LogicalTypeMetaClass):
 
 
 class Age(LogicalType):
-    """Represents Logical Types that contain positive numbers. Has 'numeric' as a standard tag.
+    """Represents Logical Types that contain non-negative numbers indicating a person's age. Has 'numeric' as a standard tag.
 
     Examples:
         .. code-block:: python
@@ -45,7 +45,7 @@ class Age(LogicalType):
 
 
 class AgeNullable(LogicalType):
-    """Represents Logical Types that contain positive numbers. Has 'numeric' as a standard tag. May also contain null values.
+    """Represents Logical Types that contain non-negative numbers indicating a person's age. Has 'numeric' as a standard tag. May also contain null values.
 
     Examples:
         .. code-block:: python
--- a/woodwork/logical_types.py
+++ b/woodwork/logical_types.py
@@ -40,6 +40,19 @@ class Age(LogicalType):
             [15, 22, 45]
             [30, 62, 87]
     """
+    primary_dtype = 'int64'
+    standard_tags = {'numeric'}
+
+
+class AgeNullable(LogicalType):
+    """Represents Logical Types that contain positive numbers. Has 'numeric' as a standard tag. May also contain null values.
+
+    Examples:
+        .. code-block:: python
+
+            [np.nan, 22, 45]
+            [30, 62, np.nan]
+    """
     primary_dtype = 'Int64'
     standard_tags = {'numeric'}
 
--- a/woodwork/tests/logical_types/test_logical_types.py
+++ b/woodwork/tests/logical_types/test_logical_types.py
@@ -1,6 +1,6 @@
 import pytest
 
-from woodwork.logical_types import Boolean, Categorical, Datetime, Ordinal
+from woodwork.logical_types import Age, Boolean, Categorical, Datetime, Ordinal
 
 
 def test_logical_eq():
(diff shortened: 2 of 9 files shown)


jeff-hernandez:
How do we want to test this logical type?

gsheni:
@jeff-hernandez We can do minimal test for this and make sure we can init with Age as well as making sure we can set a Logical Type of Age.

woodwork/logical_types.py
```diff
@@ -31,6 +31,19 @@ def __str__(self):
...
+class Age(LogicalType):
+    """Represents Logical Types that contain positive numbers. Has 'numeric' as a standard tag.
+
+    Examples:
+        .. code-block:: python
+
+            [15, 22, 45]
+            [30, 62, 87]
+    """
+    primary_dtype = 'Int64'
```
tamargrey, line 43: Does having the nullable dtype here mean we need to have a corresponding `AgeNullable` logical type as well, or is that just because of the type inference that occurs with Integer that we need to allow for inferring with and without nullable dtypes? @gsheni @thehomebrewnerd

gsheni: @tamargrey that's an interesting point and something I hadn't thought of. 

Perhaps to keep consistency, we should have Age (int64) and AgeNullable (Int64). What do you think @jeff-hernandez @tamargrey ?

jeff-hernandez: I think it makes sense to have a corresponding nullable type for consistency. When would a user want `Age` specifically instead of `AgeNullable`? I kind of thought the nullable types were a superset of the non-nullable types.

gsheni: We have 1 other Logical Type that has Int64 as the dtypes (IntegerNullable). If a user does not have any NaNs we use int64 and Integer. 
This Age Logical Type will be another Int64/int64 dtype and so we want to get a similar behavior has Integer. 

If a user has NaNs, they use AgeNullable and if they don't they use Age.

tamargrey: I think I'm okay with having an `AgeNullable`, but we might want to have something in the docstrings that indicate that Age won't handle null values. Since with boolean and integer, you get different inference, but with Age you have to know to use it.

jeff-hernandez: Sounds good. I'll add `AgeNullable` to this PR.

tamargrey:
I'm not 100% versed on when we do and don't need a relationship in our type system between two logical types, but I think we might want `Integer` to be a parent to `Age` and `IntegerNullable` to be a parent to `AgeNullable` in our `DEFAULT_RELATIONSHIPS`

tamargrey (approved):
Lgtm, but might be worth double checking with @thehomebrewnerd on the AgeNullable and relationship stuff makes sense.

woodwork/tests/logical_types/test_logical_types.py
```diff
@@ -40,3 +52,24 @@ def test_ordinal_init_with_order():
...
+def test_set_types(sample_df, column, logical_type):
+    sample_df.ww.init()
+    before = sample_df.ww[column].ww.logical_type
+    info = f'"{column}" already set as "{logical_type}""'
+    assert before != logical_type, info
+
+    sample_df.ww.set_types(logical_types={column: logical_type})
+    after = sample_df.ww[column].ww.logical_type
+    info = f'"{column}" not set as "{logical_type}"'
+    assert after == logical_type, info
```
thehomebrewnerd, line 75: I'm thinking we do not need these specific tests here. Instead we should update our tests such as `test_sets_int64_dtype_on_init()` to include these new types in the list of logical types that get used. Additionally, we should review our mutual information and describe tests to make sure these types are includes, as they should be valid columns as input for those functions.

thehomebrewnerd (commented):
The way the relationships are set up makes sense to me.

One other thing we need to do, in addition to the other comments I made, is add these new logical types to the API Reference.

woodwork/tests/logical_types/test_logical_types.py
jeff-hernandez: Are there existing tests that check if columns can be set to logical types using `set_types` for all logical types?

thehomebrewnerd: No we don't check setting the logical types after init for all logical types. However, the key steps involved in setting a type after init are the same as those during init, namely parsing the logical type with the `_parse_logical_type` function and updating the column dtype with the `_update_column_dtype` function. 

Based on that I'm thinking it is not necessary to test all logical types in both places (init and set_types) unless there is something unique here that isn't covered by the other tests.

jeff-hernandez: I can see how those tests are arguably not needed. Let's review with @gsheni since he asked to test setting this logical type (https://github.com/alteryx/woodwork/pull/849#issuecomment-825322199)

gsheni: @jeff-hernandez I was mistaken and @thehomebrewnerd is correct. 

Let's update `test_sets_int64_dtype_on_init` to include this new types. Mutual Info should work with Age and AgeNullable. Describe should work as well.

jeff-hernandez: @gsheni okay, will update

jeff-hernandez: @thehomebrewnerd which mutual info and describe tests do we want to include these types in?

thehomebrewnerd: For describe, I would update `test_describe_accessor_method` to include `AgeNullable` and `Age` in the lists described here: https://github.com/alteryx/woodwork/blob/ce17d7399c77d75f82930c514ea91dccfa980f5d/woodwork/tests/accessor/test_statistics.py#L240-L241

Adding them to the list should be all you need to do, and will cause them to get tested below in the appropriate sections of this test where those lists get used.

thehomebrewnerd: As for mutual info, I'm thinking the update you already made to confirm `Age` and `AgeNullable` are returned by `get_valid_mi_types` should be sufficient. I'm not thinking there is additional value in testing a mutual information calculation with those types actually used.

I think we also need to update the docstrings for the accessor `mutual_information` and `mutual_information_dict` methods to include these logical types (and the nullable integer and boolean types, which we neglected to add before).

woodwork/logical_types.py
```diff
@@ -31,6 +31,32 @@ def __str__(self):
         return str(self.__class__)
 
 
+class Age(LogicalType):
+    """Represents Logical Types that contain positive numbers. Has 'numeric' as a standard tag.
```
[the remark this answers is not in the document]
jeff-hernandez: ✅ 47b741d

thehomebrewnerd (approved):
LGTM!

PR #853, opened 2021-04-23
Remove codecov token

rwedge:
This token is not needed on public repos


rwedge:
Looks like the upload worked

tamargrey (approved):

PR #858, opened 2021-04-26
Add logical type for Address

### LINKED ISSUE #779: Add Address Logical Type
- Currently, this is detected as a NaturalLanguage column:
```
1 Miller Drive, New York, NY 12345
1 Berkeley Street Boston, MA 67891
...
```
- However, it is common to have an address column in datasets, and as such it should be a specific Logical Type

```python
class Address(LogicalType):
    primary_dtype = 'string'
    standard_tags = {}
```
- Only a 1-2 tests with using this logical type need to be added.

jeff-hernandez:
Closes #779 by adding logical type for `Address`

### DIFF
--- a/woodwork/logical_types.py
+++ b/woodwork/logical_types.py
@@ -31,6 +31,17 @@ class LogicalType(object, metaclass=LogicalTypeMetaClass):
         return str(self.__class__)
 
 
+class Address(LogicalType):
+    """Represents Logical Types that contain address values.
+
+    Examples:
+        .. code-block:: python
+            ['1 Miller Drive, New York, NY 12345', '1 Berkeley Street, Boston, MA 67891']
+            ['26387 Russell Hill, Dallas, TX 34521', '54305 Oxford Street, Seattle, WA 95132']
+    """
+    primary_dtype = 'string'
+
+
 class Age(LogicalType):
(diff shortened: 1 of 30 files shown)


woodwork/logical_types.py
```diff
@@ -31,6 +31,19 @@ def __str__(self):
...
+    """Represents Logical Types that contain address values.
+
+    Examples:
+        .. code-block:: python
+
+            ['1 Miller Drive, New York, NY 12345', '1 Berkeley Street, Boston, MA 67891']
+            ['26387 Russell Hill, Dallas, TX 34521', '54305 Oxford Street, Seattle, WA 95132']
+    """
+    primary_dtype = 'string'
+    standard_tags = set()
```
thehomebrewnerd, line 44: Not a big deal, but you don't need to define this here - it will inherit from `LogicalType` since the default is an empty set.

thehomebrewnerd (commented):
Let's update this PR to test things like you are doing in the `Age` PR, namely updating `test_sets_string_dtype_on_init` to include this logical type in the list of logical_types.

Also, make sure to add this new logical type to the API reference as well.

thehomebrewnerd (approved):
LGTM!

PR #859, opened 2021-04-26
Check tags for index and time index

### LINKED ISSUE #301: Don't allow index and time_index to be the same column
Right now, we allow a DataTable to be made where a single column is both the `index` and the `time_index`. 
This is bad practice as an index is meant to be used as a lookup key while the time index tells you the relative order in time for a row as it relates to the index. 

```python
import pandas as pd
from woodwork import DataTable
df = pd.DataFrame({
        'id': pd.Series([1, 2, 3]),
    })
dt = DataTable(df, index='id', time_index='id')
assert dt['id'].semantic_tags == {'index', 'time_index'}
assert dt.index == dt.time_index
```
#### Expected Output
We should catch this at parameter validation and also not allow this to happen when `set_index` or `set_time_index` is used.

The error message could be something like
`ValueError: The index column must be different from the time index. Please change one of the columns being assigned as index or time index.`

jeff-hernandez:
Closes #301 by validating the semantic tags when setting a column as the index or time index.

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -6,6 +6,7 @@ Release Notes
     * Enhancements
         * Add ``is_schema_valid`` and ``get_invalid_schema_message`` functions for checking schema validity (:pr:`834`)
     * Fixes
+        * Raise error when a column is set as the index and time index (:pr:`859`)
     * Changes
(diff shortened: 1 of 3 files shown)


tamargrey:
Can we also add a test where we try and set index and time_index to be the same column at TableSchema init?

tamargrey (approved):
lgtm!

PR #862, opened 2021-04-27
Allow NaNs in schema validation check

### LINKED ISSUE #861: Index column with NaN will always return invalid schema
Our check in `get_invalid_schema_message` for whether the underlying index exactly matches the Woodwork index column uses `all(dataframe.index == dataframe[schema.index])`, which means NaNs will not evaluate to True, so we say that the schema is invalid. 

```python
df = pd.DataFrame({'id':pd.Series([5,4,None,2], dtype='float64'), 'col':pd.Series(['b','b','b','d'], dtype='category')})
df.ww.init(index='id')
schema = df.ww.schema

assert not ww.is_schema_valid(df, schema)
```
We should add a test case with nans in the index column and add a fix that includes nans in the check.

This came up in a Featuretools test for `add_last_time_index` where we use `update_dataframe`.

tamargrey:
- Stops indicating that the schema is invalid if there are null values in the index column
- Closes #861

### DIFF
--- a/woodwork/accessor_utils.py
+++ b/woodwork/accessor_utils.py
@@ -153,8 +153,7 @@ def get_invalid_schema_message(dataframe, schema):
                 f'{df_dtype}, and {schema.logical_types[name]} dtype, {valid_dtype}'
     if schema.index is not None and isinstance(dataframe, pd.DataFrame):
         # Index validation not performed for Dask/Koalas
-        if not all((dataframe.index == dataframe[schema.index]) |
-                   (pd.isnull(dataframe.index) & pd.isnull(dataframe[schema.index]))):
+        if not pd.Series(dataframe.index).equals(pd.Series(dataframe[schema.index].values)):
             return 'Index mismatch between DataFrame and typing information'
         elif not dataframe[schema.index].is_unique:
             return 'Index column is not unique'
--- a/woodwork/accessor_utils.py
+++ b/woodwork/accessor_utils.py
@@ -153,7 +153,7 @@ def get_invalid_schema_message(dataframe, schema):
                 f'{df_dtype}, and {schema.logical_types[name]} dtype, {valid_dtype}'
     if schema.index is not None and isinstance(dataframe, pd.DataFrame):
         # Index validation not performed for Dask/Koalas
-        if not pd.Series(dataframe.index).equals(pd.Series(dataframe[schema.index].values)):
+        if not pd.Series(dataframe.index, dtype=dataframe[schema.index].dtype).equals(pd.Series(dataframe[schema.index].values)):
             return 'Index mismatch between DataFrame and typing information'
         elif not dataframe[schema.index].is_unique:
             return 'Index column is not unique'
--- a/woodwork/accessor_utils.py
+++ b/woodwork/accessor_utils.py
@@ -153,7 +153,8 @@ def get_invalid_schema_message(dataframe, schema):
                 f'{df_dtype}, and {schema.logical_types[name]} dtype, {valid_dtype}'
     if schema.index is not None and isinstance(dataframe, pd.DataFrame):
         # Index validation not performed for Dask/Koalas
-        if not all(dataframe.index == dataframe[schema.index]):
+        if not all((dataframe.index == dataframe[schema.index]) |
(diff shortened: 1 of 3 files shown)


woodwork/accessor_utils.py
```diff
@@ -153,7 +153,8 @@ def get_invalid_schema_message(dataframe, schema):
                 f'{df_dtype}, and {schema.logical_types[name]} dtype, {valid_dtype}'
     if schema.index is not None and isinstance(dataframe, pd.DataFrame):
         # Index validation not performed for Dask/Koalas
-        if not all(dataframe.index == dataframe[schema.index]):
+        if not all((dataframe.index == dataframe[schema.index]) |
+                   (pd.isnull(dataframe.index) & pd.isnull(dataframe[schema.index]))):
```
tamargrey, line 157: The other option I was considering was to fill nans and then keep the comparison the same from [here](https://stackoverflow.com/questions/19322506/pandas-dataframes-with-nans-equality-comparison).

Seems like it might be faster based on the size of the data , but not totally sure...

thehomebrewnerd: I haven't tested this extensively, but I think you can also do this: 
```not pd.Series(nan_df.index).equals(pd.Series(nan_df[schema.index].values))```

Some initial test comparisons:
image


This relative performance held true for 100,000 rows, 1,000,000 rows and 100,000,000 rows

Comment (tamargrey) [reply woodwork/accessor_utils.py:157]: ! That seems promising.

**Review by thehomebrewnerd — APPROVED** [review]
I think these changes look fine. Up to you if you want to try to optimize the `if` comparison.

File: woodwork/accessor_utils.py (hunk 1 above)
Comment (thehomebrewnerd) [reply woodwork/accessor_utils.py:157]: This was as a 2 column dataframe with a single `NaN` value in the last row of the index column.

Comment (thehomebrewnerd) [reply woodwork/accessor_utils.py:157]: Almost exactly the same times if I put the `NaN` value in the first row instead.

Comment (tamargrey) [reply woodwork/accessor_utils.py:157]: That seems reason enough to use `.equals` here. Thanks for looking into it!

Comment (tamargrey) [reply woodwork/accessor_utils.py]: A potential problem with using equals here is that it seems like a `pd.Index` cannot have the `string` dtype, so because it'll have the `object` dtype, the series will not be equal if the index is NaturalLanguage-like. 

```python
df = pd.DataFrame({'id':pd.Series([5,4,None,2], dtype='float64'), 'col':pd.Series(['b','k','x','d'], dtype='string')})
df.ww.init(index='col', logical_types={'col':'natural_language'})
schema = df.ww.schema

idx = df.index
col = df[schema.index]

assert df.index.astype('string').dtype == 'object'

idx_series = pd.Series(idx)
col_series = pd.Series(col.values)

assert not idx_series.equals(col_series)
```

tamargrey: @thehomebrewnerd  what do you think of this modification to handle the dtype issue:
```python
pd.Series(dataframe.index, dtype=dataframe[schema.index].dtype).equals(pd.Series(dataframe[schema.index].values))
```

thehomebrewnerd: Yeah, I was thinking something similar:

```
not pd.Series(nan_df.index, dtype='object').equals(pd.Series(nan_df[schema.index].values, dtype='object'))
```

thehomebrewnerd: I can double check the performance of those to make sure that doesn't do something strange...

thehomebrewnerd: Setting both to `object` slows it down quite a bit and makes this option the worst of the three for an integer index.

Your suggestion is still the fastest for an integer index and about a wash with the original code for a string index. The fillna approach didn't work in my test because of trying to add a new value to a 'category' dtype.

So I think your proposed solution should still be the best overall.

thehomebrewnerd: This is the comparison for 10,000,000 string rows:
image

tamargrey: updated to set the dtype of the index column!

thehomebrewnerd (approved):
LGTM, assuming CI checks all pass!

PR #863, opened 2021-04-27
Fix bug related to casting to bool in pandas >=1.2.0

### LINKED ISSUE #846: It's possible to create an invalid schema with ww.init
It's possible to create an invalid schema by typing a column with pandas dtype `category` and missing values as Boolean in ww 0.2.0.

I would expect `ww.init` to error out letting me know the type I requested is not compatible with my data, just like what happens when you try to go from `Category` to `Double`.

This isn't blocking evalml. Users can catch this with the `is_schema_valid` helper function introduced in https://github.com/alteryx/woodwork/pull/834 but it might be better to check the schema during `ww.init`. 

```python
import pandas as pd
import woodwork as ww
from woodwork.table_accessor import _get_invalid_schema_message

df = pd.DataFrame(pd.Series(["a", "b", None], dtype='category'))
df.ww.init(logical_types={0: "Boolean"})
assert _get_invalid_schema_message(df, df.ww.schema)
df
```
```
Out[4]:
      0
0  True
1  True
2   NaN
```

thehomebrewnerd:
- Fix bug related to casting to bool in pandas >=1.2.0
- Closes #846 

Fixes a bug when using versions of pandas >=1.2.0 in which setting a column logical type to `Boolean` when missing values were present did not raise a `TypeConversionError` as it should as pandas did not error, but also did not convert the column dtype to `bool` as requested.

### DIFF
--- a/woodwork/accessor_utils.py
+++ b/woodwork/accessor_utils.py
@@ -157,7 +157,7 @@ def get_invalid_schema_message(dataframe, schema):
                 f'{df_dtype}, and {schema.logical_types[name]} dtype, {valid_dtype}'
     if schema.index is not None and isinstance(dataframe, pd.DataFrame):
         # Index validation not performed for Dask/Koalas
-        if not all(dataframe.index == dataframe[schema.index]):
+        if not pd.Series(dataframe.index, dtype=dataframe[schema.index].dtype).equals(pd.Series(dataframe[schema.index].values)):
             return 'Index mismatch between DataFrame and typing information'
         elif not dataframe[schema.index].is_unique:
             return 'Index column is not unique'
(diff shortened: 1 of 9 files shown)


woodwork/accessor_utils.py
```diff
@@ -82,10 +85,11 @@ def _update_column_dtype(series, logical_type):
                     series = pd.to_datetime(series, format=logical_type.datetime_format)
             else:
                 series = series.astype(new_dtype)
+                if series.dtype != new_dtype:
```
tamargrey, line 88: Probably worth doing `if str(series.dtype) != new_dtype` to avoid potential weirdness incomparing a dtype object to a string

thehomebrewnerd: Updated!

tamargrey (approved):
lgtm!

PR #870, opened 2021-04-28
Remove Ordinal validation in ColumnSchema

### LINKED ISSUE #865: Remove Ordinal validation check in ColumnSchema
- If a user uses ColumnSchema with no data, we don't want check for the Ordinal ordering.
- The check here should be removed:
  -  https://github.com/alteryx/woodwork/blob/main/woodwork/column_schema.py#L166

thehomebrewnerd:
- Remove Ordinal validation in ColumnSchema
- Closes #865 

Removes the check in ColumnSchema that requires an instance of Ordinal to be passed. This does not remove the corresponding check in `_parse_logical_type` so attempting to initialize a series or table accessor without and instance will still raise an error.


tamargrey (approved):
LGTM, assuming tests pass!

rwedge (approved):

PR #885, opened 2021-05-03
Updates for koalas 1.8.0

### LINKED ISSUE #836: Remove pandas version specific code in deserialize.py and test_table_accessor.py
PR #830 added pandas version specific code in `deserialize.py` due to differences in behavior between pandas `1.1.5` and later versions. This is needed because Koalas currently restricts pandas to `<1.2.0`, so users that have installed Koalas but are using pandas could encounter issues without this code.

Once Koalas has been updated to support pandas `>=1.2.0`, Woodwork should make sure our pandas min version is `1.2.0` and remove the pandas version specific code here:
https://github.com/alteryx/woodwork/blob/5e3a6fca9947f1a4f32c507bb0bffa786e3b01bf/woodwork/deserialize.py#L88-L91

We should also remove the `if pd.__version__ > '1.1.5:` line from `test_table_accessor.py`, but leave the test code that is present in that block.

thehomebrewnerd:
- Updates for koalas 1.8.0
- Closes #836 

Several changes related to Koalas 1.8.0
- bump koalas min version to 1.8.0
- bump pandas min version to 1.2.0
- bump pyarrow min version to 3.0.0
- remove pandas specific code for versions <1.2.0
- Fix tests to use different type casting compatible with koalas 1.8.0
- Update minimum dependency txt files
- Remove numpy from requirements

### DIFF
--- a/docs/source/install.ipynb
+++ b/docs/source/install.ipynb
@@ -70,7 +70,7 @@
     "|-------------------|-------------|----------------------------------------|\n",
     "| boto3             | 1.10.45     | Required to read/write to URLs and S3  |\n",
     "| smart_open        | 5.0.0       | Required to read/write to URLs and S3  |\n",
-    "| pyarrow           | 2.0.0       | Required to serialize to parquet       |\n",
+    "| pyarrow           | 3.0.0       | Required to serialize to parquet       |\n",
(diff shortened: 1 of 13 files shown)


gsheni (approved):
lgtm

tamargrey (approved):
lgtm!

PR #888, opened 2021-05-04
Keep index and time tags for single column

### LINKED ISSUE #867: Keep index and time index tags when selecting a single column
Currently, we remove `index` and `time_index` tags from a Woodwork Column if it is created off of a Table via `loc`, `iloc`, or `__getitem__`. We should keep the index tags so that `df.ww.columns['col_name'] == df.ww['col_name'].ww.schema`

jeff-hernandez:
Closes #867

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -17,7 +17,7 @@ Future Release
     * Enhancements
     * Fixes
     * Changes
-        * Keep index and time index tags for single column (:pr:`887`)
+        * Keep index and time index tags for single column (:pr:`888`)
     * Documentation Changes
     * Testing Changes
 
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -8,23 +8,12 @@ Future Release
         * Add ``deep`` parameter to Woodwork Accessor and Schema equality checks (:pr:`889`)
     * Fixes
     * Changes
-    * Documentation Changes
-    * Testing Changes
-
-    Thanks to the following people for contributing to this release:
-    :user:`tamargrey`
-
-Future Release
-==============
-    * Enhancements
-    * Fixes
-    * Changes
         * Keep index and time index tags for single column when selecting from a table (:pr:`888`)
     * Documentation Changes
     * Testing Changes
 
     Thanks to the following people for contributing to this release:
-    :user:`jeff-hernandez`
+    :user:`jeff-hernandez`, :user:`tamargrey`
 
 v0.3.0 May 3, 2021
 ==================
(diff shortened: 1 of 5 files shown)


woodwork/indexers.py
```diff
@@ -42,7 +42,6 @@ def _process_selection(selection, original_data):
         elif _is_dataframe(original_data):
             # Selecting a single column from a DataFrame
             schema = original_data.ww.schema.columns[selection.name]
-            schema.semantic_tags = schema.semantic_tags - {'index'} - {'time_index'}
             if schema.use_standard_tags:
                 schema.semantic_tags |= schema.logical_type.standard_tags
```
tamargrey, line 46: Now that we're keeping the `index` tag, I'm not sure that we want to add standard tags back like this (and at the table accessor getitem) since at the Woodwork table level, you either have standard tags or the index tag but not both. 

Not sure if maybe at the column level it does make sense to have both. @thehomebrewnerd thoughts?

thehomebrewnerd: Yeah, my thought going into this was that when we grab a column from the dataframe, the column schema for the individual column should match the schema that is in the dataframe for that column - that way no mater where the user looks they see the same info.

tldr: I vote for not adding back in standard tags.

docs/source/release_notes.rst
```diff
@@ -12,6 +12,18 @@ Release Notes
 
 .. Thanks to the following people for contributing to this release:
 
+Future Release
+==============
+    * Enhancements
+    * Fixes
+    * Changes
+        * Keep index and time index tags for single column (:pr:`888`)
```
thehomebrewnerd, line 20: I'd suggest a little more context here - like adding `...when selecting from a table`

woodwork/indexers.py
jeff-hernandez: ✅   ceccc4d

docs/source/release_notes.rst
jeff-hernandez: ✅  fe673e2

woodwork/table_accessor.py
```diff
@@ -206,7 +206,6 @@ def __getitem__(self, key):
 
         series = self._dataframe[key]
         column = copy.deepcopy(self._schema.columns[key])
-        column.semantic_tags -= {'index', 'time_index'}
         if column.use_standard_tags:
             column.semantic_tags |= column.logical_type.standard_tags
```
jeff-hernandez, line 210: Do we want to remove standard tags here too?

tamargrey: yes! It should have the same behavior as the indexers

jeff-hernandez: ✅  5fe0b07

thehomebrewnerd (approved):
LGTM

tamargrey (approved):

PR #897, opened 2021-05-06
Use minimum dependency generator GitHub Action

### LINKED ISSUE #890: Create GitHub Action with minimum dependency checker
- We have a minimum dependency checker
  - https://github.com/alteryx/woodwork/tree/main/tools/minimum_dependency
- Instead of copying these files to our other OS projects, we can create a GH action that uses this

gsheni:
- Closes #890
- When deps have not changed, no pull request is generated:
  - [CI Run showing no PR is generated](https://github.com/alteryx/woodwork/pull/897/checks?check_run_id=2548693578)
- When deps have changed (test commit), pull request is generated:
  - [CI Run](https://github.com/alteryx/woodwork/pull/897/checks?check_run_id=2548751489)
  - [Pull Request](https://github.com/alteryx/woodwork/pull/907/files)


.github/workflows/minimum_dependency_checker.yml
```diff
@@ -0,0 +1,93 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+      - name: Run min dep generator - test reqs
+        id: min_dep_gen_test
+        uses: alteryx/minimum-dependency-generator@v2.1
```
rwedge, line 22: Do we want to point to v2.1 specifically, or use a generic v2 tag that points to the latest 2.x.y version (we can update the v2 tag to point to new releases as we add them)

.github/workflows/minimum_dependency_checker.yml
```diff
@@ -0,0 +1,93 @@
...
+        id: check_min_test
+        continue-on-error: true
+        run: |
+          mkdir /tmp/dependencies_updated_artifacts
+          printf "${{ steps.min_dep_gen_test.outputs.min_reqs }}" > /tmp/minimum_test_requirements.txt
+          diff /tmp/minimum_test_requirements.txt woodwork/tests/minimum_test_requirements.txt
+      - name: if min test deps have changed, write output file
+        if: steps.check_min_test.outcome != 'success'
+        run: |
+          printf "${{ steps.min_dep_gen_test.outputs.min_reqs }}" > woodwork/tests/minimum_test_requirements.txt
```
rwedge, line 35: this could be a single line since there is only one command, although I could see using the current approach due to line length

.github/workflows/minimum_dependency_checker.yml
gsheni: Fixed to point to v2 tag.

.github/workflows/minimum_dependency_checker.yml
gsheni: To make it easier for a developer reading the workflow, I did the multi-line approach. I tried removing it, and it's much more difficult to read, so I'll keep it as is for now.

rwedge (approved):
LGTM

PR #909, opened 2021-05-11
Add support for parquet to read_file

### LINKED ISSUE #839: Add support for reading Parquet file (read_file)
- As a user, I wish I could use Woodwork to read parquet files

#### Code Example

```python
from woodwork import read_file
df = read_file('data.parquet')
df = read_file('data', content_type="application/parquet")
```

thehomebrewnerd:
- Add support for parquet to read_file
- Closes #839 

Adds support from reading from a parquet file into a Woodwork dataframe using the `ww.read_file` function. Checks that `pyarrow` is installed and raises `ImportError` if not, instructing users to install pyarrow.

### DIFF
--- a/woodwork/utils.py
+++ b/woodwork/utils.py
@@ -27,6 +27,7 @@ PYARROW_ERR_MSG = (
 # Add new mimetypes
 add_type('application/parquet', '.parquet')
 
+
 def import_or_none(library):
     """Attempts to import the requested library.
 
--- a/woodwork/utils.py
+++ b/woodwork/utils.py
@@ -12,6 +12,7 @@ import woodwork as ww
 type_to_read_func_map = {
     'csv': pd.read_csv,
     'text/csv': pd.read_csv,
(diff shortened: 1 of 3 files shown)


woodwork/utils.py
```diff
@@ -115,6 +124,10 @@ def read_file(filepath=None,
     if content_type not in type_to_read_func_map:
         raise RuntimeError('Reading from content type {} is not currently supported'.format(content_type))
 
+    if content_type == 'application/parquet':
```
frances-h, line 127: It might be worthwhile to note that `guess_type` won't infer the type of a parquet file automatically. I think we could add it ourselves though using [`mimetypes.add_type`](https://docs.python.org/3/library/mimetypes.html#mimetypes.add_type), but users can also just manually specify the type as well instead.

thehomebrewnerd: I'm fine with either option here. @gsheni Do you have a preference?

PR #915, opened 2021-05-12
Add logical type method to transform data

### LINKED ISSUE #887: Spike: Investigate a generic way to transform user data inputted
- Currently, if a user changes a column Logical Type, we have if/else statements to check the new Logical Type.
- We want a generic way to transform data.
- Perhaps there is a function on each Logical Type, that defines how to transform the 
- The implementation should also remove the if/else statements we have when changing logical Types.
```python
class Address(LogicalType):

    primary_dtype = 'string' 
    def _validate_data(self, series):
          pass
    def transform_data(self, series):
         pass
```

jeff-hernandez:
Closes #887 by adding a logical type method to update and format column data. Logical types will be instantiated if they haven't been already.


thehomebrewnerd (commented):
I think this looks pretty good overall. Just a couple comments/questions.

woodwork/logical_types.py
```diff
@@ -30,6 +35,23 @@ def __eq__(self, other, deep=False):
     def __str__(self):
         return str(self.__class__)
 
+    def transform(self, series):
+        new_dtype = _get_valid_dtype(type(series), self)
```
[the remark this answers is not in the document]
jeff-hernandez: No reason, we can certainly do that. We can also make it a class method since it doesn't require any instance-specific attributes.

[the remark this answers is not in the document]
thehomebrewnerd: Good point. I wasn't thinking about the dtypes being class attributes.

woodwork/tests/test_utils.py
```diff
@@ -542,21 +542,21 @@ def test_get_valid_mi_types():
...
+    assert isinstance(_parse_logical_type('Datetime', 'col_name'), Datetime)
+    assert isinstance(_parse_logical_type(Datetime, 'col_name'), Datetime)
 
     ymd_format = Datetime(datetime_format='%Y-%m-%d')
     assert _parse_logical_type(ymd_format, 'col_name') == ymd_format
 
 
 def test_parse_logical_type_errors():
-    error = 'Must use an Ordinal instance with order values defined'
+    error = "missing 1 required positional argument: 'order'"
```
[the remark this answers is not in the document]
jeff-hernandez: Sure, I'll raise the original error inside `Ordinal.__init__`

woodwork/logical_types.py
```diff
@@ -149,6 +171,19 @@ class Datetime(LogicalType):
...
+        if new_dtype != str(series.dtype):
+            if dd and isinstance(series, dd.Series):
+                name = series.name
+                series = dd.to_datetime(series, format=self.datetime_format)
+                series.name = name
+            elif ks and isinstance(series, ks.Series):
+                series = ks.Series(ks.to_datetime(series.to_numpy(), format=self.datetime_format), name=series.name)
+            else:
+                series = pd.to_datetime(series, format=self.datetime_format)
+        return series
```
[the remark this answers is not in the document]
jeff-hernandez: Nope, must have missed it. Thanks!

woodwork/logical_types.py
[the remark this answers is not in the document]
jeff-hernandez: ✅ 6f118b5

woodwork/tests/test_utils.py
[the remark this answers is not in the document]
jeff-hernandez: ✅ 16163b0

woodwork/logical_types.py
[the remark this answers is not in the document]
jeff-hernandez: ✅ 1f2f6cb

woodwork/column_accessor.py
```diff
@@ -270,7 +270,7 @@ def _validate_logical_type(self, logical_type):
 
         if isinstance(logical_type, Ordinal):
             logical_type._validate_data(self._series)
-        elif logical_type == LatLong:
+        elif isinstance(logical_type, LatLong):
```
tamargrey, line 273: I was going to say we should be using `_get_ltype_class` here (and we probably should have before), but it got me thinking about whether we need `_get_ltype_class` anymore. The reason we needed it before was because we needed a way of comparing logical types when some would be instantiated and some would not be, but now that they're always going to be instances, I think we can always use `isinstance` anywhere that we're using `_get_ltype_class`.

woodwork/logical_types.py
```diff
@@ -149,6 +178,19 @@ class Datetime(LogicalType):
...
+        new_dtype = self._get_valid_dtype(type(series))
+        if new_dtype != str(series.dtype):
+            if dd and isinstance(series, dd.Series):
+                name = series.name
+                series = dd.to_datetime(series, format=self.datetime_format)
+                series.name = name
+            elif ks and isinstance(series, ks.Series):
+                series = ks.Series(ks.to_datetime(series.to_numpy(), format=self.datetime_format), name=series.name)
+            else:
+                series = pd.to_datetime(series, format=self.datetime_format)
```
tamargrey, line 191: What's the error here if there's an error while converting to `datetime64[ns]`? Before it was getting caught in the try except, so if we've lost the "Error converting datatype" message here, I think it'd be worth adding it back in.

woodwork/tests/accessor/test_accessor_utils.py
```diff
@@ -2,90 +2,19 @@
...
-    Datetime,
-    NaturalLanguage
-)
 from woodwork.utils import import_or_none
 
 dd = import_or_none('dask.dataframe')
 ks = import_or_none('databricks.koalas')
 
 
-def test_init_series_valid_conversion_specified_ltype(sample_series):
```
tamargrey, line 25: Even though this test and the three below it all use `_get_valid_dtype`, I think they are mainly testing `init_series`, so we probably want them to stay in `test_accessor_utils`

tamargrey (commented):
Just a couple of comments--I'm really liking these changes!!

woodwork/column_accessor.py
jeff-hernandez: I'll refactor to remove `_get_ltype_class`

woodwork/logical_types.py
jeff-hernandez: I don't mind adding this back in. Did we have a test to check this error?

tamargrey: I couldn't find one, but that was probably because we weren't differentiating between datetime and the other logical types with that error, but now we'll need to add it specially to `Datetime.transform` because it wouldn't get caught in the super() call.

woodwork/tests/accessor/test_accessor_utils.py
jeff-hernandez: Let's check with @thehomebrewnerd who suggested this change ([comment](https://github.com/alteryx/woodwork/pull/915#discussion_r632054205)) to get more details on which associated tests he's referring to.

woodwork/logical_types.py
jeff-hernandez: I can create a test for this. Do we want it in `test_logical_types`?

tamargrey: hmmm that's a good question. It looks like we mostly have tested that error through the DataFrame, so `test_invalid_dtype_casting` is in `test_table_accessor`. I'd say if you're going to go through a DataFrame to test it, it should probably be in `test_table_accessor` as well. But if you want to just test something like `Datetime.transform(pd.Series(..))` then it should happen in `test_logical_types`. 

Maybe we should have a set of tests that just test the various transform cases on a series passed to the logical type method instead of always going through the DataFrame.

jeff-hernandez: I like the idea of testing various transform cases. It'd help to highlight the expected behavior of a transform.

woodwork/tests/accessor/test_accessor_utils.py
thehomebrewnerd: @jeff-hernandez Sorry, I wasn't very clear in my original comment. I'm thinking we should move `test_get_valid_dtype` since that is specifically testing we get the expected dtype value and that is no longer a utility function.  

We should leave the other `test_init_series*` tests where they were as they are really testing out that the `init_series` utility function behaves as expected in various circumstances.

jeff-hernandez: Okay, no problem. I'll update the tests.

jeff-hernandez:
Based on [this test](https://github.com/alteryx/woodwork/blob/main/woodwork/tests/schema/test_table_schema.py#L463-L467), it seems like we can't set logical types in the table schema, but we can using the table accessor. Is there a reason for this?

```python
    error_message = ("Logical Types must be of the LogicalType class "
                     "and registered in Woodwork's type system. "
                     "Double does not meet that criteria.")
    with pytest.raises(TypeError, match=error_message):
        schema.set_types(logical_types={'id': 'Double'})
```

jeff-hernandez:
We can use `issubclass` [here](https://github.com/alteryx/woodwork/blob/main/woodwork/table_schema.py#L436-L437) to check if the selector class is a logical type class, but this would require additional changes to the current tests.

```python
            elif maybe_ltype_class == ww.logical_types.LogicalType.__class__:
                raise TypeError(f"Specified LogicalType selector {maybe_ltype} is not registered in Woodwork's type system.")
```

thehomebrewnerd:
> Based on [this test](https://github.com/alteryx/woodwork/blob/main/woodwork/tests/schema/test_table_schema.py#L463-L467), it seems like we can't set logical types using strings in the table schema, but we can in the table accessor. Is there a reason for this?
> 
> ```python
>     error_message = ("Logical Types must be of the LogicalType class "
>                      "and registered in Woodwork's type system. "
>                      "Double does not meet that criteria.")
>     with pytest.raises(TypeError, match=error_message):
>         schema.set_types(logical_types={'id': 'Double'})
> ```

@jeff-hernandez The ability to use strings is convenience for users, and we really don't ever want people to change a logical type directly through the schema as this would potentially bypass any dtype conversion that would need to happen in order to keep the data dtype in sync with the dtype specified in the LogicalType.

Since users shouldn't be setting logical types through the schema, I don't think we thought there was a need to allow string inputs.

jeff-hernandez:
> ... we really don't ever want people to change a logical type directly through the schema ..

@thehomebrewnerd  It seems like the schema is set as a private attribute to the table accessor `DataFrame.ww._schema` but the user can access a copy of the schema using `DataFrame.ww.schema`. What would happen if this method is used from the copy of the schema `DataFrame.ww.schema.set_types`. As an alternative approach, couldn't we make it a private method in the original schema `DataFrame.ww.schema._set_types`?

thehomebrewnerd:
> > ... we really don't ever want people to change a logical type directly through the schema ..
> 
> @thehomebrewnerd It seems like the schema is set as a private attribute to the table accessor `DataFrame.ww._schema` but the user can access a copy of the schema using `DataFrame.ww.schema`. What would happen if this public method is used from the copy of the schema `DataFrame.ww.schema.set_types`? As an alternative approach, couldn't we make it a private method in the original schema `DataFrame.ww.schema._set_types`?

I don't think anything would happen if you called `set_types` on the copy of the schema that is returned aside from updating that schema copy. It shouldn't have any impact on the schema that is used by the woodwork dataframe. That was by design to prevent someone from making schema changes accidentally that caused the schema to get invalidated.

It does feel like `_set_types` should be a private method to me (maybe that should be a separate PR?). @tamargrey Am I overlooking a scenario that would require users to call `set_types` on the schema directly?

tamargrey:
@thehomebrewnerd 
> It does feel like _set_types should be a private method to me (maybe that should be a separate PR?). @tamargrey Am I overlooking a scenario that would require users to call set_types on the schema directly?

No, but I think the idea with leaving a lot of the methods public on the Schemas is that, in theory, someone could use them for another typing-related purpose (kind of like how we decided to use ColumnSchema for featuretools primitives!). Then again, if we're saying that users might want to use TableSchema or ColumnSchema for something, maybe they should have the convenience of using strings.

tamargrey:
> We can use issubclass here to check if the selector class is a logical type class, but this would require additional changes to the current tests.


@jeff-hernandez is this related to trying to get rid of the use of `_get_ltype_class` in `maybe_ltype_class = _get_ltype_class(maybe_ltype)`?

I forgot that we used it this way, and I think its use here is probably still relevant. If it complicates things a ton to completely remove `_get_ltype_class`, then we can keep the call to it on line 429 or do something similar inline.

jeff-hernandez:
> No, but I think the idea with leaving a lot of the methods public on the Schemas is that, in theory, someone could use them for another typing-related purpose (kind of like how we decided to use ColumnSchema for featuretools primitives!). Then again, if we're saying that users might want to use TableSchema or ColumnSchema for something, maybe they should have the convenience of using strings.

@thehomebrewnerd @tamargrey the logic to parse a string and return a logical type is separate from actually changing the schema and it can also be a convenience for developers. I would still consider making it a private method since we aren't necessarily sure how users can find that method helpful.

jeff-hernandez:
> @jeff-hernandez is this related to trying to get rid of the use of `_get_ltype_class` in `maybe_ltype_class = _get_ltype_class(maybe_ltype)`?
> 
> I forgot that we used it this way, and I think its use here is probably still relevant. If it complicates things a ton to completely remove `_get_ltype_class`, then we can keep the call to it on line 429 or do something similar inline.

Yes, it’s related. It seems like `_get_ltype_class` is closely coupled to the validation and test logic. If we remove this function, it seems like we'll need to make additional changes to the current tests.

woodwork/column_accessor.py
jeff-hernandez: Removed `_get_ltype_class` wherever it didn't break too many tests. This results in the behavior where `DataFrame.ww._schema.set_types` will not instantiate logical types.

tamargrey (approved):
lgtm!

PR #916, opened 2021-05-13
Add option to return schema to accessor select

### LINKED ISSUE #883: Ability to query schema without creating a new dataframe
- As a [user/developer], I wish I could use Woodwork to filter the table schema based on logical types and semantic tags without creating a new dataframe. Users currently can use `select` and then access the metadata they need but it might be overkill to create an entirely new dataframe for some use cases. For example, I might want to check if there are any `Categorical` columns so I can do a one-hot encoding. I don't need to create a new dataframe in that case just to determine if there are `Categorical` columns.

#### Code Example

I think we can add a `select` method to the `schema` or add a `return_schema` flag to the existing `select` method.

```python
schema_with_properties = df.ww.schema.select(include=["Integer", "Categorical"])
schema_with_properties = df.ww.select(include=["Integer", "Categorical"], return_schema=True)
```

thehomebrewnerd:
- Add option to return schema to accessor select
- Closes #883

This PR adds a `return_schema` boolean argument to the table accessor select method that returns a TableSchema object for the column matches instead of a DataFrame object.

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -2,6 +2,7 @@
 
 Release Notes
 -------------
+
 Future Release
 ==============
     * Enhancements
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -3,7 +3,7 @@
 Release Notes
 -------------
 Future Release
-  ==============
+==============
     * Enhancements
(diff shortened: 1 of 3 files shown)


tamargrey (commented):
lgtm!

tamargrey (approved):

PR #932, opened 2021-05-16
Add concat_columns function to combine woodwork objects

### LINKED ISSUE #884: Ability to concat dataframes through woodwork
- As a [user/developer], I wish I could use Woodwork to concat multiple dataframes through the woodwork accessor and preserve the schema. Users can currently do this by looping over the columns in a new dataframe and doing `df.ww["col_name"] = col` but it would be much better to avoid having to write the loop each time.

I think we should prioritize concatting along `axis 1`, i.e. adding new columns as opposed to rows. 

#### Code Example

```python
new_dataframe = ww.concat([original_dataframe, new_df])
new_dataframe = ww.concat([original_dataframe, new_df, df_2])
```

The two tricky cases I can imagine right now are dataframes with different primary keys/time indices. I think it's ok to error out if the indices don't match. This might be better than what `pandas.concat` does which is will fill in NaNs where the indices don't overlap I believe.

tamargrey:
- Adds `ww.concat` that follows the pandas `pd.concat` format to allow combining the data and typing information from multiple Woodwork dataframes and series
- Contains the `join` and `axis` parameters
- Closes #884


woodwork/utils.py
```diff
@@ -298,3 +299,149 @@ def _parse_logical_type(logical_type, name):
...
+        object, type of objs: When objs contains at least one DataFrame, a Woodwork DataFrame is returned.
+            When concatenating along the columns (axis=1), a Woodwork DataFrame is returned.
+            When concatenating along the index (axis=1) with only series, a series is returned
+            without Woodwork initialized.
+    """
+    if not objs:
+        raise ValueError('No objects to concatenate')
+
+    # --> consider adding sort
+    # --> consider adding ignore index
```
tamargrey, line 330: I'm thinking it doesn't necessarily make sense to add `ignore_index` in as a parameter, because we'll need to determine how to handle the case where a user wants to ignore the index but there is a Woodwork index. We need the underlying index of the concatenated dataframe to match the Woodwork index column exactly (in pandas only).

Our options would be to either
1. Only allow ignore index if there's no Woodwork index set
2. After concatenation, update the underlying index to match whatever the index column is

Adding `ignore_index` is only possible for Koalas and Pandas, so we'll need to make separate concat calls for each library, which increases complexity.

woodwork/utils.py
```diff
@@ -298,3 +299,149 @@ def _parse_logical_type(logical_type, name):
...
+    Returns:
+        object, type of objs: When objs contains at least one DataFrame, a Woodwork DataFrame is returned.
+            When concatenating along the columns (axis=1), a Woodwork DataFrame is returned.
+            When concatenating along the index (axis=1) with only series, a series is returned
+            without Woodwork initialized.
+    """
+    if not objs:
+        raise ValueError('No objects to concatenate')
+
+    # --> consider adding sort
```
tamargrey, line 329: Adding `sort` as a parameter for Pandas DataFrames (and possibly Koalas, since the parameter is available there as well) means we might want to take into account `already_sorted` at woodwork init. 

I think it might make sense to revisit this in a separate issue, and we can add it if it's requested.

woodwork/utils.py
```diff
@@ -298,3 +299,149 @@ def _parse_logical_type(logical_type, name):
...
+                    elif obj_idx != table_idx:
+                        raise IndexError(f'Cannot concat dataframes with different Woodwork {idx_type} columns.')
+                    elif axis in {1, 'columns'} and obj_idx == table_idx:
+                        # Do not include the column from the dataframe to avoid duplicates columns
+                        drop_cols.append(table_idx)
+
+            # To avoid changing input objects, create a new DataFrame without the index columns
+            if drop_cols:
+                obj_to_add = obj.ww.drop(columns=drop_cols)
+            ww_columns = obj_to_add.ww.schema.columns
```
tamargrey, line 380: Wanted to bring attention to this behavior here: In order to avoid having duplicate index and time index columns in the concatenated DataFrame, we need to only have these columns present once. We're keeping the first occurrence of them. 

It's not ideal to have to copy the data here, but the other option was the remove the column inplace with `pop`, but I didn't want to alter the input data. Happy to explore something different, if there's an option I missed.

woodwork/utils.py
```diff
@@ -298,3 +299,149 @@ def _parse_logical_type(logical_type, name):
...
+
+    concat_objs = []
+    # Record the typing information for all the columns that have Woodwork schemas
+    col_names_seen = set()
+    for obj in objs:
+        ww_columns = {}
+        obj_to_add = obj
+        if obj is None:
+            # --> not confident that this is the right choice. We need to be able to use the elements of objs to get the library type so it's hard if any element can be None
+            raise ValueError('Cannont include None in list of objects to concatenate')
```
tamargrey, line 350: Only Pandas allows `None` as inputs to concat. It just ignores it. But Because we rely on checking the type of an object in `objs` to know the library to use for concat, having possible `None` values in `objs` means we'll need a more complex check to determine the library, and since two of them don't allow None as input, I figured we could just disallow it.

woodwork/utils.py
```diff
@@ -298,3 +299,149 @@ def _parse_logical_type(logical_type, name):
...
+    elif dd and isinstance(obj, (dd.Series, dd.DataFrame)):
+        lib = dd
+
+    combined_df = lib.concat(concat_objs, axis=axis, join=join)
+
+    # Cannot initialize Woodwork on series output
+    # --> we can handle if we really want to we'll just need to account for this case when collecting typging info
+    if not isinstance(combined_df, lib.DataFrame):
+        warnings.warn('Woodwork not initialized on the output of ww.concat', ww.exceptions.WoodworkNotInitWarning)
+        return combined_df
```
tamargrey, line 416: The situation where a user concatenates multiple series along `axis=0` is the only way we get a series as output from concat, and we can initialize Woodwork by choosing a similar approach where we only use the first column's typing information, but it would add enough further complexity here, that I figured it might make sense to wait until we have a solid use case for it.

woodwork/utils.py
```diff
@@ -298,3 +299,149 @@ def _parse_logical_type(logical_type, name):
...
+            index = None
+
+        def filter_cols(typing_dict):
+            return {name: typing_val for name, typing_val in typing_dict.items() if name in combined_cols}
+
+        logical_types = filter_cols(logical_types)
+        semantic_tags = filter_cols(semantic_tags)
+        col_metadata = filter_cols(col_metadata)
+        col_descriptions = filter_cols(col_descriptions)
+        use_standard_tags = filter_cols(use_standard_tags)
```
tamargrey, line 433: I'm considering pulling all of this logic into a single for-loop in order to not have 5 separate `filter_cols` calls that each can loop over the number of columns in the resulting dataframe.

In general, this `concat` function feels very messy to me. All of the disparate typing variables make it hard to pull logic into separate functions without having to return a length-7 tuple, and it's hard to separate the portions of this code that are changing data vs compiling the typing information.

woodwork/tests/test_utils.py
```diff
@@ -566,3 +569,452 @@ def test_parse_logical_type_errors():
...
+    assert list(combined_df1.columns) == ['id', 'signup_date', 'age']
+    assert list(combined_df2.columns) == ['id', 'age', 'signup_date']
+
+
+def test_concat_cols_all_series(sample_df):
+    age_series = ww.init_series(sample_df['age'], logical_type='Double')
+    combined_df = concat([sample_df['id'], age_series, sample_df['is_registered']], axis=1)
+
+    assert isinstance(combined_df.ww.schema, ww.table_schema.TableSchema)
+    assert set(combined_df.columns) == {'id', 'age', 'is_registered'}
```
thehomebrewnerd, line 995: Think we should test the column order here like we did in the test immediately preceding this?

woodwork/tests/test_utils.py
```diff
@@ -566,3 +569,452 @@ def test_parse_logical_type_errors():
...
+    assert combined_df.ww.logical_types['age'] == Double
+
+
+def test_concat_rows_all_series(sample_df):
+    warning = 'Woodwork not initialized on the output of ww.concat'
+    with pytest.warns(ww.exceptions.WoodworkNotInitWarning, match=warning):
+        combined_series = concat([sample_df['age'], sample_df['age']], axis=0)
+
+    assert combined_series.ww.schema is None
+    assert combined_series.name == 'age'
```
thehomebrewnerd, line 1005: Why can't we initialize Woodwork in this scenario?
```python
combined_series = concat([sample_df['age'], sample_df['age']], axis=0)
combined_series.ww.init()
combined_series.ww
```
```
<Series: age (Physical Type = Int64) (Logical Type = IntegerNullable) (Semantic Tags = {'numeric'})>
```

woodwork/utils.py
thehomebrewnerd: My vote: delay this for now

woodwork/utils.py
thehomebrewnerd: I also agree that we should revisit this separately if needed

woodwork/utils.py
```diff
@@ -298,3 +299,149 @@ def _parse_logical_type(logical_type, name):
...
         raise TypeError(f"Invalid logical type specified for '{name}'")
+
+
+def concat(objs, axis=1, join='outer', validate_schema=True):
+    """
+    Concatenate Woodwork objects along a particular axis. Cannot concatenate DataFrames
+    with conflicting Woodwork index or time index columns. If DataFrames have the same Woodwork
+    index or time index columns, the first appearance of the column will be included in the
+    concatenated DataFrame. As Woodwork does not allow duplicate column names,
+    will not allow duplicate columns at concatenation.
```
thehomebrewnerd, line 310: I'm not sure how exactly best to do this, but I'm thinking that some of the behaviors between our concat and pandas concat are different (for example how we deal with a set ww index that is duplicated). It might be good for us to show some typical use cases to show how this concat works, similar to how pandas has several examples in their [documentation](https://pandas.pydata.org/docs/reference/api/pandas.concat.html). 

Maybe for now we could add some examples to our documentation page for this function, like pandas has done, unless you have another idea.

woodwork/utils.py
thehomebrewnerd: I'm not sure we should add this check at all. The code for checking the library defaults to pandas, and it's probably rare that someone would pass weird values `None` or `3` instead of a series or dataframe.

woodwork/utils.py
```diff
@@ -298,3 +299,149 @@ def _parse_logical_type(logical_type, name):
...
+
+    # Perform concatenation with the correct library
+    dd = import_or_none('dask.dataframe')
+    ks = import_or_none('databricks.koalas')
+
+    lib = pd
+    if ks and isinstance(obj, (ks.Series, ks.DataFrame)):
+        lib = ks
+    elif dd and isinstance(obj, (dd.Series, dd.DataFrame)):
+        lib = dd
```
thehomebrewnerd, line 408: I think these checks rely on the last value of `obj` that was set in the for loop above, and will check the library of the last value in the list. Might be safer to use `objs[0]` instead of `obj` here, which will of course check the first value instead.

woodwork/utils.py
thehomebrewnerd: Made a related comment about this in a test. Not sure if this would be a common use case or not. In any case, if we leave as is, I would just make sure the associated comments indicate that we are not supporting this currently, not so much that we cannot support it.

thehomebrewnerd (commented):
Generally don't have much technical feedback on the implementation at this stage, but leve some feedback with questions/thoughts about what we should/should not support at this time.

Might need to have a more in depth discussion at some point to resolve those points and make a decision.

woodwork/utils.py
tamargrey: It won't get to the concat call, because the check for if `isinstance(obj.ww.schema, ww.table_schema.TableSchema)` will raise some sort of `NoneType has no attribute ww` error. Then again, if we're not supporting None, maybe that's a clear enough error.

woodwork/tests/test_utils.py
```diff
@@ -566,3 +569,452 @@ def test_parse_logical_type_errors():
...
+    df2 = sample_df.ww[['phone_number', 'age', 'signup_date', 'is_registered']]
+
+    combined_df = concat([df1, df2])
+
+    assert combined_df.ww.semantic_tags['id'] == {'numeric'}
+    assert combined_df.ww.semantic_tags['full_name'] == set()
+    assert combined_df.ww.semantic_tags['age'] == set()
+
+
+def test_concat_rows_different_schema(sample_df):
```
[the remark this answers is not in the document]
tamargrey: It sounds like one way to simplify things is to be more restrictive about what inputs we allow. Ideally we wouldn't have to be so restrictive with axis=0 inputs that we'd require the table schemas to match exactly, but I do think it greatly simplifies things to require that table schemas are equal, though I'd suggest we do a shallow check because different metadata shouldn't stop us from being able to concat. 

I think that would also require all the DataFrames to have Woodwork initialized. Which is another requirement that would simplify things

woodwork/tests/test_utils.py
```diff
@@ -566,3 +569,452 @@ def test_parse_logical_type_errors():
...
+
+    combined_df_2 = concat([df2, df1], axis=0)
+    combined_df_2.ww.logical_types['age'] == IntegerNullable
+
+    combined_df_3 = concat([df3, df2], axis=0)
+    combined_df_3.ww.logical_types['age'] == NaturalLanguage
+    assert str(combined_df_3['age'].dtype) == 'string'
+
+
+def test_concat_cols_different_schema(sample_df):
```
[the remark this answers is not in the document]
tamargrey: To raise an error, we could store the index and time index logical types along with the name of the index and confirm that the logical types are all the same. However, this would add complexity.

To simplify, I think I'd suggest that we only allow one dataframe to have an index column set. That way, users can handle how they want to drop down to one index column instead of us handling taking just the first index column and dropping all the others.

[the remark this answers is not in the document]
tamargrey: Maybe the result of Woodwork's concat shouldn't be any different than the result of a plain concat call. Just that we have additional requirements about no duplicate columns and that we will try to update the dtypes to keep the schema valid.

woodwork/tests/test_utils.py
```diff
@@ -566,3 +569,452 @@ def test_parse_logical_type_errors():
...
+
+    combined_df_2 = concat([df2.ww.copy(), df1.ww.copy()], axis=1)
+    combined_df_2.ww.logical_types['id'] == IntegerNullable
+
+
+def test_concat_rows_different_columns(sample_df):
+    sample_df.ww.init(time_index='signup_date')
+    df1 = sample_df.ww[['id', 'signup_date', 'full_name', 'email']]
+    df2 = sample_df.ww[['id', 'signup_date', 'phone_number', 'age', 'is_registered']]
+    # should work if dtypes dont change
```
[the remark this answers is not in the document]
tamargrey: My mentality had mainly been that we let Woodwork initialize on the results of concat and if it fails, we let to woodwork initialization failure come through. So we'd get `TypeConversionError`s but it's also how we catch the duplicate column errors or any time the index column is suddenly not unique. It helps keep additional complexity out of the concat function.

I think if we try and stick with the comment above that the result of Woodwork's concat shouldn't make too many changes to the structure of the resulting dataframe (mostly coming from the handling of index and time index columns), then it could be less confusing about why woodwork initialization fails.

woodwork/tests/test_utils.py
```diff
@@ -566,3 +569,452 @@ def test_parse_logical_type_errors():
...
+
+def test_concat_combine_metadatas(sample_df):
+    df1 = sample_df[['id', 'full_name', 'email']]
+    df1.ww.init(table_metadata={'created_by': 'user0', 'test_key': 'test_val1'}, column_metadata={'id': {'interesting_values': [1, 2]}})
+    df2 = sample_df[['phone_number', 'age', 'signup_date', 'is_registered']]
+    df2.ww.init(table_metadata={'table_type': 'single', 'test_key': 'test_val2'}, column_metadata={'age': {'interesting_values': [33]}})
+
+    combined_df_1 = concat([df1, df2])
+
+    # The first table sets the metadata when there's overlap
```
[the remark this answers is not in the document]
tamargrey: raising an error here shouldn't be too hard. I'd be fine with saying that there can't be overlap and if there is the user has to handle that.

woodwork/tests/test_utils.py
tamargrey: We could, just didn't add support for it, so happy to clarify in the warning that it's not supported rather than making it seem like it's not possible.

woodwork/utils.py
tamargrey: No longer relevant with behavior change to only allow one index and time index to be specified across the input objects.

woodwork/utils.py
tamargrey: Removed the specific None check for simplicity

woodwork/utils.py
tamargrey: Situation no longer possible now that we restrict `axis=1`

woodwork/utils.py
tamargrey: @thehomebrewnerd made the simplification changes (fixing `axis=1`, `join=outer`, and only allowing one index and time index colum to be specified across the inputs) along with some other suggestions that have simplified this util function. 

Let me know what you think!

woodwork/tests/test_utils.py
```diff
@@ -566,3 +569,452 @@ def test_parse_logical_type_errors():
...
+    assert len(set(combined_df.columns)) == len(set(sample_df.columns))
+
+    # confirm that the columns that are "lost" in the resulting df are still in the original dataframes
+    assert 'id' in df1.columns
+    assert 'signup_date' in df1.columns
+    assert 'id' in df2.columns
+    assert 'signup_date' in df2.columns
+
+
+def test_concat_rows_with_ww_indexes(sample_df):
```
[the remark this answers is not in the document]
tamargrey: No longer allowing row concatenation, so I think this situation is no longer possible.

woodwork/tests/test_utils.py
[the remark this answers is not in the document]
tamargrey: No longer allowing `axis=0`, so no longer a problem for this PR.

woodwork/tests/test_utils.py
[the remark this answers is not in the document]
tamargrey: No longer allowing multiple index columns to be specified, so we will not run into conflicting logical types that could result in this situation.

woodwork/tests/test_utils.py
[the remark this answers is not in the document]
tamargrey: Added an error if there's overlapping keys between table metadata: `Cannot resolve overlapping keys in table metadata:`

woodwork/tests/test_utils.py
```diff
@@ -566,3 +569,452 @@ def test_parse_logical_type_errors():
...
+    concat([sample_df, copy_df], axis=0, validate_schema=False)
+
+    assert not mock_validate_accessor_params.called
+
+    concat([sample_df, copy_df], axis=0, validate_schema=True)
+
+    assert mock_validate_accessor_params.called
+
+
+def test_concat_cols_mismatched_index_adds_nans(sample_df):
```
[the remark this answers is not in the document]
tamargrey: Added a test that does add multiple nans to the index column, causing an error to be raised for pandas dataframes because the index is no longer unique. It's still just being caught by woodwork init, though.

woodwork/tests/test_utils.py
```diff
@@ -566,3 +569,452 @@ def test_parse_logical_type_errors():
...
+    else:
+        df2['id'] = pd.Series([4, 5, 6, 7], dtype='int64')
+
+    axis_0 = concat([df1, df2], axis=0)
+    axis_index = concat([df1, df2], axis='index')
+    assert axis_0.ww == axis_index.ww
+    assert to_pandas(axis_0).equals(to_pandas(axis_index))
+
+
+def test_concat_different_table_types(sample_df_pandas, sample_df_dask):
```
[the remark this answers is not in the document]
tamargrey: Removed this test

woodwork/tests/test_utils.py
tamargrey: I think `test_concat_table_order` is technically testing the relative column order for `axis=1` concatenation of multiple dataframes, and we can add column order when concatenating just series. 

But I think the equivalent of the original column order for row concatenation is to check concatenating columns where the indexes are the same but in different orders. I'll add a test for that

woodwork/utils.py
tamargrey: updated to use `objs[0]`

woodwork/tests/test_utils.py
tamargrey: Situation no longer relevant

woodwork/tests/test_utils.py
```diff
@@ -566,3 +569,452 @@ def test_parse_logical_type_errors():
...
+    assert df1.ww.schema is None
+    assert df2.ww.schema is None
+    assert combined_df.ww.schema is not None
+
+    # Though the input dataframes don't have Woodwork initalized,
+    # the concatenated DataFrame has Woodwork initialized for all the columns
+    assert combined_df.ww.logical_types['id'] == Integer
+
+    df1.ww.init()
+    df2.ww.init()
```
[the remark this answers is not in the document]
tamargrey: yep, for dtype conversions so we can get the pd.concat call to, in theory, produce the same results

woodwork/tests/test_utils.py
```diff
@@ -566,3 +569,452 @@ def test_parse_logical_type_errors():
...
+    df2 = sample_df[['phone_number', 'age', 'signup_date', 'is_registered']]
+
+    combined_df = concat([df1, df2])
+    assert df1.ww.schema is None
+    assert df2.ww.schema is None
+    assert combined_df.ww.schema is not None
+
+    # Though the input dataframes don't have Woodwork initalized,
+    # the concatenated DataFrame has Woodwork initialized for all the columns
+    assert combined_df.ww.logical_types['id'] == Integer
```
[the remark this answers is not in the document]
tamargrey: good call. chagning

woodwork/tests/test_utils.py
```diff
@@ -566,3 +569,452 @@ def test_parse_logical_type_errors():
...
+    assert to_pandas(combined_df).equals(pandas_combined_df)
+
+
+def test_concat_cols_with_series(sample_df):
+    df = sample_df[['id', 'full_name', 'email', 'phone_number', 'age']]
+    s1 = sample_df['signup_date']
+    sample_df.ww.init()
+    s2 = sample_df['is_registered']
+
+    combined_df = concat([df, s1, s2])
```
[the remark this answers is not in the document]
tamargrey: Added a check that `combined_df`'s schema is equal to `sample_df`'s

woodwork/tests/test_utils.py
```diff
@@ -566,3 +569,452 @@ def test_parse_logical_type_errors():
...
+    assert sample_df.ww.semantic_tags['age'] == set()
+
+    df1 = sample_df.ww[['id', 'full_name', 'email']]
+    df2 = sample_df.ww[['phone_number', 'age', 'signup_date', 'is_registered']]
+
+    combined_df = concat([df1, df2])
+
+    assert combined_df.ww.semantic_tags['id'] == {'numeric'}
+    assert combined_df.ww.semantic_tags['full_name'] == set()
+    assert combined_df.ww.semantic_tags['age'] == set()
```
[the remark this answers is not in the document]
tamargrey: added checks for the logical type change and all the use standard tags values set at init

woodwork/utils.py
tamargrey: with the changes, the main difference is how we restrict to just allowing concatenation of columns and that we don't allow duplicate columns. 

I think these might be a little harder to showcase in examples, so it might be better suited for a guide or somewhere where we have more space to explain. Not totally sure though

woodwork/utils.py
```diff
@@ -298,3 +298,112 @@ def _parse_logical_type(logical_type, name):
...
+                                                 ('time_index', obj.ww.time_index, time_index)]:
+                if obj_idx is not None:
+                    if table_idx is None:
+                        if idx_type == 'index':
+                            index = obj_idx
+                        elif idx_type == 'time_index':
+                            time_index = obj_idx
+                    else:
+                        raise IndexError(f'Cannot have {idx_type} columns set on more than one table in objs. '
+                                         f'Please remove the {idx_type} columns from all but one table.')
```
thehomebrewnerd, line 362: While I don't see any problems with this code block I find it pretty hard to follow with all the nested if conditions inside the for loop. Is there a way we could refactor this to make it more readable for the future? Maybe check index and time index settings separately instead of using the same set of code?

woodwork/utils.py
```diff
@@ -298,3 +298,112 @@ def _parse_logical_type(logical_type, name):
...
+            ww_columns = {obj.name: obj.ww.schema}
+
+        concat_objs.append(obj)
+
+        # Compile the typing information per column
+        for name, col_schema in ww_columns.items():
+            # Do not look at typing information for duplicate columns.
+            # Means that index columns will only have first occurrance's typing information
+            if name in col_names_seen:
+                continue
```
thehomebrewnerd, line 375: If I understand right, if we have duplicate column names, we keep the first one we encounter and ignore the rest. I'm wondering if it would be better to raise an error instead of continue here so the user has an opportunity to resolve.

woodwork/tests/test_utils.py
```diff
@@ -566,3 +568,315 @@ def test_parse_logical_type_errors():
...
+
+    new_df2 = df2.ww.drop(['id', 'signup_date'])
+
+    # Because underlying index is set, this won't change concat operation
+    pd.testing.assert_index_equal(to_pandas(df1.index), to_pandas(new_df2.index))
+
+    combined_df = concat_columns([df1, new_df2])
+    assert combined_df.ww.index == 'id'
+    assert combined_df.ww.time_index == 'signup_date'
+    assert len(set(combined_df.columns)) == len(set(sample_df.columns))
```
thehomebrewnerd, line 694: Is the conversion to `set` here needed for anything?

woodwork/utils.py
thehomebrewnerd: Upon further review, it seems the in the case of duplicate column names we get an error below when we try to init Woodwork. Am I understanding correctly? If so, I wonder if it would be better to go ahead and raise that error here instead of waiting for init? Also, I think the comment above is no longer valid if we error when duplicate cols are present.

thehomebrewnerd (commented):
This seems a fair amount cleaner and easier to understand. Just a few more comments/questions.

Also, since the concat test take up so much of the `test_utils.py` file, do you think it would be worth reorganizing a bit - creating a utils folder in the test folder and then splitting the concat tests out of `test_utils.py` into a separate `test_concat_columns.py` file and the putting both in the new utils folder?

woodwork/utils.py
tamargrey: So the comment above these lines isn't relevant no matter what. But I'm not opposed to just catching the duplicate columns there. Plus, and this just occured to me, we can catch duplicate index columns under the general duplicate column check, which would greatly simplify things, but would mean we'd lose some of the specificity in the error message. Though, the instructions are the same whether or not the column is an index.

An idea for an error message:

> Duplicate column 'id' has been found in more than one input object. Please remove duplicate columns from all but one table.

tamargrey:
> do you think it would be worth reorganizing a bit
@thehomebrewnerd yes, very happy to reorganize the utils tests so that concat_columns tests can have their own file.

woodwork/utils.py
thehomebrewnerd: I think the loss of specificity in the error message is worth it to simplify the code. I also think the error message you have suggested sounds good!

tamargrey: I might have been overly optimistic here. We still need to be able to catch the case where the index columns might have different names, so some of the logic will have to stay, but we can catch that as a separate case, and I'll separate the index and time index logic os it's hopefully less confusing

tamargrey: @thehomebrewnerd Made these changes. Let me know if the duplicate column logic and index logic make sense to you!

woodwork/tests/test_utils.py
tamargrey: when we were allowing multiple index columns across dataframes, I the idea was to confirm that we had ended up without any of the duplicate columns (though this would have been caught by woodwork init), but since that behavior has changed, it's probably even less necessary

woodwork/utils.py
thehomebrewnerd: I'm thinking examples probably aren't as needed with this updated version since there are fewer ways this function can be used now.

woodwork/utils.py
thehomebrewnerd: I like it! Much easier for my simple mind to follow!

woodwork/tests/utils/test_concat.py
```diff
@@ -0,0 +1,341 @@
+
```
thehomebrewnerd, line 1: Extra blank line

thehomebrewnerd (approved):
One super-small item, but looks good!

woodwork/tests/utils/test_concat.py
tamargrey: whoopse, removed!

PR #942, opened 2021-05-17
Add setter and getter methods for name

### LINKED ISSUE #918: Setting Schema attribute from Accessor doesn't update Schema
Our table and column Accessors allow users to access attributes and properties from `TableSchema` and `ColumnSchema` objects through the accessor, but we run into issues if a user tries to set those attributes from the Accessor.

```python
df = sample_df.copy()
df.ww.init()

assert df.ww.name is None

df.ww.name = 'name'

assert df.ww.name == 'name'
assert df.ww.schema.name is None
```

I first noticed this with `TableSchema.name` in performing a `drop` operation. The dataframe from above will produce the following situation because we pass the Schema's `name` attribute to the resulting DataFrame, not the Accessor's:

```python
assert df.ww.name == 'name'

dropped_df = df.ww.drop(['id'])
assert dropped_df.ww.name is None
assert dropped_df.ww.schema.name is None
```

This will be an issue with any attribute of a Column or Table Schema that does not have a `getter` or `setter` on the Accessor. 

We have not run into this with table metadata, because it's a mutable dictionary, so we can get that value and update it with no issues, but if someone were to try and reassign that value, we'd run into the same situation. 

My suggestion is to either add just properties to the Accessor, essentially telling users that if they want to reassign `name` or `metadata` they have to go through `._schema` (note that going through `.schema` won't work because it's a copy of the schema.)

```python
df = sample_df.copy()
df.ww.init()

assert df.ww.name is None

df.ww._schema.name = 'name'

assert df.ww.schema.name =='name'
assert df.ww.name == 'name'
```

The other option is to add both getters and setters so that users can perform `df.ww.name = 'new_name'`. I'm leaning towards this option.

tuethan1999:
- Add getter and setter methods to use `TableAccessor._schema`'s name and metadata as the name and metadata properties
- Change the ColumnAccessor._schema's description and metadata to use the underlying `ColumnSchema`'s description and metadata
- Closes #918

### DIFF
--- a/docs/source/api_reference.rst
+++ b/docs/source/api_reference.rst
@@ -19,8 +19,10 @@ WoodworkTableAccessor
     WoodworkTableAccessor.init
     WoodworkTableAccessor.loc
     WoodworkTableAccessor.logical_types
+    WoodworkTableAccessor.metadata
     WoodworkTableAccessor.mutual_information
     WoodworkTableAccessor.mutual_information_dict
+    WoodworkTableAccessor.name
     WoodworkTableAccessor.physical_types
     WoodworkTableAccessor.pop
     WoodworkTableAccessor.remove_semantic_tags
--- a/docs/source/api_reference.rst
+++ b/docs/source/api_reference.rst
@@ -74,9 +74,11 @@ TableSchema
     TableSchema.add_semantic_tags
     TableSchema.index
     TableSchema.logical_types
+    TableSchema.metadata
     TableSchema.rename
     TableSchema.remove_semantic_tags
     TableSchema.reset_semantic_tags
+    TableSchema.name
     TableSchema.semantic_tags
     TableSchema.set_index
     TableSchema.set_time_index
@@ -93,10 +95,12 @@ ColumnSchema
     :toctree: generated/
 
     ColumnSchema
+    ColumnSchema.description
     ColumnSchema.is_boolean
     ColumnSchema.is_categorical
     ColumnSchema.is_datetime
     ColumnSchema.is_numeric
+    ColumnSchema.metadata
     
 Serialization
 =============
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -8,7 +8,7 @@ Future Release
     * Enhancements
         * Add option to return ``TableSchema`` instead of ``DataFrame`` from table accessor ``select`` method (:pr:`916`)
     * Fixes
-        * Fix bug when setting table name through accessor (:pr:`942`)
+        * Fix bug when setting table name and metadata through accessor (:pr:`942`)
     * Changes
     * Documentation Changes
     * Testing Changes
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -8,6 +8,7 @@ Future Release
     * Enhancements
         * Add option to return ``TableSchema`` instead of ``DataFrame`` from table accessor ``select`` method (:pr:`916`)
     * Fixes
+        * Use ``TableSchema``'s name as ``WoodworkTableAccessor``'s name (:pr:`942`)
     * Changes
     * Documentation Changes
     * Testing Changes
@@ -15,7 +16,7 @@ Future Release
         * Create separate worksflows for each CI job (:pr:`919`)
 
     Thanks to the following people for contributing to this release:
-    :user:`gsheni`, :user:`thehomebrewnerd`
+    :user:`gsheni`, :user:`thehomebrewnerd`, :user:`tuethan1999`
 
 
 v0.3.1 May 12, 2021
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -8,7 +8,7 @@ Future Release
     * Enhancements
         * Add option to return ``TableSchema`` instead of ``DataFrame`` from table accessor ``select`` method (:pr:`916`)
     * Fixes
-        * Use ``TableSchema``'s name as ``WoodworkTableAccessor``'s name (:pr:`942`)
+        * Fix bug when setting table name through accessor (:pr:`942`)
     * Changes
     * Documentation Changes
     * Testing Changes
--- a/woodwork/column_schema.py
+++ b/woodwork/column_schema.py
@@ -83,7 +83,7 @@ class ColumnSchema(object):
     def description(self):
         """Description of the column"""
         return self._description
-    
+
     @description.setter
     def description(self, description):
         _validate_description(description)
@@ -93,7 +93,7 @@ class ColumnSchema(object):
     def metadata(self):
         """Metadata of the column"""
         return self._metadata
-    
+
     @metadata.setter
     def metadata(self, metadata):
         metadata = metadata or {}
--- a/woodwork/column_schema.py
+++ b/woodwork/column_schema.py
@@ -29,16 +29,12 @@ class ColumnSchema(object):
             metadata (dict[str -> json serializable], optional): Extra metadata provided by the user.
             validate (bool, optional): Whether to perform parameter validation. Defaults to True.
         """
-        if metadata is None:
-            metadata = {}
         self.metadata = metadata
+        self.description = description
 
         if validate:
             if logical_type is not None:
                 _validate_logical_type(logical_type)
-            _validate_description(description)
-            _validate_metadata(metadata)
-        self.description = description
         self.logical_type = logical_type
 
         self.use_standard_tags = use_standard_tags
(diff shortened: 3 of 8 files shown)


thehomebrewnerd (changes requested):
I think we also need to make this same update for the accessor metadata.

```python
df = sample_df.copy()
df.ww.init()

df.ww.metadata = {'new': 'metadata'}
assert df.ww.metadata == {'new': 'metadata'}  # This creates a new accessor attribute, and is not setting the schema metadata
assert df.ww.schema.metadata == {}  # <- This should be the new metadata not an empty dict
```

docs/source/release_notes.rst
```diff
@@ -8,14 +8,15 @@ Future Release
     * Enhancements
         * Add option to return ``TableSchema`` instead of ``DataFrame`` from table accessor ``select`` method (:pr:`916`)
     * Fixes
+        * Use ``TableSchema``'s name as ``WoodworkTableAccessor``'s name (:pr:`942`)
```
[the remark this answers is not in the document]
thehomebrewnerd: If we make an update for the metadata, this release notes entry should also reflect that.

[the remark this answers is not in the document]
tamargrey: I'd vote for adding a getter and setter for the table metadata in this PR as well and updating this release note to reflect nate's suggestion.

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -132,6 +132,36 @@ def test_accessor_schema_property(sample_df):
...
 
+def test_set_accessor_name(sample_df):
+    df = sample_df.copy()
+    error = re.escape("Woodwork not initialized for this DataFrame. Initialize by calling DataFrame.ww.init")
+    with pytest.raises(WoodworkNotInitError, match=error):
+        df.ww.name
+    with pytest.raises(WoodworkNotInitError, match=error):
+        df.ww.name = 'name'
+
+    df.ww.init()
```
tamargrey, line 143: Can we also add a test case where we init woodwork on a DataFrame with a name and then change it to something else?

woodwork/table_accessor.py
```diff
@@ -248,6 +248,20 @@ def _get_typing_info(self):
         typing_info = typing_info.loc[list(self._dataframe.columns), :]
         return typing_info
 
+    @property
+    def name(self):
+        """Name of the underlying schema"""
```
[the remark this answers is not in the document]
tuethan1999: done!

woodwork/table_accessor.py
```diff
@@ -248,6 +248,20 @@ def _get_typing_info(self):
...
+        if self._schema is None:
+            _raise_init_error()
+        return self._schema.name
+
+    @name.setter
+    def name(self, new_name):
+        """Sets the name of the underlying schema"""
```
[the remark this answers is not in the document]
tuethan1999: done!

woodwork/tests/accessor/test_table_accessor.py
tuethan1999: done!

woodwork/table_accessor.py
```diff
@@ -248,6 +248,34 @@ def _get_typing_info(self):
...
+
+    @name.setter
+    def name(self, new_name):
+        """Set name of the DataFrame"""
+        if self._schema is None:
+            _raise_init_error()
+        self._schema.name = new_name
+
+    @property
+    def metadata(self):
```
tamargrey, line 266: `name` and `metadata` should probably be added to the api reference for WoodworkTableAccessor

docs/source/release_notes.rst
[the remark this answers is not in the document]
tuethan1999: done!

tamargrey:
Just had a thought about how we validate `name` and `metadata` in the TableSchema at init. We confirm `name` is a string and that `metadata` is a dictionary, so I'm wondering if we want to add those kind of checks in at the setters.

Up until now we haven't even had setters for those properties, so we definitely weren't restricting what they could be set to, so maybe it's not worth it to add those checks in. But if we're getting formal enough to add in the getter and setter to the Accessor class, I suspect we want to be validating the inputs.

We do something like this for the Column Accessor's `description` and `metadata` where we have setters that validate the new values. We pull in the validation functions from the `column_schema` file, but validation happens on the accessor.

tuethan1999:
> Just had a thought about how we validate `name` and `metadata` in the TableSchema at init. We confirm `name` is a string and that `metadata` is a dictionary, so I'm wondering if we want to add those kind of checks in at the setters.
> 
> Up until now we haven't even had setters for those properties, so we definitely weren't restricting what they could be set to, so maybe it's not worth it to add those checks in. But if we're getting formal enough to add in the getter and setter to the Accessor class, I suspect we want to be validating the inputs.
> 
> We do something like this for the Column Accessor's `description` and `metadata` where we have setters that validate the new values. We pull in the validation functions from the `column_schema` file, but validation happens on the accessor.

@tamargrey 
I've added type validation similar to Column Accessor's `description` and `metadata`. I was wondering why we chose to validate the types in the Accessor as supposed to validating types in the Schema. I feel like the type checking in the Accessor is like type checking in the frontend

tamargrey:
@tuethan1999 
> I feel like the type checking in the Accessor is like type checking in the frontend
That's a pretty apt way if thinking of it. Honestly not super sure why. Maybe @thehomebrewnerd remembers if it was on purpose, but I think it'd make more sense if we had corresponding getters and setters on the Schemas that do the actual validation. And the Accessors would be calling them.

thehomebrewnerd:
> @tuethan1999
> 
> > I feel like the type checking in the Accessor is like type checking in the frontend
> 
> That's a pretty apt way if thinking of it. Honestly not super sure why. Maybe @thehomebrewnerd remembers if it was on purpose, but I think it'd make more sense if we had corresponding getters and setters on the Schemas that do the actual validation. And the Accessors would be calling them.

I don't remember that being an intentional design decision. I'm guessing it is because we took the approach of updating a schema attribute directly rather than creating a separate schema setter that does the validation. I don't have a strong opinion on which path we take here.

docs/source/api_reference.rst
```diff
@@ -19,8 +19,10 @@ WoodworkTableAccessor
     WoodworkTableAccessor.init
     WoodworkTableAccessor.loc
     WoodworkTableAccessor.logical_types
+    WoodworkTableAccessor.metadata
```
tamargrey, line 22: I believe we also need entries in the table and column schemas for the properties that were added.

woodwork/column_schema.py
```diff
@@ -29,16 +29,15 @@ def __init__(self,
             metadata (dict[str -> json serializable], optional): Extra metadata provided by the user.
             validate (bool, optional): Whether to perform parameter validation. Defaults to True.
         """
-        if metadata is None:
-            metadata = {}
-        self.metadata = metadata
```
tamargrey, line 34: good call moving this to after validation

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -132,6 +132,105 @@ def test_accessor_schema_property(sample_df):
...
+    df = sample_df.copy()
+    df.ww.init(name='name')
+    assert df.ww.name == 'name'
+    df.ww.name = 'new_name'
+    assert df.ww.schema.name == 'new_name'
+    assert df.ww.name == 'new_name'
+
+
+def test_name_error_on_init(sample_df):
+    err_msg = "TableSchema name must be a string"
```
thehomebrewnerd, line 160: It seems to me like this error message isn't generic enough. This could happen when setting a name for either the schema and the accessor (like you are doing here), so it could be confusing to reference `TableSchema`. Maybe we should just say `Name must be a string` instead?

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -132,6 +132,105 @@ def test_accessor_schema_property(sample_df):
...
+    error = re.escape("Woodwork not initialized for this DataFrame. Initialize by calling DataFrame.ww.init")
+    with pytest.raises(WoodworkNotInitError, match=error):
+        df.ww.metadata
+    with pytest.raises(WoodworkNotInitError, match=error):
+        df.ww.metadata = {'new': 'metadata'}
+
+    df.ww.init()
+    assert df.ww.metadata == {}
+    df.ww.metadata = {'new': 'metadata'}
+    df.ww.metadata
```
thehomebrewnerd, line 195: Delete?

thehomebrewnerd (commented):
Looking pretty good. Just a few relatively minor comments.

woodwork/column_schema.py
```diff
@@ -80,6 +79,25 @@ def _get_column_tags(self, semantic_tags, validate):
 
         return semantic_tags
 
+    @property
+    def description(self):
```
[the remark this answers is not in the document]
tuethan1999: Done!

woodwork/column_schema.py
```diff
@@ -80,6 +79,25 @@ def _get_column_tags(self, semantic_tags, validate):
...
+        return self._description
+    
+    @description.setter
+    def description(self, description):
+        _validate_description(description)
+        self._description = description
+
+    @property
+    def metadata(self):
```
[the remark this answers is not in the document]
tuethan1999: Done!

docs/source/api_reference.rst
tuethan1999: Done!

woodwork/tests/accessor/test_table_accessor.py
tuethan1999: Done!

thehomebrewnerd (approved):
LGTM!

tamargrey (approved):
looks good to me!

PR #943, opened 2021-05-20
Add progress callback to mutual information functions

### LINKED ISSUE #938: Add progress callback to mutual information
- As a user, I wish I could use Woodwork to get the progress of my mutual information calculation.
- The would be turn on via a parameter when running the mutual_info function, called `add_callback`

#### Code Example

```python
import pandas as pd
from woodwork.demo import load_retail

df = load_retail()
df.ww

# function to be called with incremental progress updates
def add_callback(update, overall_progress_percent, overall_time_elapsed):
    # update: percentage change (float between 0 and 100) in progress since last call 
    # overall_progress_percent: percentage (float between 0 and 100) of total computation completed 
    # overall_time_elapsed: total time in seconds that has elapsed since start of call
    pass 

df.ww.mutual_information(add_callback=add_callback)
```

thehomebrewnerd:
- Add progress callback to mutual information functions
- Closes #938

### DIFF
--- a/woodwork/statistics_utils.py
+++ b/woodwork/statistics_utils.py
@@ -178,11 +178,11 @@ def _get_mutual_information_dict(dataframe, num_bins=10, nrows=None, include_ind
             included as long as its LogicalType is valid for mutual information calculations.
             If False, the index column will not have mutual information calculated for it.
             Defaults to False.
-        callback (callable): function to be called with incremental updates.
-            Has the following parameters:
-            update: percentage change (float between 0 and 100) in progress since last call
-            progress_percent: percentage (float between 0 and 100) of total computation completed
-            time_elapsed: total time in seconds that has elapsed since start of call
+        callback (callable): function to be called with incremental updates. Has the following parameters:
+
+            - update: percentage change (float between 0 and 100) in progress since last call
+            - progress_percent: percentage (float between 0 and 100) of total computation completed
+            - time_elapsed: total time in seconds that has elapsed since start of call
 
     Returns:
         list(dict): A list containing dictionaries that have keys `column_1`,
@@ -213,7 +213,7 @@ def _get_mutual_information_dict(dataframe, num_bins=10, nrows=None, include_ind
         data = data.loc[:, not_null_cols]
 
     # Setup for progress callback and make initial call
-    # Assume 1 unit for preprocessing, n for replace nans, n for make categorical and (n*n+n/2) for main calculation loop
+    # Assume 1 unit for preprocessing, n for replace nans, n for make categorical and (n*n+n)/2 for main calculation loop
     n = len(data.columns)
     total_loops = 1 + 2 * n + (n * n + n) / 2
     current_progress = _update_progress(start_time, timer(), 1, 0, total_loops, callback)
--- a/woodwork/statistics_utils.py
+++ b/woodwork/statistics_utils.py
@@ -1,3 +1,5 @@
+from timeit import default_timer as timer
+
 import numpy as np
 import pandas as pd
 from sklearn.metrics.cluster import normalized_mutual_info_score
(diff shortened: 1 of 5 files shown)


woodwork/statistics_utils.py
```diff
@@ -210,18 +219,35 @@ def _get_mutual_information_dict(dataframe, num_bins=10, nrows=None, include_ind
...
+
+    def update_progress(start_time, increment, current, total, progress_callback):
+        new_progress = current + increment
+        if progress_callback is not None:
+            elapsed_time = timer() - start_time
+            progress_callback((increment / total) * 100, (new_progress / total) * 100, elapsed_time)
+
+        return new_progress
+
+    current_progress = update_progress(start_time, 0, 0, total_loops, progress_callback)
```
tamargrey, line 234: Depending on how long `_replace_nans_for_mutual_info` and `_make_categorical_for_mutual_info` take, this could be stuck at 0 for a while, right (or maybe it wouldn't get to the progress bar for a little while--I need to spend some more time understanding into how this works)?

I'm not convinced we need to include those as a part of the whole function's progress (plus I don't know how you'd indicate the two helper functions, percentage-wise), but if we're just logging the actual mutual information calculation, maybe we should be clear about that.

thehomebrewnerd: Yeah, I was thinking something similar. Maybe since those functions both loop over the columns, we can add those counts to the total expected loop count and update after each one is called. Still won't be perfect, but might be better - and would give a progress update sooner.

tamargrey: Yes, that would make sense to me! Is it possible to have separate sections of the progress bar? Like "data preparation" vs "mutual info calculations" that might let us be a little more precise?

gsheni: @tamargrey We will likely add stages name and stage progress in the future. So you get overall progress, and can see the current stage it is on.

thehomebrewnerd: Right now this is just implementing the ability to call a progress callback function. The progress bar update will come later in another PR - not sure what we will be able to do there yet, as I haven't looked into that yet.

woodwork/utils.py
```diff
@@ -298,3 +298,21 @@ def _parse_logical_type(logical_type, name):
...
+    If provided, the callback function should accept the following parameters:
+        update: percentage change (float between 0 and 100) in progress since last call
+        progress_percent: percentage (float between 0 and 100) of total computation completed
+        time_elapsed: total time in seconds that has elapsed since start of call"""
+    new_progress = current_progress + progress_increment
+    if callback_function is not None:
+        elapsed_time = current_time - start_time
+        callback_function((progress_increment / total) * 100, (new_progress / total) * 100, elapsed_time)
+
+    return new_progress
```
tamargrey, line 318: I'm not sure I understand why we still return `new_progress` even if `callback_function` is None. We don't seem to do anything with `current_progress` in the mutual info calculation other than feed it to the next call of _update_progress, so if there's no callback function, are we doing anything with this information that we're saving?

thehomebrewnerd: I'm not sure now if I had a state at one point where I needed it outside the `if` block, or if I just overlooked it. In any case, it think it is safe to move the update calculation and return statement inside the `if` block.

woodwork/statistics_utils.py
```diff
@@ -204,8 +212,17 @@ def _get_mutual_information_dict(dataframe, num_bins=10, nrows=None, include_ind
     if set(not_null_cols) != set(valid_columns):
         data = data.loc[:, not_null_cols]
 
+    # Setup for progress callback and make initial call
+    # Assume 1 unit for preprocessing, n for replace nans, n for make categorical and (n*n+n/2) for main calculation loop
+    n = len(data.columns)
+    total_loops = 1 + 2 * n + (n * n + n) / 2
```
angela97lin, line 218: I think the comment is different from this; is it (n * n + n) / 2 or (n * n + n/2)? Where should the division by 2 go?

thehomebrewnerd: @angela97lin Good catch. The code was correct, and the comment was wrong. I will fix.

woodwork/statistics_utils.py
```diff
@@ -176,13 +178,19 @@ def _get_mutual_information_dict(dataframe, num_bins=10, nrows=None, include_ind
             included as long as its LogicalType is valid for mutual information calculations.
             If False, the index column will not have mutual information calculated for it.
             Defaults to False.
+        callback (callable): function to be called with incremental updates.
```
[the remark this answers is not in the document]
thehomebrewnerd: Made an update so this looks a little better. Still does something a little strange with the line breaks, but at least the information for the three required parameters shows up better.

tamargrey (approved):
lgtm!

PR #949, opened 2021-05-25
Fix serialization bug for categorical types

### LINKED ISSUE #947: Serialization/deserialization does not preserve dtype of categories for Ordinal columns
When serializing and deserializing a table with an Ordinal column, the dtype of the order values is not maintained. This causes the new dataframe to fail an equality check with the original dataframe. The same problem is present with Categorical columns.

```python
>>> import pandas as pd
>>> import woodwork as ww
>>> from woodwork.deserialize import read_woodwork_table
>>> df = pd.DataFrame({
...     'id': [0, 1, 2],
...     'reason': [1, 2, 1],
... })
>>> df.ww.init(logical_types={'reason': ww.logical_types.Ordinal(order=[1, 2])}) 
>>> df.ww.to_disk('.')
>>> new_df = read_woodwork_table('.')
>>> df.equals(new_df)
False
>>> df['reason']
0    1
1    2
2    1
Name: reason, dtype: category
Categories (2, int64): [1, 2]
>>> new_df['reason']
0    1
1    2
2    1
Name: reason, dtype: category
Categories (2, object): [1, 2]
```

thehomebrewnerd:
- Fix serialization bug for categorical types
- Closes #947 

Fixes a bug in which deserialization did not properly restore the dtype used to store the values associated with a categorical column causing the original and deserialized dataframes to evaluate as unequal during equality checks.

### DIFF
--- a/woodwork/serialize.py
+++ b/woodwork/serialize.py
@@ -17,7 +17,7 @@ from woodwork.utils import _is_s3, _is_url, import_or_none
 dd = import_or_none('dask.dataframe')
 ks = import_or_none('databricks.koalas')
 
-SCHEMA_VERSION = '9.0.0'
+SCHEMA_VERSION = '10.0.0'
 FORMATS = ['csv', 'pickle', 'parquet']
 
 
@@ -48,7 +48,8 @@ def typing_info_to_dict(dataframe):
          'physical_type': {
              'type': str(dataframe[col_name].dtype),
              # Store categorical values so they can be recreated if they are modified during serialization
-             'cat_values': dataframe[col_name].dtype.categories.to_list() if str(dataframe[col_name].dtype) == 'category' else None
+             'cat_values': dataframe[col_name].dtype.categories.to_list() if str(dataframe[col_name].dtype) == 'category' else None,
+             'cat_dtype': str(dataframe[col_name].dtype.categories.dtype) if str(dataframe[col_name].dtype) == 'category' else None
          },
          'semantic_tags': sorted(list(col.semantic_tags)),
          'description': col.description,
--- a/woodwork/serialize.py
+++ b/woodwork/serialize.py
@@ -38,11 +38,9 @@ def typing_info_to_dict(dataframe):
         dataframe = dataframe.ww.categorize(columns=category_cols)
     ordered_columns = dataframe.columns
 
-    def get_physical_type_dict(column):
-        type_dict = {}
-        column_dtype = str(column.dtype)
-        type_dict['type'] = column_dtype
-        if column_dtype == 'category':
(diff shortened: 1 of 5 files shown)


woodwork/serialize.py
```diff
@@ -48,7 +48,8 @@ def typing_info_to_dict(dataframe):
          'physical_type': {
              'type': str(dataframe[col_name].dtype),
              # Store categorical values so they can be recreated if they are modified during serialization
-             'cat_values': dataframe[col_name].dtype.categories.to_list() if str(dataframe[col_name].dtype) == 'category' else None
+             'cat_values': dataframe[col_name].dtype.categories.to_list() if str(dataframe[col_name].dtype) == 'category' else None,
+             'cat_dtype': str(dataframe[col_name].dtype.categories.dtype) if str(dataframe[col_name].dtype) == 'category' else None
```
rwedge, line 52: could we omit these keys if the column is not categorical?

thehomebrewnerd: Yes - the keys aren't needed otherwise. I'll try to clean that up.

woodwork/serialize.py
```diff
@@ -37,6 +37,16 @@ def typing_info_to_dict(dataframe):
         category_cols = [colname for colname, col in dataframe.ww._schema.columns.items() if col.is_categorical]
         dataframe = dataframe.ww.categorize(columns=category_cols)
     ordered_columns = dataframe.columns
+
+    def get_physical_type_dict(column):
+        type_dict = {}
+        column_dtype = str(column.dtype)
+        type_dict['type'] = column_dtype
```
rwedge, line 44: Not very important but this could be:

```python
type_dict = {'type': str(column.dtype)}
```

or
```python
type_dict = {}
type_dict['type'] = str(column.dtype)
```

if we're using the `cat_values` and `cat_dtype` assignments as references for how complex a single can be

woodwork/serialize.py
```diff
@@ -37,6 +37,16 @@ def typing_info_to_dict(dataframe):
         category_cols = [colname for colname, col in dataframe.ww._schema.columns.items() if col.is_categorical]
         dataframe = dataframe.ww.categorize(columns=category_cols)
     ordered_columns = dataframe.columns
+
+    def get_physical_type_dict(column):
```
rwedge, line 41: do we want this to be private?

thehomebrewnerd: Sure. Updated.

woodwork/serialize.py
thehomebrewnerd: Cleaned this up a bit.

rwedge (approved):
looks good

PR #955, opened 2021-05-28
Only copy dataframe for writing to parquet if contains LatLong

gsheni:


thehomebrewnerd (approved):
Looks good provided tests pass!

PR #970, opened 2021-06-08
Add install workflow to test update checker 

gsheni:
- This PR tests that the update checker + other future add-ons can be properly installed, and used in WW.

### DIFF
--- a/.github/workflows/install_test.yml
+++ b/.github/workflows/install_test.yml
@@ -7,7 +7,7 @@ on:
 
 name: Install Test
 jobs:
-  install_ft_complete:
+  install_ww_complete:
     name: ${{ matrix.python_version }} install woodwork complete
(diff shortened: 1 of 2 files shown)


.github/workflows/install_test.yml
```diff
@@ -0,0 +1,35 @@
+on:
+  pull_request:
+    types: [opened, synchronize]
+  push:
+    branches:
+      - main
+
+name: Install Test
+jobs:
+  install_ft_complete:
```
jeff-hernandez, line 10: Name contains `ft`

jeff-hernandez (approved):
Looks good! Just needs a minor update

.github/workflows/install_test.yml
gsheni: fixed

PR #972, opened 2021-06-08
Update ordinal validation

### LINKED ISSUE #971: Cannot initialize a ColumnSchema without initialized Ordinal logical type
PR #870 made a change which allowed users to initialize a column schema object with an ordinal logical type, without first instantiating the ordinal logical type, as long as the column schema was not tied to data. This seems to have been broken with the 0.4.0 release of woodwork, and it is no longer possible to do this.

The ability to use a non-instantiated Ordinal logical type is needed/preferred for the Featuretools integration.

#### Code Sample - Woodwork 0.3.1

```python
>>> from woodwork.column_schema import ColumnSchema
>>> from woodwork.logical_types import Ordinal
>>> ColumnSchema(logical_type=Ordinal)
<ColumnSchema (Logical Type = Ordinal)>

```

#### Code Sample - Woodwork 0.4.0

```python
>>> from woodwork.column_schema import ColumnSchema
>>> from woodwork.logical_types import Ordinal
>>> ColumnSchema(logical_type=Ordinal)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/nate.parsons/dev/tmp/ordinal-test/env/lib/python3.8/site-packages/woodwork/column_schema.py", line 35, in __init__
    logical_type = logical_type()
  File "/Users/nate.parsons/dev/tmp/ordinal-test/env/lib/python3.8/site-packages/woodwork/logical_types.py", line 369, in __init__
    raise TypeError("Must use an Ordinal instance with order values defined")
TypeError: Must use an Ordinal instance with order values defined

```

thehomebrewnerd:
- Update ordinal validation
- Closes #971 

Move some validation checks for Ordinal logical type from `init` to `_validate_data` so that it is now possible to create a column schema using an Ordinal logical type without having order values defined. Attempting to use the ordinal logical type on a series or dataframe without having proper order values defined will still result in an error as before.

### DIFF
--- a/woodwork/tests/logical_types/test_logical_types.py
+++ b/woodwork/tests/logical_types/test_logical_types.py
@@ -39,18 +39,12 @@ def test_instantiated_type_str():
 
 def test_ordinal_order_errors():
     series = pd.Series([1, 2, 3]).astype('category')
-    error = "Must use an Ordinal instance with order values defined"
-    with pytest.raises(TypeError, match=error):
-        series.ww.init(logical_type='Ordinal')
-
-    with pytest.raises(TypeError, match=error):
-        series.ww.init(logical_type=Ordinal)
 
     with pytest.raises(TypeError, match='Order values must be specified in a list or tuple'):
-        series.ww.init(logical_type=Ordinal(order='not_valid'))
+        Ordinal(order='not_valid').transform(series)
 
     with pytest.raises(ValueError, match='Order values cannot contain duplicates'):
-        series.ww.init(logical_type=Ordinal(order=['a', 'b', 'b']))
+        Ordinal(order=['a', 'b', 'b']).transform(series)
 
 
 def test_ordinal_init_with_order():
--- a/woodwork/tests/logical_types/test_logical_types.py
+++ b/woodwork/tests/logical_types/test_logical_types.py
@@ -38,11 +38,19 @@ def test_instantiated_type_str():
 
 
 def test_ordinal_order_errors():
+    series = pd.Series([1, 2, 3]).astype('category')
+    error = "Must use an Ordinal instance with order values defined"
+    with pytest.raises(TypeError, match=error):
+        series.ww.init(logical_type='Ordinal')
+
+    with pytest.raises(TypeError, match=error):
+        series.ww.init(logical_type=Ordinal)
+
     with pytest.raises(TypeError, match='Order values must be specified in a list or tuple'):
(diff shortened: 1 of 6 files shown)


jeff-hernandez (approved):
Looks good!

woodwork/tests/logical_types/test_logical_types.py
```diff
@@ -38,11 +38,19 @@ def test_instantiated_type_str():
...
+    error = "Must use an Ordinal instance with order values defined"
+    with pytest.raises(TypeError, match=error):
+        series.ww.init(logical_type='Ordinal')
+
+    with pytest.raises(TypeError, match=error):
+        series.ww.init(logical_type=Ordinal)
+
     with pytest.raises(TypeError, match='Order values must be specified in a list or tuple'):
-        Ordinal(order='not_valid')
+        series.ww.init(logical_type=Ordinal(order='not_valid'))
```
tamargrey, line 50: might be more related to the logical type tests to do `Ordinal(order='not_valid').transform(series)` here and in the other check below.

Though the top two checks here can't be done outside of the series init... but I think they get checked in `test_column_accessor` so I'm not sure they're necessary

woodwork/tests/schema/test_column_schema.py
```diff
@@ -410,3 +410,9 @@ def test_schema_repr():
             "<ColumnSchema (Semantic Tags = ['category', 'foreign_key'])>")
     assert (repr(ColumnSchema()) ==
             "<ColumnSchema>")
+
+
+def test_ordinal_without_init():
```
tamargrey, line 415: I'm I'm understanding correctly, this should also be possible in the Table Schema--can we add a similar test there as well?

tamargrey (commented):
Moving the ordinal checks to `_validate_data` makes so much sense! A couple of comments about testing, but other than that, looking good.

woodwork/tests/logical_types/test_logical_types.py
thehomebrewnerd: Yeah, makes sense. I removed the two that are tested in `test_column_accessor` and updated the others to use `transform` as you suggested.

woodwork/tests/schema/test_column_schema.py
thehomebrewnerd: Yep - added!

tamargrey (approved):
LGTM!

PR #979, opened 2021-06-11
Update callback function arguments

### LINKED ISSUE #966: Add total, progress and unit to describe callback
- As a user, I wish I could use Woodwork to get total, progress and unit of my describe callback.
```json
- progress (int/float): the progress so far in the calculations
- total (int/float): the total number of calculations to do
- unit (str): what is the units of progress/total
```

### LINKED ISSUE #967: Add total, progress and unit to mutual info callback
- As a user, I wish I could use Woodwork to get total, progress and unit of my mutual info callback.
```json
- progress (int/float): the progress so far in the calculations
- total (int/float): the total number of calculations to do
- unit (str): what is the units of progress/total
```
- **The implementation should remove `progress_percent`, as it is redundant.**

thehomebrewnerd:
- Update callback function arguments
- Closes #966
- Closes #967 

This PR updates the required arguments for the progress callback function that can be passed to the `describe` or `mutual_information` methods. 

The measurements were changed from percentage complete to total units. Callback functions should now accept the following:
- update: change since last call
- progress: total progress so far
- total: total units of calculations to perform
- unit: the measurement unit for the progress
- time_elapsed: total time elapsed since call started

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -19,11 +19,10 @@ Future Release
 Breaking Changes
 ++++++++++++++++
     * Progress callback functions are now expected to accept five parameters: progress increment since last call,
-    progress units complete so far, total units to complete, the progress unit of measurement, and time elapsed
-    since start of calculation. Additionally, progress is now being reported in the units specified by the unit of
-    measurement parameter instead of percentage of total.
+      progress units complete so far, total units to complete, the progress unit of measurement, and time elapsed
+      since start of calculation. Additionally, progress is now being reported in the units specified by the unit of
+      measurement parameter instead of percentage of total.
 
-    
 v0.4.1 Jun 9, 2021
 ==================
     * Enhancements
@@ -39,7 +38,7 @@ v0.4.1 Jun 9, 2021
         * Fix bug in ``test_list_logical_types_default`` (:pr:`954`)
         * Update minimum unit tests to run on all pull requests (:pr:`952`)
         * Pass token to authorize uploading of codecov reports (:pr:`969`)
-        
+
     Thanks to the following people for contributing to this release:
     :user:`frances-h`, :user:`gsheni`, :user:`tamargrey`, :user:`thehomebrewnerd`
     
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -6,6 +6,7 @@ Release Notes
 Future Release
 ==============
     * Enhancements
+        * Pass additional progress information in callback functions (:pr:`979`)
     * Fixes
     * Changes
     * Documentation Changes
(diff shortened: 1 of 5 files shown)


woodwork/statistics_utils.py
```diff
@@ -23,8 +23,10 @@ def _get_describe_dict(dataframe, include=None, callback=None):
             will be returned.
         callback (callable, optional): function to be called with incremental updates. Has the following parameters:
 
-            - update: percentage change (float between 0 and 100) in progress since last call
-            - progress_percent: percentage (float between 0 and 100) of total computation completed
+            - update (int/float): change in progress since last call
```
gsheni, line 26: Is there a way to make this consistently return one type? Or is this `int/float` for a reason?

thehomebrewnerd: I pulled that directly from the original issue, but I the current implementation will only return `int` values. I'm thinking in general we could have progress measured in non whole number increments, but that might be unlikely. 

I think we could restrict to just `int` for the time being, but if we ever had function in which progress was calculated as a `float` (like percentage for, example) we would just need to remember to update this or cast the values to int before calling the callback function.

gsheni: That's fair. Let's change the doc string to `int` to accurately reflect the returned value. 

We could also cast to `float` to better description the progress (rather than `int`) when the situation you describe comes up.

thehomebrewnerd: Sounds good.

docs/source/release_notes.rst
```diff
@@ -6,15 +6,23 @@ Release Notes
...
-    :user:`gsheni`
-    
+    :user:`gsheni`, :user:`thehomebrewnerd`
+
+Breaking Changes
+++++++++++++++++
+    * Progress callback functions are now expected to accept five parameters: progress increment since last call,
+      progress units complete so far, total units to complete, the progress unit of measurement, and time elapsed
+      since start of calculation. Additionally, progress is now being reported in the units specified by the unit of
+      measurement parameter instead of percentage of total.
```
rwedge, line 24: thoughts on making the parameters a nested bulleted list?

thehomebrewnerd: Probably would make it easier to read. I'll update!

docs/source/release_notes.rst
```diff
@@ -6,15 +6,28 @@ Release Notes
...
+++++++++++++++++
+    * Progress callback functions are now expected to accept five parameters: 
     
+        * progress increment since last call
+        * progress units complete so far
+        * total units to complete
+        * the progress unit of measurement
+        * time elapsed since start of calculation
+        
+    Additionally, progress is now being reported in the units specified by the unit of measurement parameter instead of percentage of total.
```
rwedge, line 29: https://woodwork.alteryx.com/en/callback-update/release_notes.html

This indentation looks off

thehomebrewnerd: I reworked the entire block to (hopefully) better adhere to the bulleted format we have been using, combining the leading and trailing sentences into a single first-level bullet.

rwedge: looks good!

woodwork/statistics_utils.py
thehomebrewnerd: @gsheni Aside from this change was there anything else you saw that needs to be updated?

gsheni: nope, looks good

gsheni (approved):

PR #987, opened 2021-06-15
Test compatibility with upcoming pandas release 1.3.0 

### LINKED ISSUE #985: Test compatibility with upcoming pandas release 1.3.0 for Woodwork
- We should make a branch that uses the release candidate and test if it's compatible with Woodwork currently

tamargrey:
FuruteWarnings that now show up:
- `test_dataframe_methods_on_accessor_returning_series[sample_df_dask, sample_df_pandas, ]` - FutureWarning: Dropping of nuisance columns in DataFrame reductions (with 'bool_only=None') is deprecated; in a future version this will raise TypeError.  Select only valid columns before calling the reduction.
- `test_sets_datetime64_dtype_on_init` - FutureWarning: Inferring datetime64[ns] from data containing strings is deprecated and will be removed in a future version. To retain the old behavior explicitly pass Series(data, dtype={value.dtype})
- in `statistics_utils.py:169` - FutureWarning: casting datetime64[ns] values to int64 with .astype(...) is deprecated and will raise in a future version. Use .view(...) instead.
    data[col_name] = pd.qcut(data[col_name].astype('int64'), num_bins, duplicates="drop")
- in a bunch of places - FutureWarning: The `inplace` parameter in pandas.Categorical.set_categories is deprecated and will be removed in a future version. Removing unused categories will always return a new Categorical object.

Errors that now show up
- on `sample_df_dask` fixture (though not everywhere that it's used) - `AttributeError: 'Series' object has no attribute 'view'`
- in statistics tests because of `_get_mode` - TypeError: int() argument must be a string, a bytes-like object or a number, not 'NAType'

closes #985


tamargrey:
> in statistics tests because of _get_mode - TypeError: int() argument must be a string, a bytes-like object or a number, not 'NAType'

Pandas 1.3.0 can no longer calculate the mode for a `BooleanDtype` if it contains nans: `pd.Series([False, True, False, None], dtype='boolean').mode()`

The temporary fix might be to remove the series from the stastistics tests that are booleans with nans.

tamargrey:
> on sample_df_dask fixture (though not everywhere that it's used) - AttributeError: 'Series' object has no attribute 'view'

At Woodwork initialization, we check `_is_numeric_series`, and part of that check uses `pd.to_numeric`, which has been updated to use `DataFrame.view('int64')` in pandas 1.3.0 (instead of `astype('int64'), I think). 

It seems we were able to use `pd.to_numeric` for Dask DataFrames in the past, but Dask (or at least version `2.30.0`) has not added a `view` method to Series, so this is not yet possible. Koalas dataframes use `to_pandas`, so my suggestion to fix this now would be to call compute on the Dask series as well. 

Side note:

Changing from `astype('int64') to `view('int64')` is one of the warnings that Pandas is now raising, so if Dask does not yet have `view` for Series, it may mean we will not want to make those changes yet.

thehomebrewnerd:
> Changing from `astype('int64')` to `view('int64')` is one of the warnings that Pandas is now raising, so if Dask does not yet have `view` for Series, it may mean we will not want to make those changes yet.

We are currently restricting our Dask version to `dask[dataframe]>=2.30.0,<2012.12.0` so even if Dask is adding support we wouldn't get it. Maybe we need to take a look into removing this Dask version restriction now?

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -906,7 +906,9 @@ def test_sets_datetime64_dtype_on_init():
         pd.Series(['2020-01-01', None, '2020-01-03'], name=column_name),
         pd.Series(['2020-01-01', np.nan, '2020-01-03'], name=column_name),
         pd.Series(['2020-01-01', pd.NA, '2020-01-03'], name=column_name),
-        pd.Series(['2020-01-01', pd.NaT, '2020-01-03'], name=column_name),
+        # Inferring datetime64[ns] from data containing strings is deprecated
+        # and will be removed in a future version. Only this was a problem.
+        # pd.Series(['2020-01-01', pd.NaT, '2020-01-03'], name=column_name),
```
tamargrey, line 911: I don't entirely understand why only this Series raised the warning. Not sure if we should find some other way to test inference with `NaT` or ignore for now

woodwork/type_sys/utils.py
```diff
@@ -40,6 +41,8 @@ def _is_numeric_series(series, logical_type):
     for the purposes of determining if it can be a time_index."""
     if ks and isinstance(series, ks.Series):
         series = series.to_pandas()
+    if dd and isinstance(series, dd.Series):
+        series = series.compute()
```
thehomebrewnerd, line 45: Can we use only the first partition here as well?

woodwork/tests/accessor/test_table_accessor.py
thehomebrewnerd: Can we set the series dtype explicitly here (maybe `object`?) to avoid this warning?

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -1340,7 +1342,7 @@ def test_dataframe_methods_on_accessor_inplace(sample_df):
 
 
 def test_dataframe_methods_on_accessor_returning_series(sample_df):
-    schema_df = sample_df.copy()
+    schema_df = sample_df[['id', 'age', 'is_registered']]
```
thehomebrewnerd, line 1345: Is there a reason we had to make this change for this test?

thehomebrewnerd (commented):
Just a couple small comments/questions.

woodwork/tests/accessor/test_table_accessor.py
tamargrey: yes, this is related to the `FutureWarning: Dropping of nuisance columns in DataFrame reductions (with 'bool_only=None') is deprecated; in a future version this will raise TypeError. Select only valid columns before calling the reduction.` warning. Seems like reductions like `min`, `max,` and `all` that reduce a DataFrame to a Series now should only be called on valid columns.

woodwork/type_sys/utils.py
tamargrey: yep!

tamargrey: I don't think I meant to add it to `_is_valid_latlong_series` in the first place, but I guess it's a part of type inference, so maybe we actually want it there...

woodwork/tests/accessor/test_table_accessor.py
tamargrey: object works! thanks for the suggestion!

woodwork/statistics_utils.py
```diff
@@ -148,6 +148,9 @@ def _get_describe_dict(dataframe, include=None, callback=None,
 
 def _get_mode(series):
     """Get the mode value for a series"""
+    if str(series.dtype) == 'boolean':
+        # TODO: remove when pandas fixes bug where mode can't be calculated on BooleanDtype with nans
+        series = series.astype('object')
```
tamargrey, line 153: A [backport PR](https://github.com/pandas-dev/pandas/pull/42131) has been merged in, so, assuming that means that this change will get into pandas 1.3.0, we should be able to remove this. (tested those changes locally as well to confirm they fix the problems in `_get_mode`)

woodwork/tests/accessor/test_table_accessor.py
thehomebrewnerd: Got it. I think we could include the `signup_date` column as well, but not a big deal either way.

woodwork/statistics_utils.py
thehomebrewnerd: Do you think we should wait to merge in this PR until they release 1.3.0 so we can confirm this was fixed?

tamargrey: Possibly. It seems that they're still going to be making various fixes in the mean time, so it might be worth waiting just to confirm that a new bug isn't introduced

tamargrey: now that pandas 1.3.0 is in, removed these lines, and the tests still pass :)

gsheni (approved):

docs/source/release_notes.rst
```diff
@@ -45,6 +45,7 @@ v0.4.2 Jun 23, 2021
         * Temporarily remove update checker to fix docs warnings (:pr:`1011`)
     * Testing Changes
         * Add env setting to update checker (:pr:`978`, :pr:`994`)
+        * Test woodwork on pandas 1.3.0 release candidate and make fixes (:pr:`987`)
```
thehomebrewnerd, line 48: I think I'd update this and move it to the `Changes` section to reinforce we are making updates to be compatible with pandas 1.3.0, not just updating our tests.

thehomebrewnerd (approved):
One suggestion regarding the release notes, but otherwise looks fine to me!

PR #988, opened 2021-06-15
Add option to compute extra stats with describe

### LINKED ISSUE #960: Add additional optional outputs to df.ww.describe()
As a user, I wish Woodwork's `describe` and `describe_dict` methods would optionally include additional stats depending on the column type:

- numeric -> add histogram with specified number of bins
- categorical -> add "top N" items by occurrence
- datetime -> add "N most recent" dates

This would require adding some additional parameters to the methods to control this behavior and provide the necessary inputs. The default behavior would produce results identical to those obtained currently.
#### Code Example

```python
df.ww.describe(extra_stats=True, bins=5, top_n=10, n_most_recent=8)
df.ww.describe_dict(extra_stats=True, bins=5, top_n=10, n_most_recent=8)
```

thehomebrewnerd:
- Add option to compute extra stats to describe
- Closes #960

Adds the ability to compute extra optional stats with the `describe` method by passing `extra_stats=True` along with other optional parameter to control the calculations. Extra stats include:
- histogram (numeric cols)
- top_values (categorical and certain integer cols)
- most_recent (datetime cols)

### DIFF
--- a/woodwork/statistics_utils.py
+++ b/woodwork/statistics_utils.py
@@ -4,14 +4,21 @@ import numpy as np
 import pandas as pd
 from sklearn.metrics.cluster import normalized_mutual_info_score
 
-from woodwork.logical_types import Datetime, Double, LatLong
+from woodwork.logical_types import (
+    Datetime,
+    Double,
+    Integer,
+    IntegerNullable,
+    LatLong
+)
 from woodwork.utils import _update_progress, get_valid_mi_types, import_or_none
 
 dd = import_or_none('dask.dataframe')
 ks = import_or_none('databricks.koalas')
 
 
-def _get_describe_dict(dataframe, include=None, callback=None):
+def _get_describe_dict(dataframe, include=None, callback=None,
+                       extra_stats=False, bins=10, top_x=10, recent_x=10):
     """Calculates statistics for data contained in a DataFrame using Woodwork typing information.
 
     Args:
@@ -28,6 +35,16 @@ def _get_describe_dict(dataframe, include=None, callback=None):
             - total (int): the total number of calculations to do
             - unit (str): unit of measurement for progress/total
             - time_elapsed (float): total time in seconds elapsed since start of call
+
+        extra_stats (bool): If True, will calculate a histogram and top values for numeric columns, top values
+            for categorical columns and recent values for datetime columns. Output can be controlled by bins,
+            top_x and recent_x parameters.
+        bins (int): Number of bins to use when calculating histogram
+            for numeric columns. Defaults to 10.
+        top_x (int): Number of values to get when getting the most frequent
+            values for categorical columns. Defaults to 10
+        recent_x (int): Number of recent values to get when getting the most recent
+            dates to calculate frequency for datetime columns. Defaults to 10.
 
     Returns:
         dict[str -> dict]: A dictionary with a key for each column in the data or for each column
@@ -106,6 +123,20 @@ def _get_describe_dict(dataframe, include=None, callback=None):
         values["physical_type"] = series.dtype
         values["logical_type"] = logical_type
         values["semantic_tags"] = semantic_tags
+
+        # Calculate extra details stats, if requested
+        if extra_stats:
+            if column.is_numeric:
+                values["histogram"] = _get_histogram_values(series, bins=bins)
+                if isinstance(column.logical_type, (Integer, IntegerNullable)):
+                    _range = range(int(values["min"]), int(values["max"]) + 1)
+                    if len(_range) <= bins:
+                        values["top_values"] = _get_top_values_numeric(series, _range)
+            elif column.is_categorical:
+                values["top_values"] = _get_top_values_categorical(series, top_x)
+            elif column.is_datetime:
+                values["recent_values"] = _get_recent_values(series, recent_x)
+
         results[column_name] = values
         current_progress = _update_progress(start_time, timer(), 1, current_progress, total_loops, unit, callback)
     return results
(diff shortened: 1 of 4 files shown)


woodwork/statistics_utils.py
```diff
@@ -29,6 +36,17 @@ def _get_describe_dict(dataframe, include=None, callback=None):
             - unit (str): unit of measurement for progress/total
             - time_elapsed (float): total time in seconds elapsed since start of call
 
+        extra_stats (bool): If True, will calculate a histogram and top values for numeric columns, top values
```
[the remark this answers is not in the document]
tamargrey: Also, since the logic where we make that decision of whether or not to also calculate top values is a little hard to parse, can we add a comment above those lines explaining why we do or don't calculate it? (If I'm understanding correctly, something like "only calculate top values if there are more bins than possible values" might make sense)

woodwork/statistics_utils.py
```diff
@@ -306,3 +338,79 @@ def _get_value_counts(dataframe, ascending=False, top_n=10, dropna=False):
...
+def _get_top_values_categorical(series, num_x):
+    """Get the most frequent values in a pandas Series. Will exclude null values.
+
+    Args:
+        column (pd.Series): data to use find most frequent values
+        num_x (int): the number of top values to retrieve
+
+    Returns:
+        top_list (list(dict)): a list of dictionary with keys `count` and
+            `value`
```
tamargrey, line 370: Probably worth mentioning that the list sorted so that the later in the list, the higher the count (I initially assumed the opposite, and since the dataframe output only shows the first element of the list, was confused as to why it didn't match the value at mode).

Either that, or the order should be reversed.

tamargrey: Also, I think that the numeric top count is sorted on `value` and not `count`, so they should probably match or be more clearly differentiated.

tamargrey: Another small inconsistency is that the dict here is `{value:n, count:n}`, and in the numeric calculation its `{count:n, value:n}`

woodwork/statistics_utils.py
```diff
@@ -106,6 +124,20 @@ def _get_describe_dict(dataframe, include=None, callback=None):
...
+        # Calculate extra detailed stats, if requested
+        if extra_stats:
+            if column.is_numeric:
+                values["histogram"] = _get_histogram_values(series, bins=bins)
+                if isinstance(column.logical_type, (Integer, IntegerNullable)):
+                    _range = range(int(values["min"]), int(values["max"]) + 1)
+                    if len(_range) <= bins:
+                        values["top_values"] = _get_top_values_numeric(series, _range)
+            elif column.is_categorical:
+                values["top_values"] = _get_top_values_categorical(series, top_x)
```
tamargrey, line 137: Likely not worth the extra logic required of this, because it seems we don't include the index column, but if the column is unique, we probably don't need to calculate top values and maybe histogram. Since we've already calculated count and nunique it wouldn't take additional computation to check.

woodwork/statistics_utils.py
```diff
@@ -306,3 +338,79 @@ def _get_value_counts(dataframe, ascending=False, top_n=10, dropna=False):
...
+    frequencies = series.value_counts(dropna=True)
+    df = frequencies.head(num_x).reset_index()
+    df.columns = ["value", "count"]
+    top_lt = list(df.to_dict(orient="index").values())
+    top_lt = sorted(top_lt, key=lambda i: (i["count"], i["value"]))
+    return top_lt
+
+
+def _get_recent_values(column, num_x):
+    """Get the most frequent, recent values for a given datetime column.
```
tamargrey, line 381: Reading this, I'm not sure if we're getting the "n most recent dates and listing the counts for each" or the "n most frequent dates sorted by recency" or something. The recent values language in general is a bit confusing and could maybe be a bit more consistent.

woodwork/statistics_utils.py
```diff
@@ -306,3 +338,79 @@ def _get_value_counts(dataframe, ascending=False, top_n=10, dropna=False):
         values = list(df.to_dict(orient="index").values())
         val_counts[col] = values
     return val_counts
+
+
+def _get_top_values_numeric(series, _range):
+    """Get the most frequent values in a range of integers. Will ignore
+        null values.
```
tamargrey, line 345: This function isn't really getting the top numeric values. It's probably more accurate to say something like "Get the frequencies of the values in a range of integers" since this is just grabbing all the frequencies and not sorting them in any way currently.

For example the column `'age': pd.Series([33, 34, 36, 34], dtype='Int64'),` produces this dictionary:

```python
'top_values': [{'count': 1, 'value': 33},
   {'count': 2, 'value': 34},
   {'count': 0, 'value': 35},
   {'count': 1, 'value': 36}]},
```

woodwork/table_accessor.py
```diff
@@ -825,16 +825,28 @@ def describe_dict(self, include=None, callback=None):
...
             with a value containing a dictionary containing relevant statistics for that column.
         """
         if self._schema is None:
             _raise_init_error()
-        return _get_describe_dict(self._dataframe, include=include, callback=callback)
+        return _get_describe_dict(self._dataframe, include=include, callback=callback,
+                                  extra_stats=extra_stats, bins=bins, top_x=top_x, recent_x=recent_x)
 
-    def describe(self, include=None, callback=None):
+    def describe(self, include=None, callback=None, extra_stats=False, bins=10, top_x=10, recent_x=10):
```
tamargrey, line 849: The extra stats added here aren't really understandable from the dataframe format. 

image

I guess maybe their purpose isn't to be useful for a quick onceover the way the other stats are in this format, so maybe that's okay. Just wanted to point it out, though.

gsheni: Yah these stats don't fit in neatly in a DataFrame, which is why they are off by default with doing `df.ww.describe`

woodwork/statistics_utils.py
```diff
@@ -106,6 +124,20 @@ def _get_describe_dict(dataframe, include=None, callback=None):
...
+            if column.is_numeric:
+                values["histogram"] = _get_histogram_values(series, bins=bins)
+                if isinstance(column.logical_type, (Integer, IntegerNullable)):
+                    _range = range(int(values["min"]), int(values["max"]) + 1)
+                    if len(_range) <= bins:
+                        values["top_values"] = _get_top_values_numeric(series, _range)
+            elif column.is_categorical:
+                values["top_values"] = _get_top_values_categorical(series, top_x)
+            elif column.is_datetime:
+                values["recent_values"] = _get_recent_values(series, recent_x)
```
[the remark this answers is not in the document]
thehomebrewnerd: I think we probably could do that, but I'd propose we hold off on doing that for now since that wasn't part of the initial implementation we are transferring over to Woodwork. Maybe we wait and see if we identify a need for that and implement at that time?

woodwork/table_accessor.py
```diff
@@ -875,6 +899,8 @@ def describe(self, include=None, callback=None):
             'num_true',
             'num_false',
         ]
+        if extra_stats:
+            index_order = index_order + ['histogram', 'top_values', 'recent_values']
```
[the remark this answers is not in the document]
thehomebrewnerd: Updated

woodwork/statistics_utils.py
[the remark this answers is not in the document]
thehomebrewnerd: Updated the docstring and added a comment in the code to clarify the logic a bit

woodwork/table_accessor.py
thehomebrewnerd: We could leave this option out of `describe` and only include it for `describe_dict`?

thehomebrewnerd: Will only use this in `describe_dict` for now.

woodwork/statistics_utils.py
thehomebrewnerd: Left sorting order the same for now, but updated all dictionaries so the are formatted as `{"value": n, "count": x}`

thehomebrewnerd: Also updated docstrings to mention the sorting order of the output.

woodwork/statistics_utils.py
thehomebrewnerd: Updated the wording for this. The function was actually getting the value counts for the `n` oldest dates in the series. Updated the function name and the docstring.

woodwork/statistics_utils.py
thehomebrewnerd: Updated this function name and docstring - output is actually sorted by the values defined in the input range.

woodwork/statistics_utils.py
```diff
@@ -306,3 +342,81 @@ def _get_value_counts(dataframe, ascending=False, top_n=10, dropna=False):
...
+
+    Args:
+        column (pd.Series): data to use find value counts
+        num_x (int): the number of values to retrieve
+
+    Returns:
+        value_counts (list(dict)): a list of dictionary with keys `value` and
+            `count`. Output is sorted in ascending order by date.
+    """
+    datetimes = pd.to_datetime(column, infer_datetime_format=True, errors="coerce")
```
gsheni, line 396: The column will already be a datetime column, right? So no need to re-cast it? 
```suggestion
    datetimes = column
```

thehomebrewnerd: Yeah, we can remove that `pd.to_datetime` col since Woodwork would have already done that conversion. I had to rework the test a little bit because of this though to remove the string values which would never be allowed in a Woodwork datetime column, but overall this cleans things up.

woodwork/tests/accessor/test_statistics.py
```diff
@@ -648,6 +654,45 @@ def __call__(self, update, progress, total, unit, time_elapsed):
...
+    ltypes = {
+        'category_col': 'Categorical',
+        'datetime_col': 'Datetime',
+        'numeric_col': 'Double',
+        'nullable_integer_col': 'IntegerNullable',
+        'integer_col': 'Integer',
+        'small_range_col': 'Integer',
+    }
+    describe_df.ww.init(index='index_col', logical_types=ltypes)
+    desc_dict = describe_df.ww.describe_dict(extra_stats=True)
```
tamargrey, line 673: can we add a check that confirms that the extra stats are not present in a dictionary where extra stats=False?

thehomebrewnerd: I updated `test_get_describe_dict` to confirm that the extra stats rows are not present.

tamargrey (approved):
lgtm!

gsheni (approved):

PR #1000, opened 2021-06-18
Remove make_index from Woodwork init and all references to it

### LINKED ISSUE #984: Remove `make_index` from TableAccessor
Woodwork's `make_index` does not currently get maintained anytime `df.ww.init(schema=schema)` is used. This impacts `getitem`, `select`, `drop`, and the indexers. 

It's also been determined that along with being buggy, `make_index` is not all that useful for Woodwork, and is primarily useful for Featuretools' purposes. We should remove the `make_index` attribute of `WoodworkTableAccessor` as well as all references to it.

tuethan1999:
- Remove `make_index` from `WoodworkTableAccessor` and all references to it
- Closes #984

### DIFF
--- a/docs/source/start.ipynb
+++ b/docs/source/start.ipynb
@@ -58,7 +58,7 @@
    "source": [
     "import woodwork as ww\n",
     "\n",
-    "df.ww.init(name=\"retail\", make_index=True, index=\"order_product_id\")\n",
+    "df.ww.init(name=\"retail\")\n",
     "df.ww"
    ]
   },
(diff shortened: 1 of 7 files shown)


docs/source/start.ipynb
```diff
@@ -58,15 +58,15 @@
    "source": [
     "import woodwork as ww\n",
     "\n",
-    "df.ww.init(name=\"retail\", make_index=True, index=\"order_product_id\")\n",
+    "df.ww.init(name=\"retail\")\n",
```
tamargrey, line 61: Later on in the docs, I believe we try and set the index to `order_product_id` which will now fail. I believe there is no unique column on this dataframe, so in order to show setting the index, you may have to manually add the index to `df` here.

tamargrey (commented):
Looking good!

gsheni (approved):

PR #1001, opened 2021-06-21
Raise type conversion error in datetime

### LINKED ISSUE #991: Datetime logical type transform does not throw TypeConversionError
When transforming data for a specific logical type with `LogicalType.transform`, if the transformation to the correct dtype is not possible, a `TypeConversionError` should be thrown. This is not currently the case with the Datetime logical type.

The following code, which should produce a woodwork `TypeConversionError`
```python
import pandas as pd
from woodwork.logical_types import Datetime

s = pd.Series(['a', 'b', 'c'])
Datetime().transform(s)
```
produces the following error: `ValueError: Given date string not likely a datetime.`

The base LogicalType class' transform catches any errors during the `astype` call and raises a TypeConversionError. The other logical types with their own transform functions might raise separate types of errors (ordinal with a data validation error, or latlong if the format does not match the required latlong format) before calling the super transform, but those are not related to the dtype change, so it makes sense that they're not `TypeConversionErrors`.

However with Datetime, we perform a separate astype call via the `to_datetime` function. This transformation was previously treated the same as the other dtype changes and, if an error occurred, was raised to the user as a `TypeConversionError`. Now, however, we do not have a try/except around the datetime-specific logic in `Datetime.transform`, so the pandas error is what we see.

We should make sure that a TypeConversionError is thrown in this case.

jeff-hernandez:
Closes #991 by raising a type conversion error when a Datetime logical type cannot parse data to a datetime dtype.

### DIFF
--- a/woodwork/logical_types.py
+++ b/woodwork/logical_types.py
@@ -47,14 +47,15 @@ class LogicalType(object, metaclass=LogicalTypeMetaClass):
         new_dtype = self._get_valid_dtype(type(series))
         if new_dtype != str(series.dtype):
             # Update the underlying series
+            error = TypeConversionError(series, new_dtype, type(self))
             try:
                 series = series.astype(new_dtype)
                 if str(series.dtype) != new_dtype:
                     # Catch conditions when Panads does not error but did not
                     # convert to the specified dtype (example: 'category' -> 'bool')
-                    raise TypeConversionError(series, new_dtype, type(self))
+                    raise error
             except (TypeError, ValueError):
-                raise TypeConversionError(series, new_dtype, type(self))
+                raise error
         return series
 
 
--- a/woodwork/logical_types.py
+++ b/woodwork/logical_types.py
@@ -47,17 +47,14 @@ class LogicalType(object, metaclass=LogicalTypeMetaClass):
         new_dtype = self._get_valid_dtype(type(series))
         if new_dtype != str(series.dtype):
             # Update the underlying series
-            error_msg = f'Error converting datatype for {series.name} from type {str(series.dtype)} ' \
-                f'to type {new_dtype}. Please confirm the underlying data is consistent with ' \
-                f'logical type {type(self)}.'
             try:
                 series = series.astype(new_dtype)
                 if str(series.dtype) != new_dtype:
                     # Catch conditions when Panads does not error but did not
                     # convert to the specified dtype (example: 'category' -> 'bool')
-                    raise TypeConversionError(error_msg)
+                    raise TypeConversionError(series, new_dtype, type(self))
             except (TypeError, ValueError):
-                raise TypeConversionError(error_msg)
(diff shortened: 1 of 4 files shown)


woodwork/logical_types.py
```diff
@@ -183,14 +183,20 @@ def transform(self, series):
...
+                    series = dd.to_datetime(series, format=self.datetime_format)
+                    series.name = name
+                elif ks and isinstance(series, ks.Series):
+                    series = ks.Series(ks.to_datetime(series.to_numpy(), format=self.datetime_format), name=series.name)
+                else:
+                    series = pd.to_datetime(series, format=self.datetime_format)
+            except (TypeError, ValueError):
+                message = f'Error converting datatype for {series.name} from type {str(series.dtype)} '
+                message += f'to type {new_dtype}. Please confirm the underlying data is consistent with '
+                message += f'logical type {type(self)}.'
```
tamargrey, line 198: Can we add this error message to the `TypeConversionError` class since the error is the same both here and when its called in the original `transform`?

jeff-hernandez: Yeah do we want `series`, `new_dtype`, and `logical_type` as parameters for `TypeConversionError`?

tamargrey: that makes sense to me!

jeff-hernandez: ✅  c19341e

woodwork/tests/logical_types/test_logical_types.py
```diff
@@ -110,6 +111,14 @@ def test_datetime_transform(datetimes):
         assert str(transform.dtype) == 'datetime64[ns]'
 
 
+def test_datetime_conversion_error():
+    series = pd.Series(['a', 'b', 'c'])
+    match = 'Error converting datatype for None from type object to type datetime64[ns]. '
```
thehomebrewnerd, line 116: Minor detail, but I think it would be good to name the series in this test just to make sure that the series name is being displayed properly.

thehomebrewnerd (approved):
One small suggestion, but otherwise looks good to me.

woodwork/tests/logical_types/test_logical_types.py
```diff
@@ -110,6 +111,14 @@ def test_datetime_transform(datetimes):
         assert str(transform.dtype) == 'datetime64[ns]'
 
 
+def test_datetime_conversion_error():
+    series = pd.Series(['a', 'b', 'c'], name='series')
```
tamargrey, line 115: Can we test with the `sample_series` fixture? Pandas and Koalas both fail as expected, but it seems Dask wouldn't show the error until a compute call is made, so probably worth xfailing there or something

jeff-hernandez: Yeah updated in 2dd47b6

tamargrey (approved):
lgtm!

Issue #965, opened 2021-06-04
Add a way of initializing Woodwork with partial Table or Column Schemas


tamargrey:
As a user, I wish I could initialize Woodwork on a DataFrame with a partial schema in a way that would perform inference on the remaining columns and be able to update the dtype of the columns that did have typing information defined.

Currently, there are two ways of initializing Woodwork on a DataFrame:
1. With all the woodwork parameters passed in separately - this allows for specifying partial typing information and inferring on the columns without typing information
2. With a `TableSchema` - requires all typing information be present and correct

Recently, there've been a few places where we needed to use the first initialization option when a significant portion of a DataFrame's typing information is known, but either the dtypes of the DataFrame may have changed or we don't have all of the typing information for all of the columns in the DataFrame so the second approach wasn't an option (woodwork's concat function, and featuretools' `EntitySet.add_last_time_indexes` are two that come to mind). 

The first initialization option makes this kind of initialization from table to table with partial typing information possible, but it is error-prone if a parameter to Woodwork initialization is forgotten or if a parameter should be deepcopied. It also adds more places that we need to remember to update any time Woodwork initialization changes.

Two options for this new method of initialization are as follows:
- Init with a dictionary mapping `col_name -> ColumnSchema` 
  - the main problem with this approach is that table-wide parameters like index and table metadata would still need to get passed in separately.
- Init with a subset `TableSchema`

Both of these would confirm that every column was actually present 

Once we're adding a third way of initializing Woodwork, it's probably too much to have all the logic handled within `df.ww.init`. It might be beneficial to at least have this in a separate method, but we can also consider having three methods:
- `df.ww.init` - pass individual parameters; anything not specified gets inferred.
- `df.ww.init_with_full_schema` - pass full table schema
- `df.ww.init_with_partial_schema` - pass partial schema; anything not specified gets inferred.

gsheni:
@tamargrey is this required for the featuretools integration or is it something we can get to later?

tamargrey:
@gsheni This can wait until after the featuretools integration!

thehomebrewnerd:
Having the ability to initialize a TableSchema from a ColumnSchema, or list of ColumnSchemas might also be useful, and could help clean up some of the code implemented in PR #1004 in `_make_series_call`.

PR #1004, opened 2021-06-22
Retain schema on series.ww.to_frame call

### LINKED ISSUE #974: Preserve Woodwork schema with series.ww.to_frame function
- As a user, I wish I could use Woodwork to preserve my schema when calling `series.to_frame()`. 
  - This would take a Series, convert it to a DataFrame, and preserve the Woodwork schema/typing.

thehomebrewnerd:
- Retain schema on series.ww.to_frame call
- Closes #974

Implements changes to `WoodworkColumnAccessor._make_series_call` to retain schema information when possible if a dataframe is returned from the series call, for example when calling `series.ww.to_frame()`.

### DIFF
--- a/woodwork/column_accessor.py
+++ b/woodwork/column_accessor.py
@@ -3,9 +3,13 @@ import warnings
 import weakref
 
 import pandas as pd
-from woodwork import logical_types
 
-from woodwork.accessor_utils import _is_series, init_series, _is_dataframe, is_schema_valid
+from woodwork.accessor_utils import (
+    _is_dataframe,
+    _is_series,
+    init_series,
+    is_schema_valid
+)
 from woodwork.column_schema import ColumnSchema
 from woodwork.exceptions import (
     ParametersIgnoredWarning,
--- a/woodwork/column_accessor.py
+++ b/woodwork/column_accessor.py
@@ -3,8 +3,9 @@ import warnings
 import weakref
 
 import pandas as pd
+from woodwork import logical_types
 
-from woodwork.accessor_utils import _is_series, init_series
+from woodwork.accessor_utils import _is_series, init_series, _is_dataframe, is_schema_valid
 from woodwork.column_schema import ColumnSchema
 from woodwork.exceptions import (
     ParametersIgnoredWarning,
@@ -13,6 +14,7 @@ from woodwork.exceptions import (
 )
 from woodwork.indexers import _iLocIndexer, _locIndexer
 from woodwork.logical_types import LatLong, Ordinal
+from woodwork.table_schema import TableSchema
 from woodwork.utils import (
     _get_column_logical_type,
     _is_valid_latlong_series,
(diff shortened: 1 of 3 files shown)


woodwork/column_accessor.py
```diff
@@ -246,6 +248,18 @@ def wrapper(*args, **kwargs):
...
+                elif _is_dataframe(result):
+                    # Initialize Woodwork if a valid table schema can be created from the original column schema
+                    col_schema = self.schema
+                    table_schema = TableSchema(column_names=[self.name],
+                                               logical_types={self.name: col_schema.logical_type},
+                                               semantic_tags={self.name: col_schema.semantic_tags},
+                                               column_metadata={self.name: col_schema.metadata},
+                                               use_standard_tags={self.name: col_schema.use_standard_tags},
+                                               column_descriptions={self.name: col_schema.description},
+                                               validate=False)
```
tamargrey, line 264: Maybe another point towards implementing something in #965 that would let us initialize Woodwork directly here?

woodwork/column_accessor.py
```diff
@@ -246,6 +248,18 @@ def wrapper(*args, **kwargs):
                                                                                           invalid_schema_message,
                                                                                           'Series')
                         warnings.warn(warning_message, TypingInfoMismatchWarning)
+                elif _is_dataframe(result):
```
tamargrey, line 255: Are there other examples of Series calls that return a DataFrame? Both ones that will and will not have a valid woodwork schema

Because we're adding functionality for those DataFrames returned that can have a valid schema, and I don't think we'll be losing functionality for non woodwork dataframes removed since we still return the DataFrame, but it'd be nice to confirm that.

woodwork/tests/accessor/test_column_accessor.py
```diff
@@ -755,3 +755,13 @@ def test_validation_methods_called_init_with_schema(mock_validate_schema, sample
     assert mock_validate_schema.called
     assert validated.ww == not_validated.ww
     pd.testing.assert_series_equal(to_pandas(validated), to_pandas(not_validated))
+
+
+def test_to_frame_maintains_schema(sample_series):
+    sample_series.ww.init(semantic_tags={'test_tag'},
+                          description='custom description',
+                          metadata={'custom key': 'custom value'})
+    sample_frame = sample_series.ww.to_frame()
```
tamargrey, line 764: Not sure if this would be redundantly checking pandas behavior, but would it be worth confirming that the type of `sample_frame` is `DataFrame`?

tamargrey (commented):
Looking good, a couple of comments about testing, but looking good implementation-wise.

It's awesome that we can expand to add `to_frame` without adding a `to_frame` method to woodwork (yay accessors!).

woodwork/column_accessor.py
thehomebrewnerd: Yep - added a comment to that issue related to this!

woodwork/tests/accessor/test_column_accessor.py
thehomebrewnerd: Don't see any harm in adding it. I'll update.

woodwork/column_accessor.py
thehomebrewnerd: I'm not sure how many exist, but calling `series.reset_index(drop=False)` also returns a frame, which will have an invalid schema. I added this case to the test and renamed the test to reflect the increased scope.

tamargrey (approved):
LGTM!

PR #1010, opened 2021-06-22
Update collections import to avoid deprecation warning with python 3.8

frances-h:


gsheni (approved):
verified that the docs no longer generate that error

PR #1015, opened 2021-06-24
Add input validation to init_series to raise a better error

### LINKED ISSUE #631: Add input validation to ww.init_series
The `init_series` function added as part of the accessor update does not validate that the input passed in for the series is a valid series type. If the user passes in a non-series object an error is currently raised, but the error message is not helpful in understanding what went wrong.  Current behavior:
```python
>>> ww.init_series(3)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/nate.parsons/dev/woodwork/woodwork/accessor_utils.py", line 38, in init_series
    logical_type = _get_column_logical_type(series, logical_type, series.name)
AttributeError: 'int' object has no attribute 'name'
```

We should add a validation check to this function and raise a better error if improper inputs are supplied.

simha104:
- Fixes #631


tamargrey:
The failing CI tests should go away once you get the latest changes from main!

docs/source/release_notes.rst
```diff
@@ -22,6 +22,7 @@ v0.4.2 Jun 23, 2021
     * Fixes
         * Raise type conversion error in ``Datetime`` logical type (:pr:`1001`)
         * Try collections.abc to avoid deprecation warning (:pr:`1010`)
+        * Add input validation to ww.init_series (:pr: '1017')
```
tamargrey: The release notes here are actually from the release that went out yesterday. This line and your name from below should go in the `Future Release` section above (you'll need to remove the two dots `..` in the two places to uncomment it, and it should follow the formatting in this section)

simha104: I have now put it in the future release!

tamargrey: You just need add your own name at the bottom of the  future release section :) everyone else will add their own when they put up PRs

docs/source/release_notes.rst
```diff
@@ -6,11 +6,13 @@ Release Notes
   ==============
```
gsheni, line 6: Uncomment the `Future Release` line by remove the `..`

docs/source/release_notes.rst
simha104: Oh okay I added the name now haha

woodwork/tests/utils/test_accessor_utils.py
```diff
@@ -37,6 +37,12 @@ def test_init_series_valid_conversion_specified_ltype(sample_series):
     assert series.ww.semantic_tags == set()
 
 
+def test_init_series_with_invalid_type(sample_df):
+    error_message = f'Input must be of series type. The current input is of type {type(sample_df)}'
+    with pytest.raises(TypeError, match=error_message):
+        init_series(sample_df)
```
tamargrey, line 43: I like the idea of testing a DataFrame here! 

Two other cases that would be nice to have here as well are some simpler object like an `int` or a `str` as well as `None` which are other possible accidental inputs to `init_series`.

simha104: Are there any fixtures for those two data types that I can use?

tamargrey: Nope, we generally use fixtures when we need to build an object up for a test, which lets us avoid repetitive code, but something like `1` or `None` is just an object that doesn't need any initialization, so you can just use it directly

simha104: Alright I've added checks for those three datatypes

woodwork/accessor_utils.py
```diff
@@ -30,14 +30,16 @@ def init_series(series, logical_type=None, semantic_tags=None,
...
+    if _is_series(series):
+        logical_type = _get_column_logical_type(series, logical_type, series.name)
+        new_series = logical_type.transform(series)
+        new_series.ww.init(logical_type=logical_type,
+                           semantic_tags=semantic_tags,
+                           use_standard_tags=use_standard_tags,
+                           description=description,
+                           metadata=metadata)
+    else:
+        raise TypeError(f'Input must be of series type. The current input is of type {type(series)}')
```
thehomebrewnerd, line 42: This is a minor point, but we could refactor like this:
```
if not _is_series(series):
    raise TypeError(f'Input must be of series type. The current input is of type {type(series)}')

logical_type = _get_column_logical_type(series, logical_type, series.name)
new_series = logical_type.transform(series)
new_series.ww.init(logical_type=logical_type,
                   semantic_tags=semantic_tags,
                   use_standard_tags=use_standard_tags,
                   description=description,
                   metadata=metadata)
```
This keeps all the normal initialization stuff outside of the if/else block and (I think) makes it a little more clear that we are raising the TypeError when the input is not a series, as it puts the `raise` right by the if clause.

woodwork/tests/utils/test_accessor_utils.py
```diff
@@ -37,6 +37,21 @@ def test_init_series_valid_conversion_specified_ltype(sample_series):
...
+    error_message = f'Input must be of series type. The current input is of type {type(1)}'
+    with pytest.raises(TypeError, match=error_message):
+        init_series(1)
+    error_message = f'Input must be of series type. The current input is of type {type("string")}'
+    with pytest.raises(TypeError, match=error_message):
+        init_series("string")
+    error_message = f'Input must be of series type. The current input is of type {type(None)}'
+    with pytest.raises(TypeError, match=error_message):
+        init_series(None)
```
thehomebrewnerd, line 52: We have a lot of duplicate code here. We could clean this up by putting all the input values we want to test inside a list and then looping over that list, updating the error message and calling init_series to make sure it works properly:
```
inputs = [sample_df, 1, "string", None]
for input in inputs:
    # Update error message with input
    # Call init_series inside with context mgr to check error is raised
```

thehomebrewnerd (commented):
I think the code here is fine, but just a couple suggestions on how we might be able to clean things up a little.

woodwork/accessor_utils.py
simha104: Alright yeah that sounds good, I have refactored it as such

woodwork/tests/utils/test_accessor_utils.py
simha104: I understand, I have put it all within a loop

docs/source/release_notes.rst
```diff
@@ -2,19 +2,19 @@
...
+        * Add input validation to ww.init_series (:pr:`1015`)
     * Changes
         * Remove lines in ``LogicalType.transform`` that raise error if dtype conflicts (:pr:`1012`)
         * Add ``infer_datetime_format`` param to speed up ``to_datetime`` calls (:pr:`1016`)
     * Documentation Changes
     * Testing Changes
 
     Thanks to the following people for contributing to this release:
-    :user:`tamargrey`, :user:`thehomebrewnerd`
+    :user:`simha104`, :user:`tamargrey`, :user:`tamargrey`, :user:`thehomebrewnerd`
```
tamargrey, line 17: my name got added a second time in the merge

thehomebrewnerd (approved):
One the release notes are cleaned up per @tamargrey's comment, this should be ready to merge.

tamargrey (approved):
looks good to me!

PR #1016, opened 2021-06-24
Add infer_datetime_format param to to_datetime calls

### LINKED ISSUE #218: Add infer_datetime_format to pd.to_datetime to support faster datetime inference
- Pandas to_datetime function has a arg `infer_datetime_format`, that when True, can infer formats of datetime faster
- https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.to_datetime.html

thehomebrewnerd:
- Add infer_datetime_format param to to_datetime calls
- Closes #218 

Adds `infer_datetime_format` parameter to speed up internal `to_datetime` calls during type casting and checking if a column is a datetime column or not. Also removes `dropna` call from `col_is_datetime` util function which no longer appears to be necessary. This should also improve performance.

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -2,15 +2,18 @@
 
 Release Notes
 -------------
-.. Future Release
-  ==============
+Future Release
+==============
     * Enhancements
     * Fixes
     * Changes
+        * Add ``infer_datetime_format`` param to speed up ``to_datetime`` calls (:pr:`1016`)
     * Documentation Changes
     * Testing Changes
 
-.. Thanks to the following people for contributing to this release:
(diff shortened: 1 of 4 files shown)


thehomebrewnerd:
As a test to evaluate the performance improvement here, I create a dataframe with 10,000,000 rows and 2 columns: an integer id column and a column with randomly generated datetime strings (stored as `object` dtype in the dataframe).

Running the code on main, type inference took 22ms to complete. Using the code in this branch, type inference took only about 3ms to complete. Not sure how broadly this applies, but based on this limited test, this does seem to improve datetime inference speed significantly.

gsheni (approved):

PR #1023, opened 2021-06-25
Handle Numpy inputs to Woodwork

### LINKED ISSUE #501: Handle Numpy inputs to Woodwork
With the DataTables/DataColumns api, we allowed numpy arrays as inputs so that users wouldn't have to check the input type and do `pd.DataFrame(numpy_array)` before creating a DataTable.

With the Schema/Accessor api, there's no numpy accessor, so it likely doesn't make sense to allow creating a Schema from a numpy array directly. We should determine whether Woodwork should provide a utility function that converts a numpy array to a Pandas object with the Accessor already initialized or whether converting from numpy to pandas is left up to the user.

simha104:
- Handle Numpy inputs to Woodwork
- Closes #501


docs/source/release_notes.rst
```diff
@@ -10,6 +10,7 @@ Future Release
     * Changes
         * Remove lines in ``LogicalType.transform`` that raise error if dtype conflicts (:pr:`1012`)
         * Add ``infer_datetime_format`` param to speed up ``to_datetime`` calls (:pr:`1016`)
+        * Add handling of numpy inputs to Woodwork (:pr:`1023`)
```
gsheni, line 13:
```suggestion
        * Add support for numpy inputs in ``init_series`` (:pr:`1023`)
```

woodwork/tests/utils/test_accessor_utils.py
```diff
@@ -45,6 +46,19 @@ def test_init_series_with_invalid_type(sample_df):
             init_series(input_)
 
 
+def test_init_series_with_np_array(sample_series_pandas):
+    inputs = [np.array([['a', 'b'], ['b', 'c']]), np.array(['a', 'b', 'c', 'a'])]
+    for input_ in inputs:
+        if input_.ndim == 1:
+            series = init_series(input_)
+            series2 = init_series(sample_series_pandas)  # Sample series panda contains ['a','b','c','a']
+            assert series.equals(series2)
```
tamargrey, line 55: We should also test that the Woodwork typing information is as expected.

tamargrey: Also, as this test grows, it'll probably make sense to separate the multi dimensional array case into its own test. There's not really any shared code between the two tests, so it's better to separate them out so we can potentially see two different bugs at the same time (as opposed to a bug from the first element masking one from the second)

tamargrey (changes requested):

woodwork/tests/utils/test_accessor_utils.py
```diff
@@ -47,16 +47,16 @@ def test_init_series_with_invalid_type(sample_df):
...
-            series = init_series(input_)
-            series2 = init_series(sample_series_pandas)  # Sample series panda contains ['a','b','c','a']
-            assert series.equals(series2)
-        else:
-            error_message = f'np.ndarray input must be 1 dimensional. Current np.ndarray is {input_.ndim} dimensional'
-            with pytest.raises(ValueError, match=error_message):
-                init_series(input_)
+    series = init_series(np.array(['a', 'b', 'c', 'a']))
+    series2 = init_series(sample_series_pandas)  # Sample series panda contains ['a','b','c','a']
+    assert series.equals(series2) and series.ww.logical_type == series2.ww.logical_type and series.ww.semantic_tags == series2.ww.semantic_tags
```
tamargrey, line 52: can you make these separate assertions on separate lines? Makes it a little easier to read

tamargrey (approved):
lgtm!

PR #1025, opened 2021-06-29
Save inferred datetime format

### LINKED ISSUE #1021: Save inferred datetime format on Datetime Logical Type
- As a user, I wish I could use Woodwork to save the inferred datetime format to my Logical Type, and have it be accessible later on. This would apply when the user did not specify the datetime format.
- We should infer based on the first **n** non-null rows. We then select the most common format detected and use that as a format.
- We don't need to support columns with mixed datetime formats.
- Original suggestion here: https://github.com/alteryx/woodwork/issues/898#issuecomment-840147503

```python
>>> import pandas as pd
>>> fmt = pd.core.tools.datetimes.guess_datetime_format("1/05/2021")
>>> fmt
'%m/%d/%Y'
>>> pd.to_datetime("1/05/2021", format=fmt)
Timestamp('2021-01-05 00:00:00')
```
```python
import pandas as pd

def infer_datetime_format(dates, n=100):
    first_n = dates.dropna().head(n)
    fmts = dates.map(pd.core.tools.datetimes.guess_datetime_format)
    mode_fmt = fmts.mode().loc[0] # select first most common format
    return mode_fmt
```

tuethan1999:
- Adds datetime format inference in transform if a format is not set
- Closes #1021


woodwork/utils.py
```diff
@@ -442,3 +442,15 @@ def _update_progress(start_time, current_time, progress_increment,
...
         return new_progress
+
+
+def _infer_datetime_format(dates, n=100):
+    """Helper function to infer the datetime format of the first n non-null rows of a series
+    Args:
+        dates (Series): Series of string or datetime string to guess the format of
+        n (int): the maximum number of nonnull rows to sample from the series
+    """
+    first_n = dates.dropna().head(n)
```
thehomebrewnerd, line 453: Should we take the first n or sample the data randomly? If the data happens to be sorted by date, we might not have a good representation of all the dates present which could lead to incorrect inference.

thehomebrewnerd:
Once this is working as expected, can we do some performance profiling of this against the code in main which added the `infer_datetime_format=True` parameter to the `pd.to_datetime` calls? 

If I'm looking at this correctly we will always be specifying a format in the `to_datetime` calls, and if that is the case, I'm not sure inclusion of `infer_datetime_format` is needed or provides any benefit. I think it would be good to compare inference time with this code to the current code to understand.

docs/source/release_notes.rst
```diff
@@ -10,12 +10,13 @@ Future Release
     * Changes
         * Remove lines in ``LogicalType.transform`` that raise error if dtype conflicts (:pr:`1012`)
         * Add ``infer_datetime_format`` param to speed up ``to_datetime`` calls (:pr:`1016`)
+        * Adds datetime format inference in transform if a format is not set (:pr:`1025`)
```
thehomebrewnerd, line 13: I think I'd call this an enhancement instead of a change and say something like `Store inferred datetime format on Datetime logical type instance`

woodwork/tests/utils/test_utils.py
```diff
@@ -447,3 +448,9 @@ def test_col_is_datetime():
     for input, expected in list(zip(inputs, expected_values)):
         actual = col_is_datetime(input)
         assert actual is expected
+
+
+def test_infer_datetime_format(datetimes):
+    for series in datetimes:
+        fmt = _infer_datetime_format(series)
+        assert fmt == '%m/%d/%Y'
```
thehomebrewnerd, line 456: Does this also work reliably if there is a time included? Might want to test for that too.

woodwork/utils.py
gsheni: randomly sampling might seem like a better approach, but talking to Max, we should do the first N rows allows for consistency. A dataset tested multiple times returns the inferred datetime format. We can solve this in the future with a random seed, although for Dask we might always want to do first N rows.

thehomebrewnerd: Fair point. 

As a follow up, what happens if we infer incorrectly based on the first n? Does our initialization break completely and we require users to specify the format? If it does break, is the error meaningful? (I'm also not sure if we can have the same issue with the code in main, FWIW).

gsheni: @thehomebrewnerd for the question `if we infer incorrectly`, Would this be if `pd.to_datetime` was passed the incorrect format, and raised an error? WW code says that it would raise an error
https://github.com/alteryx/woodwork/blob/278aa96fbe78f0e8ed406dcf2d290bb0c015b4b3/woodwork/type_sys/utils.py#L28
- I think either it raises an error, and asks the user to specify OR we coerce the incorrect values to pd.NAT (while warning the user).

thehomebrewnerd: @gsheni The code you reference is in the Datetime inference function, but we won't be using a format string then unless the user has specified it. We don't get the inferred format string that gets saved on the logical type until after our type inference has run and we attempt to convert the column to a `datetime64` dtype by calling the transform function.

Is is possible that our inference function infers that the column is a datetime, and then fails during dtype conversion if the format returned from the first n rows doesn't match the format that was identified by pandas during the first inference function call?

I'll just have to dig into this more to understand. I think it might be odd behavior if our inference correctly identified a column as datetime and then type conversion failed because we incorrectly specified the wrong format by looking at only the first n rows to get the format string. I'm not sure if that's possible though. I'll test it out...

thehomebrewnerd:
Here is a notebook that can be used to test out performance of various approaches to calling `pd.to_datetime`. This isn't using Woodwork, but I will create a similar notebook using WW and post it here as well, when I have it done.

[Datetime Perf Testing - Pandas.ipynb.zip](https://github.com/alteryx/woodwork/files/6742279/Datetime.Perf.Testing.-.Pandas.ipynb.zip)

thehomebrewnerd:
Here is a notebook that can be used to test inference times with Woodwork. Below is a brief summary of some initial result obtained initializing Woodwork in different ways on a dataframe with 10,000,000 rows:

Code in `main` branch:
`df.ww.init()`: 2.97 ms
`df.ww.init(logical_types={'dates': Datetime})`: 2.94 ms
`df.ww.init(logical_types={'dates': Datetime(datetime_format="%m/%d/%Y")})`: 3.30 ms

Current code in `save-inferred-datetime-format` branch:
`df.ww.init()`: 2.97 ms
`df.ww.init(logical_types={'dates': Datetime})`: 2.77 ms
`df.ww.init(logical_types={'dates': Datetime(datetime_format="%m/%d/%Y")})`: 2.98 ms

Not sure how definitive this is, but so far indications are that performance is similar.

[Datetime Perf Testing - Woodwork.ipynb.zip](https://github.com/alteryx/woodwork/files/6742601/Datetime.Perf.Testing.-.Woodwork.ipynb.zip)

docs/source/release_notes.rst
```diff
@@ -5,6 +5,7 @@ Release Notes
 Future Release
 ==============
     * Enhancements
+        *Store inferred datetime format on Datetime logical type instance (:pr:`1025`)
```
gsheni, line 8:
```suggestion
        * Store inferred datetime format on Datetime logical type instance (:pr:`1025`)
```

woodwork/logical_types.py
```diff
@@ -173,21 +178,22 @@ def __init__(self, datetime_format=None):
...
             try:
                 if dd and isinstance(series, dd.Series):
                     name = series.name
-                    series = dd.to_datetime(series, format=self.datetime_format, infer_datetime_format=True)
+                    series = dd.to_datetime(series, format=self.datetime_format)
                     series.name = name
                 elif ks and isinstance(series, ks.Series):
                     series = ks.Series(ks.to_datetime(series.to_numpy(),
                                                       format=self.datetime_format,
                                                       infer_datetime_format=True),
```
jeff-hernandez, line 193: I noticed we remove `infer_datetime_format=True` in `dd.to_datetime` and `pd.to_datetime`. Do we also want to remove `infer_datetime_format=True` in `ks.to_datetime` or is this needed?

jeff-hernandez (commented):
Looking good! Just had a question about the `infer_datetime_format` parameter.

woodwork/logical_types.py
tuethan1999: nope, it's not. good catch!

woodwork/deserialize.py
```diff
@@ -82,15 +82,19 @@ def _typing_information_to_woodwork_table(table_typing_info, validate, **kwargs)
...
             cat_dtype = col['physical_type']['cat_dtype']
             if table_type == 'pandas':
                 cat_object = pd.CategoricalDtype(pd.Index(cat_values, dtype=cat_dtype))
             else:
                 cat_object = pd.CategoricalDtype(pd.Series(cat_values))
-            category_dtypes[col_name] = cat_object
+            col_type = cat_object
+        elif table_type == 'koalas' and col_type == 'object':
+            col_type = 'string'
+        category_dtypes[col_name] = col_type
```
thehomebrewnerd, line 97: Minor point, but this dictionary is no longer being used to store just category dtypes, so I think it would be more clear what is happening if we update the name to something like `column_dtypes`

woodwork/utils.py
```diff
@@ -442,3 +442,18 @@ def _update_progress(start_time, current_time, progress_increment,
         callback_function(progress_increment, new_progress, total, unit, elapsed_time)
 
         return new_progress
+
+
+def _infer_datetime_format(dates, n=100):
```
thehomebrewnerd, line 447: Just wondering if 100 rows is a large enough default? For datasets with tens of thousands of rows, using only 100 rows seems like a really small sample. We can always leave this as it is for now and update later if it becomes a problem though.

woodwork/tests/utils/test_read_file.py
```diff
@@ -42,12 +43,15 @@ def test_read_file_validation_control(mock_validate_accessor_params, sample_df_p
 @pytest.mark.parametrize(
     "filepath, exportfn, kwargs, pandas_nullable_fix",
     [
-        ("sample.csv", ("to_csv", {"index": False}), {}, True),
-        ("sample.csv", ("to_csv", {"index": False}), {"content_type": "csv"}, True),
+        ("sample.csv", ("to_csv", {"index": False}), {"logical_types": {
+            'signup_date': Datetime()}}, True),
+        ("sample.csv", ("to_csv", {"index": False}), {"content_type": "csv", "logical_types": {
+            'signup_date': Datetime()}}, True),
```
thehomebrewnerd, line 49: Why did we make the changes in this file? It's been long enough since I've looked at this, that I'm now struggling to recall the details.

tuethan1999: I think signup_date wasn't getting inferred as a Datetime. Revisiting it though, it doesn't seem to be a problem. I've removed it

woodwork/utils.py
tuethan1999: set the default size to be 10% of the rows!

jeff-hernandez (approved):
Looks good!

woodwork/tests/utils/test_read_file.py
```diff
@@ -113,6 +113,10 @@ def test_read_file(sample_df_pandas, tmpdir, filepath, exportfn, kwargs, pandas_
                       logical_types=kwargs.get('logical_types'),
                       semantic_tags=kwargs.get('semantic_tags'),
                       column_origins=kwargs.get('column_origins'))
+
+    if "to_csv":
+        df.ww.logical_types['signup_date'].datetime_format = None  # read_csv reads datetimes as strings and infers datetime during transform
```
thehomebrewnerd, line 118: This if block will always get executed because "to_csv" is truthy. What are we trying to do here?

tuethan1999: that was supposed to be checking the import func!

PR #1026, opened 2021-06-29
Handle pd.ExtensionArray inputs into Woodwork

### LINKED ISSUE #570: Handle pd.ExtensionArray inputs into Woodwork
With the DataTables/DataColumns api, we allowed `pd.ExtensionArrays` as inputs so that users wouldn't have to manually convert to a `pd.Series` before creating a DataColumn.

With the Schema/Accessor api, there's no ExtensionArray accessor, so we won't be able to allow creating an initialized column from a ExtensionArray directly. We should determine whether Woodwork should provide a utility function that converts an ExtensionArray to a series object with the accessor already initialized or whether converting from an ExtensionArray to series is left up to the user.

This issue is closely related to #501 and can likely share a common solution. Also, the proposed series creation utility function discussed in #554 could possibly be extended to provide this functionality.

simha104:
- Handle pd.ExtensionArray inputs into Woodwork
- Closes #570


woodwork/tests/utils/test_accessor_utils.py
```diff
@@ -38,6 +38,22 @@ def test_init_series_valid_conversion_specified_ltype(sample_series):
     assert series.ww.semantic_tags == set()
 
 
+def test_init_series_with_pd_extension_array():
+    extension_categories = pd.Categorical([1, 2, 3])
+    series = init_series(extension_categories)
+    series2 = init_series(pd.Series([1, 2, 3], dtype='category'))
```
tamargrey, line 44: It might be nice to have a more descriptive variable name for `series2`. Maybe something like `extension_series` that way the assertions are a little more clear on why we want to compare the two series

tamargrey:
We need to update the docstring for `init_series` to indicate that ExtensionArrays and numpy arrays are valid inputs. Maybe something like:

> Initializes Woodwork typing information for a Series, numpy.ndarray or pd.api.extensions.ExtensionArray ... 

as well as adding it to the list of options in the `series` parameter

woodwork/accessor_utils.py
```diff
@@ -9,11 +9,11 @@
 
 def init_series(series, logical_type=None, semantic_tags=None,
                 use_standard_tags=True, description=None, metadata=None):
-    """Initializes Woodwork typing information for a Series, returning a new Series. The dtype
+    """Initializes Woodwork typing information for a series, numpy.ndarray or pd.api.extensions.ExtensionArray, returning a new Series. The dtype
```
tamargrey, line 12: This line is a bit too long. A good guide is that you don't want to have to scroll to see the whole line if you have the screen split or something (an upper limit is probably 100 characters, but you'll see we're not perfect about sticking to that all the time).

You can split it up into multiple lines and just even out the rest of this section of the docstring.

woodwork/accessor_utils.py
```diff
@@ -9,11 +9,11 @@
...
     of the returned series will be converted to match the dtype associated with the LogicalType.
 
     Args:
-        series (pd.Series, dd.Series, or ks.Series): The original series from which to create
+        series (pd.Series, dd.Series, ks.Series, numpy.ndarray or pd.api.extensions.ExtensionArray): The original series from which to create
```
tamargrey, line 16: this line should also be split up and combined with the one below it.

tamargrey (commented):
looking good! just a bit of nitpicking about line length

tamargrey (approved):
lgtm!

PR #1035, opened 2021-07-01
Raises warning when user adds column with index tags

### LINKED ISSUE #900: setitem allows adding column with index tags
`WoodworkTableAccessor`'s setitem does not check that `index` or `time_index` tags are not present on a woodwork column being added to the DataFrame. This means that a user can manually add a woodwork column with the `index` or `time_index` tag to the DataFrame and no index validation will occur because the column accessor does not provide any restrictions around those tags. The result of this is that the table's index or time index gets set, possibly invalidating the schema.

```python
import woodwork as ww
import pandas as pd 

sample_df.ww.init()
assert sample_df.ww.index is None

new_col = pd.Series([1,1,1,1])
new_col.ww.init(semantic_tags='index')

# setitem with the new woodwork col
sample_df.ww['new_col'] = new_col

# problematic because a woodwork column can get the index tag without any validation
assert sample_df.ww.index == 'new_col'
assert not ww.is_schema_valid(sample_df, sample_df.ww.schema)

```

We should check that the index or time index tags are not present on any woodwork columns being passed to the DataFrame. We might want to consider removing the tags and raising a warning in these cases instead of throwing an error.

simha104:
- Raises warning when user adds column with index tags
- Closes #900


woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -2117,6 +2118,28 @@ def test_setitem_invalid_input(sample_df):
...
 
+def test_setitem_indexed_column(sample_df):
+    sample_df_with_index = sample_df
+    sample_df_with_index.ww.init()
+    sample_df.ww.init()
+    sample_df_with_index.ww.set_index('id')
+    inputs = [sample_df, sample_df_with_index]
+
+    # Testing two cases: The dataframe has no index, The dataframe already had an index that was removed
+    for input_ in inputs:
```
tamargrey, line 2129: lets split the two cases into separate tests. I'd say it's only worth doing this kind of format if the code is exactly the same from input to input. But once we start having to add if statements for each input, we should put them in separate tests. (or, if the tests still seem similar enough to be in one test function, just put one after another)

In general, I'd stick with separate tests unless you realize you're about to copy and paste a bunch of code.

woodwork/table_accessor.py
```diff
@@ -193,6 +194,10 @@ def __setitem__(self, col_name, column):
         if not isinstance(column, series):
             raise ValueError('New column must be of Series type')
 
+        if column.ww.schema is not None and 'index' in column.ww.semantic_tags:
+            warnings.warn("Cannot allow adding column with index tags. Tag removed", SetItemIndexWarning)
```
tamargrey, line 198: bit of a nitpick, but `SetItemIndexWarning` makes me think that we're warning users that we are setting the column as an index. Let's call this something like `IndexTagRemovedWarning`.

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -2118,26 +2118,34 @@ def test_setitem_invalid_input(sample_df):
...
+
+    col = sample_df.ww.pop('id')
+
+    warning = "Cannot allow adding column with index tags. Tag removed"
+
+    with pytest.warns(IndexTagRemovedWarning, match=warning):
+        sample_df.ww['id'] = col
+
+    assert sample_df.ww.index is None
+    assert ww.is_schema_valid(sample_df, sample_df.ww.schema)
```
tamargrey, line 2148: nice idea to check this!

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -2118,26 +2118,34 @@ def test_setitem_invalid_input(sample_df):
         df.ww['signup_date'] = df.signup_date
 
 
-def test_setitem_indexed_column(sample_df):
-    sample_df_with_index = sample_df
-    sample_df_with_index.ww.init()
+def test_setitem_indexed_column_on_indexed_dataframe(sample_df):
```
tamargrey, line 2121: I think this test could use an additional case (you can just put it after the code that's already here) where the index of another column is still present, like:

```python
    sample_df.ww.init(index='id')
    col = sample_df.ww.pop('email')
    col.ww.init(semantic_tags='index')
```

and then after we do `sample_df.ww['id'] = col`, we want to confirm `sample_df.ww.index == 'id'` and that the index tag is no longer in `email`.

tamargrey (commented):
I just remembered an additional behavior that we need to make sure is happening here. 

If you look at `TableSchema.remove_semantic_tags` theres a comment `If the index is removed, reinsert any standard tags not explicitly removed`. This is because we remove any standard tags (`numeric` and `categorical`) if a column is made into an index. So if we're removing the index tag from a column and adding it to the TableSchema, we need to make sure that those standard tags are getting added back in (if column.ww.use_standard_tags is True).

I'm not sure that the implementation here will do that (`column.ww.remove_semantic_tags` uses the ColumnSchema version of this method, which does not have that logic), so let's add a test to confirm what the current behavior is, and if the standard tags are not being replaced on the column, add that behavior in. (I'd suggest looking a bit at the behavior of standard tags first if you need, but you can check which logical types have `standard_tags` in `logical_types.py`

woodwork/table_accessor.py
```diff
@@ -197,6 +197,9 @@ def __setitem__(self, col_name, column):
         if column.ww.schema is not None and 'index' in column.ww.semantic_tags:
             warnings.warn("Cannot allow adding column with index tags. Tag removed", IndexTagRemovedWarning)
             column.ww.remove_semantic_tags('index')
+            logical_type = column.ww.logical_type
+            if hasattr(logical_type, 'standard_tags'):
```
tamargrey, line 201: https://github.com/alteryx/woodwork/blob/main/woodwork/logical_types.py#L24-L29

the default if a LogicalType has no specific standard tags is an empty set! So you can just reference `logical_type.standard_tags` without checking that they're present, I think!

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -2131,9 +2131,13 @@ def test_setitem_indexed_column_on_indexed_dataframe(sample_df):
 
     assert sample_df.ww.index is None
     assert ww.is_schema_valid(sample_df, sample_df.ww.schema)
+    expected_tags = {}
+    if hasattr(sample_df.ww.logical_types['id'], 'standard_tags'):
+        expected_tags = sample_df.ww.logical_types['id'].standard_tags
```
tamargrey, line 2136: It's often times preferable to have our expected values be a hardcoded value, because if there was also a bug in how we got the standard tags or logical types or something, we might very well not catch it because we use very similar lines of code in the setitem iteself.

There's no hard and fast rule, but for smaller things like a set of semantic tags, we often times write it out explicitly like we do in this test: https://github.com/alteryx/woodwork/blob/main/woodwork/tests/schema/test_table_schema_init.py#L294

So in this case, I think it'd be `assert sample_df.ww['id'].ww.semantic_tags == {'numeric'}`

tamargrey: Especially because we want to be testing logical types here that have standard tags. `id` will be `Integer`, so it will have the `numeric` standard tag, but unless we write that out, there's no way of seeing that as a developer reading this test.

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -2146,6 +2150,17 @@ def test_setitem_indexed_column_on_unindexed_dataframe(sample_df):
...
+    assert sample_df.ww['id'].ww.semantic_tags == expected_tags
+
+    sample_df.ww.set_index('id')
+    col = sample_df.ww.pop('email')
+    col.ww.init(semantic_tags='index')
+    sample_df.ww['email'] = col
+    assert sample_df.ww.index == 'id'
+    assert 'index' not in sample_df.ww['email'].ww.semantic_tags
```
tamargrey, line 2163: In the same vein as the other comment about writing out the set of expected tags, we can kill a few birds with one stone by checking that `sample_df.ww['email'].ww.semantic_tags` matches a set of semantic tags explicitly written out. 

Saying `sample_df.ww['email'].ww.semantic_tags == set()` does the same thing but also would catch any unexpected tags that made their way in somehow

tamargrey: Also, if you set email's logical type at init with `sample_df.ww.init(logical_types={'email':'Categorical')` then email will also have standard semantic tags that can be tested here

woodwork/table_accessor.py
```diff
@@ -198,8 +198,7 @@ def __setitem__(self, col_name, column):
             warnings.warn("Cannot allow adding column with index tags. Tag removed", IndexTagRemovedWarning)
             column.ww.remove_semantic_tags('index')
             logical_type = column.ww.logical_type
-            if hasattr(logical_type, 'standard_tags'):
-                column.ww.add_semantic_tags(logical_type.standard_tags)
+            column.ww.add_semantic_tags(logical_type.standard_tags)
```
tamargrey, line 201: one more behavior thing - we allow users to set `use_standard_tags` for a column to `False` which will stop a column that would normally have the `numeric` tag, for example, from having it. I believe that with this implementation, `column.ww.use_standard_tags` could be False but it'd still get the standard tags.

One option would be to add onto the current implementation, and only perform this line if `column.ww.use_standard_tags is True`, but I also wanted to present another idea that would allow us to combine this line and the two above into one line:

`column.ww.set_semantic_tags(column.ww.semantic_tags - {'index'})` 

`set_semantic_tags` also takes care of the standard tag logic (https://github.com/alteryx/woodwork/blob/main/woodwork/column_schema.py#L179) which is why this lets us simplify here. 

Whichever implementation you go with, we should add an additional test function that checks that we don't have the standard tags if `use_standard_tags` is False. (use_standard_tags can be set at woodwork init for dataframe or series)

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -2117,6 +2118,62 @@ def test_setitem_invalid_input(sample_df):
...
+    warning = "Cannot allow adding column with index tags. Tag removed"
+
+    with pytest.warns(IndexTagRemovedWarning, match=warning):
+        sample_df.ww['id'] = col
+
+    assert sample_df.ww.index is None
+    assert ww.is_schema_valid(sample_df, sample_df.ww.schema)
+    assert sample_df.ww['id'].ww.semantic_tags == {'numeric'}
+
+    sample_df.ww.init(logical_types={'email': 'Categorical'})
```
tamargrey, line 2152: Defintiely a nitpick, but lets add some whitespace to break up the flow from here to line 2158. the dataframe setup should be one block, then removing the column and setting it up, then when we add it back to the dataframe, and then the assertions

tamargrey (commented):
getting close!

woodwork/table_accessor.py
```diff
@@ -193,6 +194,10 @@ def __setitem__(self, col_name, column):
         if not isinstance(column, series):
             raise ValueError('New column must be of Series type')
 
+        if column.ww.schema is not None and 'index' in column.ww.semantic_tags:
+            warnings.warn(f'Cannot add {"index"} tag on {col_name} directly to the DataFrame. The {"index"} tag has been removed from {col_name}. To set this column as a Woodwork index, please use df.ww.set_index', IndexTagRemovedWarning)
```
tamargrey, line 198: don't need the brackets around `"index"`

tamargrey (approved):
lgtm!

PR #1043, opened 2021-07-09
Infer logical type of null columns as Unknown

### LINKED ISSUE #1044: Use Unknown logical type for null columns
Based off discussion in #1038, it makes more sense to infer columns that are entirely null (i.e. made entirely up of `None`, `np.nan`, or `pd.NA`) as `Unknown` logical type instead of `Double` or `Categorical`. This way, entirely null columns are all handled in a consistent manner.

frances-h:
- Columns that are entirely null (made up of `None`, `np.nan`, or `pd.NA`) are inferred as `Unknown` instead of `Double`/`Categorical`.
- Closes #1044


woodwork/type_sys/type_system.py
```diff
@@ -261,6 +261,10 @@ def infer_logical_type(self, series):
         if ks and isinstance(series, ks.Series):
             series = series.head(100000).to_pandas()
 
+        # Special case: if the entire column is Null or NaN, use the Unknown logical type
+        if all(series.isnull()):
```
tamargrey, line 265: does this catch the column of `'nan'` strings? Or are we going to handle that in a separate PR?

frances-h: It only catches true null values, not the string nan representations. That's going to get handled in a separate PR once we figure out how we want to handle them. This does mean that the 'nan' -> Datetime bug is still present. I could add the check here that discards Datetime as a logical type if the entire column is NaT and that would at least avoid the issue for now.

tamargrey: cool, just double checking! Don't need to handle it right now, but maybe it's worth this PR having its own issue outlining the feature change to have fully null columns be Unknown

woodwork/tests/type_system/test_ltype_inference.py
```diff
@@ -134,6 +134,17 @@ def test_unknown_inference(strings):
...
 
+def test_unknown_inference_all_null(nulls):
+    dtypes = ['object', 'string', 'category', 'datetime64[ns]']
+    if ks and isinstance(nulls[0], ks.Series):
+        dtypes = get_koalas_dtypes(dtypes)
+
+    for series in nulls:
+        for dtype in dtypes:
+            inferred_type = ww.type_system.infer_logical_type(series.astype(dtype))
+            assert isinstance(inferred_type, Unknown)
```
tamargrey, line 145: we should also confirm that `inferred_type.transform(series)` works as expected for all of these series (though maybe that should go in `test_logical_types.py`)

woodwork/tests/type_system/test_ltype_inference.py
```diff
@@ -133,6 +133,11 @@ def test_unknown_inference(strings):
             inferred_type = ww.type_system.infer_logical_type(series.astype(dtype))
             assert isinstance(inferred_type, Unknown)
 
+            # verify transform works and column is still null
+            transform = inferred_type.transform(series.astype(dtype))
```
tamargrey: I think that the `series.astype(dtype)` will be redundant -- `LogicalType.transform` will perform the astype call if needed

tamargrey (approved):
lgtm once changes are merged in from main!

gsheni (approved):

PR #1047, opened 2021-07-12
Add Infer Email Address function

### LINKED ISSUE #945: Add Infer Email Address function
- As a user, I wish I could use Woodwork to automatically infer if a column is an Email Address Logical Type (if the majority of the not-NaN values are email address)
- We can use this regex, which captures the majority of situations
  - https://emailregex.com/
- We should sample with 10K rows (first 10K rows, dropna)

#### Code Example

```python
import pandas as pd
import numpy as np
  
data = pd.Series(['fl@alteryx.com', 'bad@email.com', 'no_email', np.nan])

data.str.match(pat=r"(^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+$)")
```
```
Out[2]:
0     True
1     True
2    False
3      NaN
dtype: object

```

davesque:
- Adds a type inference function for email addresses.
- Closes #945 .

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -5,6 +5,8 @@ Release Notes
 Future Release
 ==============
     * Enhancements
+        * Add support for automatically inferring the ``EmailAddress`` logical
+          type (:pr:`1047`)
     * Fixes
     * Changes
         * Entirely null columns are now inferred as the Unknown logical type (:pr:`1043`)
@@ -13,7 +15,7 @@ Future Release
     * Testing Changes
 
     Thanks to the following people for contributing to this release:
-    :user:`frances-h`, :user:`jeff-hernandez`
+    :user:`frances-h`, :user:`jeff-hernandez`, :user:`davesque`
 
 v0.5.0 Jul 7, 2021
 ==================
--- a/woodwork/config.py
+++ b/woodwork/config.py
@@ -2,7 +2,6 @@ CONFIG_DEFAULTS = {
     'categorical_threshold': 10,
     'numeric_categorical_threshold': -1,
     'email_inference_regex': r"(^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+$)",
-    'email_inference_threshold_ratio': 0.5,
 }
 
 
--- a/woodwork/config.py
+++ b/woodwork/config.py
@@ -1,6 +1,8 @@
 CONFIG_DEFAULTS = {
     'categorical_threshold': 10,
     'numeric_categorical_threshold': -1,
+    'email_inference_regex': r"(^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+$)",
+    'email_inference_threshold_ratio': 0.5,
 }
 
 
--- a/woodwork/tests/type_system/conftest.py
+++ b/woodwork/tests/type_system/conftest.py
@@ -129,7 +129,7 @@ def datetimes(request):
 def pandas_emails():
     return [
         pd.Series(['fl@alteryx.com', 'good@email.com', 'boaty@mcboatface.com', 'foo@bar.com']),
-        pd.Series(['fl@alteryx.com', 'good@email.com', 'not_an_email', np.nan]),
+        pd.Series(['fl@alteryx.com', 'good@email.com', 'boaty@mcboatface.com', np.nan]),
     ]
 
 
@@ -145,6 +145,30 @@ def koalas_emails(pandas_emails):
 
 @pytest.fixture(params=['pandas_emails', 'dask_emails', 'koalas_emails'])
 def emails(request):
+    return request.getfixturevalue(request.param)
+
+
+# Email Inference Fixtures
+@pytest.fixture
+def bad_pandas_emails():
+    return [
+        pd.Series(['fl@alteryx.com', 'no_an_email', 'good@email.com', 'foo@bar.com']),
+        pd.Series(['fl@alteryx.com', 'b☃d@email.com', 'good@email.com', np.nan]),
+    ]
+
+
+@pytest.fixture
+def bad_dask_emails(bad_pandas_emails):
+    return [pd_to_dask(series) for series in bad_pandas_emails]
+
+
+@pytest.fixture
+def bad_koalas_emails(bad_pandas_emails):
+    return [pd_to_koalas(series) for series in bad_pandas_emails]
+
+
+@pytest.fixture(params=['bad_pandas_emails', 'bad_dask_emails', 'bad_koalas_emails'])
+def bad_emails(request):
     return request.getfixturevalue(request.param)
 
 
--- a/woodwork/tests/type_system/conftest.py
+++ b/woodwork/tests/type_system/conftest.py
@@ -152,8 +152,10 @@ def emails(request):
 @pytest.fixture
 def bad_pandas_emails():
     return [
-        pd.Series(['fl@alteryx.com', 'no_an_email', 'good@email.com', 'foo@bar.com']),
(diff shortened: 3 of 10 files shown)


davesque:
@gsheni @thehomebrewnerd This isn't complete yet.  Needs docs and also there are a few test failures that still need to be worked out.

I'm actually not sure if some of the work I did was supposed to be included in this PR or another one.  There were a number of different test failures that resulted from adding a new registration for the email inference function.  I fixed most of them but the remaining ones are related to failing testing fixtures that are being downloaded from S3.  Was I supposed to add the registration for `inference_functions.email_address_func` in this ticket or was that supposed to be part of another ticket?

thehomebrewnerd:
> Was I supposed to add the registration for `inference_functions.email_address_func` in this ticket or was that supposed to be part of another ticket?

@davesque Yes, I think the registration of the new inference function should be included in this PR, and some new test failures are to be expected. Without looking into the details, I suspect the S3 related test failures have to do with the fact that the saved serialized data in S3 has the `email` column in `sample_df` with a logical type of `Unknown`, but after you register the new inference function this column is getting inferred as `EmailAddress`, so when we compare the deserialized table to the table we get through initialization on the dataframe, these columns don't match. If my guess here is correct, we will need to update the serialized data in S3, and bump our serialization schema version.

davesque:
@thehomebrewnerd Yeah, that's exactly what's happening as far as I can tell.  All of the failures that popped up were related to testing fixtures that needed to be updated with some "email" column now having the `EmailAddress` logical type.

davesque:
@thehomebrewnerd I'm a bit confused by the coverage failures.  A lot of them seem to be relating to lines that I didn't touch.

thehomebrewnerd:
> @thehomebrewnerd I'm a bit confused by the coverage failures. A lot of them seem to be relating to lines that I didn't touch.

@davesque Looks like things are fine. The way coverage works is by assembling coverage reports from multiple unit tests runs with different dependencies, so generally you will want to make sure all the tests have completed before worrying too much about the coverage reports.

I'll take a closer look at the PR tomorrow, but in the meantime it looks like you also need to sign the contributor license agreement.

docs/source/release_notes.rst
```diff
@@ -17,6 +17,8 @@ v0.5.0 Jul 7, 2021
     * Enhancements
         * Add support for numpy array inputs to Woodwork (:pr:`1023`)
         * Add support for pandas.api.extensions.ExtensionArray inputs to Woodwork (:pr:`1026`)
+        * Add support for automatically inferring the ``EmailAddress`` logical
+          type (:pr:`1047`)
```
thehomebrewnerd, line 23: All the updates should be placed in the `Future Release` section above, which will need to be uncommented since this is the first PR for the next release. Also, I think we should be fine to have this as the only release notes entry - if you want to mention the config param, you can do so here as well, but we've tended to keep these entries fairly brief.

woodwork/config.py
```diff
@@ -1,6 +1,7 @@
 CONFIG_DEFAULTS = {
     'categorical_threshold': 10,
     'numeric_categorical_threshold': -1,
+    'email_inference_regex': r"(^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+$)",
```
thehomebrewnerd, line 4: We have a documentation guide that explains the various Woodwork configuration options. I think we should also add a brief description to that guide that mentions this option: https://woodwork.alteryx.com/en/stable/guides/setting_config_options.html#Available-Config-Settings

You can do that by updating the `setting_config_options.ipynb` file in the `docs/source/guides` folder.

thehomebrewnerd (changes requested):
Looking pretty good. Just a couple docs related items to clean up and one additional suggestion for the test fixtures.

woodwork/tests/type_system/conftest.py
```diff
@@ -124,6 +124,54 @@ def datetimes(request):
...
+def emails(request):
+    return request.getfixturevalue(request.param)
+
+
+# Email Inference Fixtures
+@pytest.fixture
+def bad_pandas_emails():
+    return [
+        pd.Series(['fl@alteryx.com', 'no_an_email', 'good@email.com', 'foo@bar.com']),
+        pd.Series(['fl@alteryx.com', 'b☃d@email.com', 'good@email.com', np.nan]),
```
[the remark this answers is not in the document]
davesque: Sure, good idea.

docs/source/release_notes.rst
davesque: Sorry, this resulted from me improperly resolving a merge conflict.  I had originally placed them in the future releases section.

woodwork/type_sys/inference_functions.py
```diff
@@ -74,5 +86,26 @@ def timedelta_func(series):
...
+        sample_match_method = sample.str.match
+    except (AttributeError, TypeError):
+        # This can happen either when the inferred dtype for a series is not
+        # compatible with the pandas string API (AttributeError) *or* when the
+        # inferred dtype is not compatible with the string API `match` method
+        # (TypeError)
+        return False
+    matches = sample_match_method(pat=regex)
+
+    return matches.sum() == matches.count()
```
gsheni, line 107: Does this mean that all 10K row have to match the email regex?

davesque: More specifically, it means that all valid (non-NA) rows have to match.  I had a brief discussion with @thehomebrewnerd about this.  He said we've tended to do an all or nothing type approach with the inference functions.  My interpretation of that was that all of the valid sampled records have to match the email address regex.  Originally, I had it so that only half of the sampled records needed to match and that percentage was configurable.  Happy to discuss it further if you have other ideas.

davesque: Also, it appears that the behavior is slightly different depending on if the series came from a dask or koalas data set.  If it came from dask, then only 10k rows from the first partition in the dataset are considered (regardless of whether or not that partition actually has 10k rows).  If it came from koalas, then 10k rows are sampled from the result of a `.head(100000)` call.

davesque: Actually, I had thought to mention the above difference to @thehomebrewnerd and ask if it seemed like that difference in behavior was okay.  Or is it a little confusing?  Not sure how I feel about it.  For what it's worth, I did have to structure some of the test series a specific way to account for the fact that the test series are split into two partitions when they're sent to dask.  So if you're trying to check if a bad record exists in some test series, it better be in the first partition of the series otherwise `TypeSystem.infer_logical_type` discards.

docs/source/release_notes.rst
```diff
@@ -27,9 +29,12 @@ v0.5.0 Jul 7, 2021
         * Add ``infer_datetime_format`` param to speed up ``to_datetime`` calls (:pr:`1016`)
         * The default logical type is now the ``Unknown`` type instead of the ``NaturalLanguage`` type (:pr:`992`)
         * Add pandas 1.3.0 compatibility (:pr:`987`)
+    * Documentation Changes
+    * Testing Changes
 
     Thanks to the following people for contributing to this release:
-    :user:`jeff-hernandez`, :user:`simha104`, :user:`tamargrey`, :user:`thehomebrewnerd`, :user:`tuethan1999`
+    :user:`jeff-hernandez`, :user:`simha104`, :user:`tamargrey`,
+    :user:`thehomebrewnerd`, :user:`tuethan1999`
```
thehomebrewnerd, line 37: I think we should revert these changes, so we are only updating the Future Release section.

thehomebrewnerd (commented):
One more update needed to the release notes and then looks like there is a new merge conflict to fix.

docs/source/release_notes.rst
davesque: Ahh, sorry didn't catch that.  Thanks!

thehomebrewnerd (approved):
Looks good!

Issue #1045, opened 2021-07-09
Issues with large data sets and distributed computing


davesque:
This issue is related to another discussion that happened on #945 .

I noticed that some of the logic in woodwork that uses the Pandas API is also designed to accommodate Koalas or Dask objects.  For example, this utility func used by the datetime inferer:
https://github.com/alteryx/woodwork/blob/0d5c5c1ff160d35e6c50232ba0184fec860a8343/woodwork/type_sys/utils.py#L12-L35

There's an issue with the code on these lines which might exist in other places as well:
https://github.com/alteryx/woodwork/blob/0d5c5c1ff160d35e6c50232ba0184fec860a8343/woodwork/type_sys/utils.py#L15-L16

The problem is that a partitioned data set (as would be seen with Dask or Koalas) might be way too large to fit into the memory of whatever process is driving a pipeline.  So doing the `.to_pandas()` call might cause the python interpreter (or Kubernetes pod) to get OOM killed.

It may eventually become necessary for us to be more mindful of considerations involving large data sets on distributed compute clusters.  A lot of operations that are trivial to perform on a single core become difficult in a distributed setting.

Here's a link to a section in the Dask documentation regarding best practices for Dask: https://docs.dask.org/en/latest/dataframe-best-practices.html

That article has lots of good general advice on how to think about distributed data sets (e.g. cost of shuffling operations, partition size considerations, network communcation overhead, etc.).  Of course, some of the things the article mentions may be Dask-specific.  Dask has some inefficiencies baked in relating to its use of Pandas as an in-memory data representation and also its use of Python as the implementation language for its scheduler.  Some of the advice given is somewhat related to those things and wouldn't be as much of a concern with other frameworks.

davesque:
More thoughts on this: https://github.com/alteryx/woodwork/pull/1048#discussion_r672469659

thehomebrewnerd:
Support for Dask dataframes is being dropped.

PR #1048, opened 2021-07-12
Add outlier detection to Woodwork

### LINKED ISSUE #1029: Add outlier detection from EvalML Data Check
- As a user, I wish I could use Woodwork to detect outliers
  - https://github.com/alteryx/evalml/blob/main/evalml/data_checks/outliers_data_check.py#L137
- This issue will involve moving the code from [here](https://github.com/alteryx/evalml/blob/main/evalml/data_checks/outliers_data_check.py#L137) into the TableAccessor.
- EvalML will then leverage Woodwork's function during the data check

tamargrey:
- Adds two methods `df.ww.box_plots_dict` and `df.ww.outliers_dict` that calculate outliers for each column with the IQR method.
- Closes #1029 

To Do:

- [x] Expand nan tests
- [x] Expand edge case/ error tests
- [x] Move to be methods on column accessor
- [x] Write docstrings
- [x] Add section to statistical_insights guide
- [x] Determine if we want to sort outliers
- [x] Determine if we want to add a multiplier param for IQR bounds calculation
- [x] Determine handling of partial input dictionaries - do we only calculate for the specified columns?


woodwork/tests/accessor/test_statistics.py
```diff
@@ -830,3 +833,262 @@ def test_get_numeric_value_counts_in_range(input_series, expected):
...
+def test_calculate_iqr_bounds_with_quantiles(outliers_df_pandas):
+    expected_low = 8.125
+    expected_high = 83.125
+
+    series = outliers_df_pandas['has_outliers']
+
+    q1, q3 = np.percentile(series, [25, 75])
+    quantiles = {0.25: q1, 0.75: q3, 0.0: 100000}
+
+    low, high = _calculate_iqr_bounds(outliers_df_pandas['has_outliers'], quantiles=quantiles)
```
davesque, line 847: It seems like the quantiles that are being passed into `_calculate_iqr_bounds` here will be ignored since the function is designed to calculate them from the given series if one is provided.  Is this test case meant to probe that behavior?  If so, it can maybe be a bit more targeted and we probably wouldn't need to test it twice.  If we're not trying to test that behavior, was it maybe forgotten in this case that the function behaves that way?

davesque (commented):
Cool, nice to see this coming together.  I had a lot of thoughts/comments/questions!  Sorry that there are so many!  I think I'll just submit this review to you as is and get your thoughts and then maybe do a final pass once we've gone back and forth about the things I mentioned.

woodwork/column_accessor.py
```diff
@@ -357,6 +361,62 @@ def set_semantic_tags(self, semantic_tags):
...
+            dict[str -> float,list[number]]: a dictionary containing outlier values and their corresponding indexes.
+                The following elements will be found in the dictionary:
+
+                - low_values (list[float, int]): the values of the lower outliers
+                - high_values (list[float, int]): the values of the upper outliers
+                - low_indices (list[int]): the corresponding index values for each of the lower outliers
+                - high_indices (list[int]): the corresponding index values for each of the upper outliers
+        """
+        if self._schema is None:
+            _raise_init_error()
```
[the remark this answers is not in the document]
tamargrey: Yeah, it goes in front of every method on the column and table accessors to avoid the unexpected errors that can show up when you try and make Woodwork calls before initializing Woodwork. 

I don't think we've ever considered adding decorators, though I do think this could be a good use for that. @simha104 is working on an issue that could end up touching a lot of these lines in order to add schema validation, so that might be a good time to change all of these into decorators! (@thehomebrewnerd any thoughts?)

woodwork/column_accessor.py
```diff
@@ -357,6 +361,62 @@ def set_semantic_tags(self, semantic_tags):
...
+
+        if self._schema.is_numeric:
+            return _get_outliers_for_column(self._series, low_bound=low_bound, high_bound=high_bound)
```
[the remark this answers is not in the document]
tamargrey: I did intend that behavior (though I forgot to indicate that in the docstrings, so that definitely needs to be updated if we keep this behavior), but I'm not totally sold on whether this should be the behavior.

If we want to add anything, I'd probably prefer a warning over an error. I don't think we have a perfect example of this kind of behavior in the column accessor. The closest is in mutual information where we ignore incompatible columns, but we're performing the calculations for the whole table there, so it's that certain columns are missing rather than nothing happens when the method is called.

Right now I'm leaning towards adding a Warning.

woodwork/statistics_utils.py
```diff
@@ -343,6 +343,136 @@ def _get_value_counts(dataframe, ascending=False, top_n=10, dropna=False):
...
+        # We've already removed nans and converted to pandas
+        outliers_dict = _get_outliers_for_column(series, low_bound, high_bound, convert_series=False)
+
+    return {'low_bound': low_bound,
+            'high_bound': high_bound,
+            'quantiles': quantiles,
+            **outliers_dict}
+
+
+def _get_outliers_for_column(series, low_bound=None, high_bound=None, convert_series=True):
```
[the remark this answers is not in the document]
tamargrey: yep, happy to add that

woodwork/column_accessor.py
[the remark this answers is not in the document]
thehomebrewnerd: I think the decorator approach is better than what we are doing currently, and I like that idea.

If we could auto-magically have this applied to our methods somehow that might be even better so we don't always have to remember to add the code/decorator to new methods that get added. That might be more problematic/difficult than it's worth though...

woodwork/statistics_utils.py
```diff
@@ -343,6 +343,136 @@ def _get_value_counts(dataframe, ascending=False, top_n=10, dropna=False):
...
+    """
+    if convert_series:
+        if dd and isinstance(series, dd.Series):
+            series = series.compute()
+        if ks and isinstance(series, ks.Series):
+            series = series.to_pandas()
+
+        series = series.dropna()
+        if series.shape[0] == 0:
+            return
```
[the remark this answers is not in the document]
tamargrey: This is a holdover from evalML's implementation (https://github.com/alteryx/evalml/blob/main/evalml/data_checks/outliers_data_check.py#L148) that probably guided my choice to return `None` if the column is not numeric. To me, both cases are some form of "this column cannot have outliers calculated for it", so I'd like to handle them in the same way.

woodwork/statistics_utils.py
```diff
@@ -343,6 +343,136 @@ def _get_value_counts(dataframe, ascending=False, top_n=10, dropna=False):
...
+
+    if low_bound is None or high_bound is None:
+        low_bound, high_bound = _calculate_iqr_bounds(series)
```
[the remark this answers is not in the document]
tamargrey: I agree that it can be confusing. Part of why I didn't want to allow partial inputs with the bounds (or with `Q1` and `Q3` down on line 406) was to avoid the added complexity that would come with being able to handle any combination of inputs. It also seems like a big assumption to have one bound be a user input and have the other bound be from the IQR.

Right now, I'm thinking it would be better to avoid these assumptions and raise errors if a user has passed in partial inputs, letting them know that they either need to pass in all the required stats or remove them all and let Woodwork calculate them.

Another thought I had is that we might want to catch a situation where a user has passed in bounds where `low_bound > high_bound`.

woodwork/tests/accessor/test_statistics.py
tamargrey: whoopse, definitely a case of me forgetting how this function behaves - will remove the series!

woodwork/tests/conftest.py
```diff
@@ -608,3 +608,31 @@ def koalas_datetimes(pandas_datetimes):
...
+    return dd.from_pandas(outliers_df_pandas, npartitions=2)
+
+
+@ pytest.fixture()
+def outliers_df_koalas(outliers_df_pandas):
+    ks = pytest.importorskip('databricks.koalas', reason='Koalas not installed, skipping')
+    return ks.from_pandas(outliers_df_pandas)
+
+
+@ pytest.fixture(params=['outliers_df_pandas', 'outliers_df_dask', 'outliers_df_koalas'])
```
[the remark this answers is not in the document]
tamargrey: oops, fixing!

woodwork/statistics_utils.py
```diff
@@ -343,6 +343,136 @@ def _get_value_counts(dataframe, ascending=False, top_n=10, dropna=False):
...
+
+    low_series = series[series < low_bound]
+    high_series = series[series > high_bound]
+
+    return {
+        "low_values": low_series.tolist(),
+        "high_values": high_series.tolist(),
+        "low_indices": low_series.index.tolist(),
+        "high_indices": high_series.index.tolist()
```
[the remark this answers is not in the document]
tamargrey: This format is originally coming from evalML. Maybe @freddyaboulton knows if this is for a specific purpose or if there's a reason on their end that it shouldn't change?

docs/source/guides/statistical_insights.ipynb
```diff
@@ -152,6 +152,33 @@
...
+   "outputs": [],
+   "source": [
+    "total = df.ww['total']\n",
+    "box_plot_dict = total.ww.box_plot_dict()\n",
+    "\n",
+    "print('high bound: ', box_plot_dict['high_bound'])\n",
+    "print('low_bound: ', box_plot_dict['low_bound'])\n",
+    "print('quantiles: ', box_plot_dict['quantiles'])\n",
+    "print('number of low outliers: ', len(box_plot_dict['low_values']))\n",
+    "print('number of high outliers: ', len(box_plot_dict['high_values']))"
```
thehomebrewnerd, line 180: If I'm reading this correctly, the output is suggesting we have 1922 low outliers and 31016 high outliers. This seems like a lot of outliers - over 8% of the total data. Is this expected?

docs/source/release_notes.rst
```diff
@@ -7,6 +7,7 @@ Future Release
     * Enhancements
         * Add support for automatically inferring the ``EmailAddress`` logical
           type (:pr:`1047`)
+        * Add outlier detection and box plots to ``WoodworkColumnAccessor`` (:pr:`1048`)
```
thehomebrewnerd, line 10: Should we be more specific about the dictionaries here? Especially for box plot, as this could be interpreted as adding a graphical plot instead of just the data.

woodwork/statistics_utils.py
```diff
@@ -343,6 +343,136 @@ def _get_value_counts(dataframe, ascending=False, top_n=10, dropna=False):
     return val_counts
 
 
+def _calculate_iqr_bounds(series=None, quantiles=None):
+    """Calculates bounds for outlier detection using the 1.5*IQR method.
+
+    Args:
+        series (pd.Series, optional): Data used to calculate first and third quartiles used to calcualte
+            the interquartile range. If present, quartiles will be calculated even if passed in.
```
thehomebrewnerd, line 351: Clarification suggestion...
`If present, any values passed in for quantiles will be replaced with quartiles calculated from this series.`

woodwork/statistics_utils.py
```diff
@@ -343,6 +343,136 @@ def _get_value_counts(dataframe, ascending=False, top_n=10, dropna=False):
...
+        quantiles (dict[float->float], optional): Quantiles that can be used to calculate the interquartile
+            range. The keys of the dictionary should be the quantile floating point value.
```
thehomebrewnerd, line 353: Should this also be quartiles? If we don't pass in quartiles, won't we get key errors when we try to index with `quantiles[0.25]` and `quantiles[0.75]`? Unless of course, the quantiles happen to contain 0.25 and 0.75 values.

Sorry to nit-pick a docstring on a private function, but maybe the docstring should indicate that if quantiles are passed without a series, the values need to include 0.25 and 0.75 values.

woodwork/column_accessor.py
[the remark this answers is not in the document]
davesque: @thehomebrewnerd Yeah, in principle it's nice to avoid the possibility that someone might fail to add the magic decorator to a new method.  In practice, I always have concerns that automagical meta class stuff is hard to understand and predict for most developers.  So I often seek some kind of middle ground.  I think the decorator thing is nice because you can call it something like `require_init` and that's pretty descriptive and hard to miss.  The trade-off is that you have to remember to include it.  In any case, we'd need some way to distinguish methods that should have that behavior from those that shouldn't.  That could require some other machinery which might be just as much code as the simple decorator application.

woodwork/column_accessor.py
```diff
@@ -357,6 +361,62 @@ def set_semantic_tags(self, semantic_tags):
...
+            using the IQR method.
+
+        Returns:
+            dict[str -> float,list[number]]: a dictionary containing outlier values and their corresponding indexes.
+                The following elements will be found in the dictionary:
+
+                - low_values (list[float, int]): the values of the lower outliers
+                - high_values (list[float, int]): the values of the upper outliers
+                - low_indices (list[int]): the corresponding index values for each of the lower outliers
+                - high_indices (list[int]): the corresponding index values for each of the upper outliers
```
thehomebrewnerd, line 384: Maybe there is a reason for this outside of Woodwork, but would it be better to return the results in a dictionaries with the index as the key and the value as, well, the value, instead of separate lists?

woodwork/statistics_utils.py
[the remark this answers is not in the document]
davesque: > Right now, I'm thinking it would be better to avoid these assumptions and raise errors if a user has passed in partial inputs, letting them know that they either need to pass in all the required stats or remove them all and let Woodwork calculate them.

Yeah, I like your suggestion here.

woodwork/statistics_utils.py
thehomebrewnerd, line 464: Should this be outside of the `if convert_series` block?

woodwork/column_accessor.py
thehomebrewnerd: I now see @davesque also suggested alternate return types below. I don't have any strong opinions either way, just highlighting for discussion.

woodwork/statistics_utils.py
tamargrey: it should be inside this block, though maybe `convert_series` isn't the best name. It's more about not repeating both the conversion to pandas as well as the removal of nans if we're calling the outlier function from within the box plot function, which will have already performed those operations. 

How about something like `transform_series`?

woodwork/column_accessor.py
tamargrey: Currently the outlier values aren't sorted in any particular order, but it's very likely that we will want them returned from this function sorted by value, in which case, having them in a list (or series) will be useful

woodwork/column_accessor.py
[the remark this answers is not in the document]
davesque: This seems like a particular instance of a larger question that I often encounter which is: how does a library (or language) prefer to represent error conditions?  I usually try to avoid concrete, primitive values such as `-1` or `None`.  The `None` built-in can be useful to represent optional arguments for which no value was provided, but I don't like it as a return value to represent some kind of failure.  That's because the `None` value is totally opaque and not descriptive.  Also, it can potentially be propagated downstream in a computation if it's not handled properly where it might have a different meaning entirely (e.g. `bool(None) is False`) or where it's just confusing and out of context.  For these reasons, I usually just prefer raising an exception and preferably a custom `Exception` subclass.  A custom subclass can be as descriptive (has its own name, error message) and specific (is only used in a very particular situation) as you want.  It also *always* has the intended effect of halting program execution until it bubbles up into a handler that is expecting it.  Would be happy to have a teams chat about this with anyone who's interested in discussing the larger question further.

woodwork/statistics_utils.py
thehomebrewnerd: Got it. I think `transform_series` might be a bit more indicative of what is happening.

woodwork/column_accessor.py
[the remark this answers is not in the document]
davesque: But yeah, maybe I need to examine the context of this issue more carefully.  If it's like you say, and there's some blanket operation happening for an entire table, it would make sense to not have the entire operation fail with an exception.  In that case, ideally, I would hope that there's some way to handle the exception that indicates some operation is not supported for a column and just put a `pass` in the `except` block.  It might also make sense to emit a warning like you say so the user is aware that something weird happened internally.  But, if it's the sort of thing that might actually happen pretty often on average, it might not be worth emitting a warning.

docs/source/guides/statistical_insights.ipynb
```diff
@@ -152,6 +152,33 @@
     "mi = df.ww.mutual_information(include_index=True)\n",
     "mi[mi['column_1'].isin(['order_product_id']) | mi['column_2'].isin(['order_product_id'])]"
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Outlier Detection\n",
+    "- `Seriew.ww.outlier_dict`\n",
```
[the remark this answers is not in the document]
tamargrey: fixing!

docs/source/guides/statistical_insights.ipynb
tamargrey: Looking at all of the numeric columns in the retail dataframe, they all have somewhat tight quartiles, but a big enough standard deviation that the number of outliers makes some sense to me. But I think we don't necessarily know what would be expected in the data here. 

This does, however, make me want to confirm that the results for a normally distributed column of data are expected. Would it make sense to add that instead of (or maybe in addition to) the retail example here?

docs/source/release_notes.rst
tamargrey: Thoughts on 

> Add ability to calculate outliers and the statistical info required for box and whisker plots to WoodworkColumnAccessor

woodwork/statistics_utils.py
tamargrey: You bring up a good point about when we should be using "quantiles" vs "quartiles". We want users to be able to pass through any quantiles they'd like (0.0 and 1.0 being the main ones since, in theory, those could be used to build a slightly different box plot than one that uses the high and low bounds as whiskers).

However, we need at least the quartiles 0.25 and 0.75 to perform the calculations in these functions. I think that the docstrings here probably need to be a lot more clear about what the minimum requirements are, and what values will be returned.

docs/source/release_notes.rst
thehomebrewnerd: Sounds good!

woodwork/statistics_utils.py
tamargrey: agree that this could be clearer. Here's another suggestion

> If present, any values passed in for quantiles will be ignored. Instead,
            the first and third quartiles will be calculated from this series.

woodwork/statistics_utils.py
tamargrey: changing to `transfom_series`

docs/source/guides/statistical_insights.ipynb
tamargrey: Added a section that creates a normally distributed column of data the same size as the retail dataset, and the number of outliers pretty closely matches what we'd expect from normally distributed data (http://www.cs.uni.edu/~campbell/stat/normfact.html), so I'm guessing it's jus that the retail data isn't normally distributed.

woodwork/statistics_utils.py
thehomebrewnerd: Works for me!

woodwork/column_accessor.py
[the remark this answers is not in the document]
tamargrey: In removing the `outliers_dict` method to just have a single `box_plot_dict` method, was able to move the `is_numeric` check to exist as a part of the statistics_utils.py `_get_box_plot_info_for_column` helper.

This check will now raise a `TypeError` for non numeric columns

woodwork/statistics_utils.py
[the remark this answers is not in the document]
tamargrey: Changed this case where the series is fully null (or empty) to return a box plot dict with no outliers and `np.nan` values for the numeric elements of the dict:

```
{
            'low_bound': np.nan,
            'high_bound': np.nan,
            'quantiles': {0.0: np.nan, 0.25: np.nan, 0.5: np.nan, 0.75: np.nan, 1.0: np.nan},
            "low_values": [],
            "high_values": [],
            "low_indices": [],
            "high_indices": []
        }
```

woodwork/statistics_utils.py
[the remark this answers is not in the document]
tamargrey: In removing the separate `outliers_dict` call, there is no longer a path that allows users to pass in bounds and get the values outside of the bounds. This situation is no longer possible, as we only care now about whether the correct quantiles were passed in.

How we handle partial quantiles is now also different:
- if there is a quantiles dict passed but q1 and q3 (0.25 and 0.75) are not present, then a `ValueError` is thrown
- if no quantiles dict is passed, the quantiles are generated
- if the correct quantiles are passed, they are used to calculate IQR

woodwork/tests/accessor/test_statistics.py
tamargrey: by removing the `outliers_dict` method, there's no longer a reason to have separate `series` and `quantiles` params, so I've removed the helper function and the logic now lives in `_get_box_plot_info_for_column` since the logic for whether we need quantiles or not has to exist earlier in the box plot logic anyway.

woodwork/statistics_utils.py
tamargrey: without separate `outlier_dict` and `box_plot_dict` methods, `transform_series` variable is no longer necessary, so it's been removed.

woodwork/column_accessor.py
```diff
@@ -357,6 +358,41 @@ def set_semantic_tags(self, semantic_tags):
             _raise_init_error()
         self._schema._set_semantic_tags(semantic_tags)
 
+    def box_plot_dict(self, quantiles=None):
+        """Gets the information necessary to create a box and whisker plot with outliers for a numeric column
+        using the IQR method.
+
+    Args:
+        series (Series): Data for which the box plot and outlier information will be gathered.
+            Will be used to calculate quantiles if none are provided.
```
thehomebrewnerd, line 367: Should be removed since this isn't an argument for this method. Maybe update the quantiles docstring info to indicate they will be computed if not provided.

woodwork/tests/accessor/test_column_accessor.py
```diff
@@ -166,7 +166,8 @@ def test_error_accessing_methods_before_init(sample_series):
         'remove_semantic_tags': [{'new_tag'}],
         'reset_semantic_tags': None,
         'set_logical_type': ['Integer'],
-        'set_semantic_tags': [{'new_tag'}]
+        'set_semantic_tags': [{'new_tag'}],
+        'box_plot_dict': [{}],
```
thehomebrewnerd, line 170: Doesn't matter much, but can this just be `'box_plot_dict': None,`?

thehomebrewnerd (approved):
Just a couple small things, but otherwise looks good to me.

woodwork/tests/accessor/test_column_accessor.py
tamargrey: yep, definitely can!

woodwork/column_accessor.py
tamargrey: removed the series arg and updated quantiles

woodwork/statistics_utils.py
```diff
@@ -343,6 +343,104 @@ def _get_value_counts(dataframe, ascending=False, top_n=10, dropna=False):
...
+    if not series.ww._schema.is_numeric:
+        raise TypeError('Cannot calculate box plot statistics for non-numeric column')
+
+    if quantiles and not isinstance(quantiles, dict):
+        raise TypeError('quantiles must be a dictionary.')
+
+    if dd and isinstance(series, dd.Series):
+        series = series.compute()
+    if ks and isinstance(series, ks.Series):
+        series = series.to_pandas()
```
davesque, line 385: I'm again thinking about my previous company where we were always trying to be extremely mindful of what sorts of algorithms and techniques were viable in a distributed computing (big data) context.  I've been told we're still a ways off from really targeting that scale.  So I don't think it's worth holding up this PR.  However, I thought I'd offer some more thoughts on it here to hopefully get us thinking about these questions.

Both Dask and Koalas offer `quantile` methods that use efficient approximations to obtain quantile figures in-cluster.  If `series` were a Dask or Koalas object, we could call out to the respective method.  It might also be required to do something like `.compute()` or `.to_pandas()` to force the computation to execute inside this method in order to get quantile figures that can be used for filtering later in the method<sup>[1](#f1)</sup>.  But then the filtering itself could also be done through the Dask or Koalas API and we could eventually just pass back the resulting Dask or Koalas object.  Then, the caller could decide if it wants to do the computation right then and there to get the concrete outliers or wait until a later time.

In a big data context, it would really be much more ideal if this method just returned the filtered series object (which would be either a Dask or Koalas object) instead of forcing the values into a python list.  But I guess that includes some details about serialization required to move things between services.  I'm less familiar with that part of things right now so I don't have much to say about it.  In general, I think it's just better if methods are about one thing and don't include implicit details about the context from which they're being called (like requirements of eventual serialization of the result).

<b id="f1">1</b>: As a side note, if we do make a synchronous cluster call mid-function, that's not that great.  It's always better if you can defer the distributed computation until some time when it's expected that you might call out to a cluster and wait around a long time for a result to come back.  If you're mixing synchronous cluster calls into a utility function like this, that will make the function a lot less useful.

davesque (approved):
Looks good.  I just left a few more thoughts about how things should eventually look in the context of big data.  But I don't want to hold up the PR over it.

woodwork/statistics_utils.py
tamargrey: would it be worth adding some of these concerns to #1045? We can start compiling some of the areas where we're maybe not the happiest with how we've handled dask and koalas there.

davesque: Yeah, I suppose so.  I'll add a comment there referring to this one.

PR #1052, opened 2021-07-13
Update to preserve column ordering during select

### LINKED ISSUE #1039: TableSchema.columns as OrderedDictionary
- As an EvalML developer, I would like Woodwork TableSchema to preserve column order of the underlying DataFrame so that DataFrame behavior is more closely emulated.

It would be nice if the following test would pass.  It would allow us to remove a function from our Woodwork compatibility library.  As a caveat, this test may or may not pass as the order of the resulting columns is not guaranteed to be consistent.

#### Code Example

```python
import pytest
@pytest.mark.parametrize(
    "ww_type,pandas_type",
    [("Integer", "int"),
     ("Double", "float"),
     ("Datetime", "datetime"),
     ("NaturalLanguage", "string"),
     ("Categorical", "category"),
     ("Boolean", "boolean")]
)
def test_ww_column_order(ww_type, pandas_type):

    df = ww.demo.load_retail(nrows=100, init_woodwork=False)
    df.ww.init(name="retail")
    ww_schema_column_order = [x for x in df.ww.select([ww_type], return_schema=True).columns.keys()]
    pandas_column_order = [x for x in df.select_dtypes(include=pandas_type).columns]
    assert ww_schema_column_order== pandas_column_order

```

thehomebrewnerd:
- Update to preserve column ordering during select
- Closes #1039 

This PR implements changes to `TableSchema._filter_cols` to use a list in place of a set, thus guaranteeing that the column order returned by the method will match the column order on the dataframe to which the schema belongs.

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -10,12 +10,13 @@ Future Release
     * Fixes
     * Changes
         * Entirely null columns are now inferred as the Unknown logical type (:pr:`1043`)
+        * ``TableAccessor.select`` method will now maintain dataframe column ordering (:pr:`1052`)
     * Documentation Changes
         * Add supported types to metadata docstring (:pr:`1049`)
     * Testing Changes
 
     Thanks to the following people for contributing to this release:
-    :user:`frances-h`, :user:`jeff-hernandez`, :user:`davesque`
+    :user:`frances-h`, :user:`jeff-hernandez`, :user:`davesque`, :user:`thehomebrewnerd`
 
 v0.5.0 Jul 7, 2021
 ==================
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -10,7 +10,7 @@ Future Release
     * Fixes
     * Changes
         * Entirely null columns are now inferred as the Unknown logical type (:pr:`1043`)
-        * ``TableAccessor.select`` method will now maintain dataframe column ordering (:pr:`1052`)
+        * ``TableAccessor.select`` method will now maintain dataframe column ordering in TableSchema columns (:pr:`1052`)
     * Documentation Changes
         * Add supported types to metadata docstring (:pr:`1049`)
     * Testing Changes
--- a/woodwork/table_schema.py
+++ b/woodwork/table_schema.py
@@ -471,14 +471,17 @@ class TableSchema(object):
                 col_name_matches.add(selector)
 
         cols_to_return = []
+        cols_seen = set()
         for col_name, col in self.columns.items():
             is_match = (type(col.logical_type) in ltypes_used or
                         col.semantic_tags.intersection(tags_used) or
                         col_name in col_name_matches)
-            if include is not None and is_match and col_name not in cols_to_return:
+            if include is not None and is_match and col_name not in cols_seen:
                 cols_to_return.append(col_name)
-            elif exclude is not None and not is_match and col_name not in cols_to_return:
+                cols_seen.add(col_name)
+            elif exclude is not None and not is_match and col_name not in cols_seen:
                 cols_to_return.append(col_name)
+                cols_seen.add(col_name)
 
         return list(cols_to_return)
 
--- a/woodwork/table_schema.py
+++ b/woodwork/table_schema.py
@@ -470,15 +470,15 @@ class TableSchema(object):
             if col_names and selector in self.columns:
                 col_name_matches.add(selector)
 
-        cols_to_return = set()
+        cols_to_return = []
         for col_name, col in self.columns.items():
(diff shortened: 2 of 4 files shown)


freddyaboulton (approved):
@thehomebrewnerd Thank you for this! This meets EvalML's requirements. If you're worried about the performance hit of having to check inclusion in a list vs a set, maybe you can create a separate set to keep track of the column names we've seen before?

This implementation would work for us though!

woodwork/table_schema.py
```diff
@@ -470,15 +470,15 @@ def _filter_cols(self, include=None, exclude=None, col_names=False):
...
+        cols_to_return = []
         for col_name, col in self.columns.items():
             is_match = (type(col.logical_type) in ltypes_used or
                         col.semantic_tags.intersection(tags_used) or
                         col_name in col_name_matches)
-            if include is not None and is_match:
-                cols_to_return.add(col_name)
-            elif exclude is not None and not is_match:
-                cols_to_return.add(col_name)
+            if include is not None and is_match and col_name not in cols_to_return:
```
davesque, line 478: Mostly looks good.  Just one thought: not sure about the broader context of this but technically pandas does support columns with the same name.  Would doing this `not in` check here possibly cause issues with a dataset that has duplicate column names in the same table?

thehomebrewnerd: @davesque Although pandas does support duplicate column names, Woodwork does not, as the column name serves as the key by which we track the typing information. Any duplicates that occur here would be a result of a single column matching multiple select criteria, not because duplicate columns exist in the dataframe.

thehomebrewnerd:
> @thehomebrewnerd Thank you for this! This meets EvalML's requirements. If you're worried about the performance hit of having to check inclusion in a list vs a set, maybe you can create a separate set to keep track of the column names we've seen before?
> 
> This implementation would work for us though!

Thanks for the suggestion, I'll look into that approach a bit more.

woodwork/table_schema.py
```diff
@@ -470,15 +470,18 @@ def _filter_cols(self, include=None, exclude=None, col_names=False):
...
-            elif exclude is not None and not is_match:
-                cols_to_return.add(col_name)
+            if include is not None and is_match and col_name not in cols_seen:
+                cols_to_return.append(col_name)
+                cols_seen.add(col_name)
+            elif exclude is not None and not is_match and col_name not in cols_seen:
+                cols_to_return.append(col_name)
+                cols_seen.add(col_name)
 
         return list(cols_to_return)
```
tamargrey, line 486: now that `cols_to_return` is a list, is this line redundant?

thehomebrewnerd: Yep. Good catch. Updated.

docs/source/release_notes.rst
```diff
@@ -10,12 +10,13 @@ Future Release
     * Fixes
     * Changes
         * Entirely null columns are now inferred as the Unknown logical type (:pr:`1043`)
+        * ``TableAccessor.select`` method will now maintain dataframe column ordering (:pr:`1052`)
```
tamargrey, line 13: should this specify "maintain dataframe column ordering when ``return_schema` is True" or something?

tamargrey (commented):
I'm wondering if this change to `TableSchema_filter_cols` means that `_get_subset_df_with_schema` no longer needs `use_dataframe_order`? The only other place where `_get_subset_df_with_schema` gets called with `use_dataframe_order=True` is in drop, and I think that the columns passed in from there should match the dataframe column ordering. 

Might not be worth removing the keyword, because it could definitely come in handy later, but just wanted to bring this up.

docs/source/release_notes.rst
thehomebrewnerd: How about: `...maintain dataframe column ordering in TableSchema columns?`

tamargrey: that sounds good to me!

thehomebrewnerd:
> I'm wondering if this change to `TableSchema_filter_cols` means that `_get_subset_df_with_schema` no longer needs `use_dataframe_order`? The only other place where `_get_subset_df_with_schema` gets called with `use_dataframe_order=True` is in drop, and I think that the columns passed in from there should match the dataframe column ordering.
> 
> Might not be worth removing the keyword, because it could definitely come in handy later, but just wanted to bring this up.

Looks to me like you are correct about this. I don't think the `use_dataframe_order` argument is necessary now. `drop` should always maintain the order based on the way columns are generated, so the only other place this was used is in `select`.

I think we should remove this rather than leaving it in for some possible future use. I will update.

tamargrey (approved):
lgtm!

PR #1056, opened 2021-07-15
Add column origin attribute

### LINKED ISSUE #927: Add feature origin attribute
- As a user, I wish I could use Woodwork to keep track of the origin of my feature. 
- My feature/series/column could be a base feature in my original dataset, an engineered feature.
- This will be a primary attribute on ColumnScheme, similar to `description`. 

#### Code Example

```python
series = pd.Series(['red', 'blue', 'green', 'green', 'blue'])
series.ww.init(source='engineered')
series.ww.source = 'source'

df = ww.demo.load_retail(nrows=100, init_woodwork=False)
df.ww.init(name="retail", source={'order_id':'original', 'YEAR(order_date)': 'engineered'})
df.ww['order_date'].ww.source = 'source'

```
#### Open Questions
- Should we use [enumeration](https://docs.python.org/3/library/enum.html) define all the valid sources?
- What are the valid sources?
- Do these appropriately map to feature types? (From Featuretools)

frances-h:
- Add feature origin attribute to column schema
- Closes #927


gsheni:
@tamargrey agreed, that's a good parameter for `read_file`

woodwork/accessor_utils.py
```diff
@@ -45,6 +46,7 @@ def init_series(series, logical_type=None, semantic_tags=None,
                        semantic_tags=semantic_tags,
                        use_standard_tags=use_standard_tags,
                        description=description,
+                       origin=origin,
```
tamargrey, line 49: in `test_accessor_utils` theres a `test_init_series_all_parameters` test that could have `origin` added to it

woodwork/tests/schema/test_table_schema_init.py
```diff
@@ -379,6 +394,42 @@ def test_schema_col_descriptions_errors(sample_column_names, sample_inferred_log
...
+    origins = {
+        'invalid_col': 'not a valid column',
+        'signup_date': 'base'
+    }
+    err_msg = re.escape("column_origins contains columns that are not present in TableSchema: ['invalid_col']")
+    with pytest.raises(ColumnNotPresentError, match=err_msg):
+        TableSchema(sample_column_names, sample_inferred_logical_types, column_origins=origins)
+
+    origins = {
+        1: 'bad key',
```
tamargrey, line 425: Sorry!!  Just realized i said keys in my comment when I meant values 🤦‍♀️ The `1` should hopefully get caught since it's not a column in the dataframe (and we do actually want to allow non-string column names). I was thinking more like `{'id':1}` as a case to be checking.

tamargrey (approved):
lgtm!

PR #1058, opened 2021-07-15
Improve sample df

### LINKED ISSUE #999: Improve sample_df fixture
The `sample_df` test fixture is used in a large number of tests throughout Woodwork. This fixture could be improved to include more logical types, which would make our testing more robust. We should review and make sure the following logical types are adequately represented:
- Double with np.nan (not covered)
- Double without np.nan (not covered)
- Integer/NullableInteger (may be ok, but some test use age as Ordinal instead of integer)
- Boolean/NullableBoolean (only boolean with nulls included currently)
- Categorical (not explicitly covered with default inference)
- Datetime - with pd.NaT (not covered)
- Datetime - without pd.NaT (probably ok as is)
- Timestamp (? maybe not, since this won't work with Koalas)

tuethan1999:
- Updates the sample_df fixture to have more logical_type coverage. 
These additions include:
  - Double with np.nan 
  - Double without np.nan 
  - Integer/NullableInteger 
  - Boolean
  - Categorical
  - Datetime - with pd.NaT
- Closes #999


woodwork/serialize.py
```diff
@@ -207,5 +207,9 @@ def _create_archive(tmpdir):
 
 def save_orc_file(dataframe, filepath):
     from pyarrow import Table, orc
-    pa_table = Table.from_pandas(dataframe, preserve_index=False)
+    df = dataframe.copy()
```
gsheni, line 210: We should be careful with copy operations (especially with large datasets). Is there a way to avoid this?

woodwork/tests/accessor/test_statistics.py
```diff
@@ -613,6 +613,14 @@ def test_describe_with_include(sample_df):
     multi_params_df['full_name'].equals(sample_df.ww.describe()['full_name'])
 
 
+def test_pandas_nullable_integer_quantile_fix():
+    """Should fail when https://github.com/pandas-dev/pandas/issues/42626 gets fixed"""
```
gsheni, line 623:
```suggestion
    """Should fail until https://github.com/pandas-dev/pandas/issues/42626 gets fixed"""
```

woodwork/tests/accessor/test_statistics.py
```diff
@@ -603,17 +603,28 @@ def test_describe_with_include(sample_df):
...
+    expected = ['full_name', 'age', 'signup_date', 'datetime_with_NaT']
     for col_name in expected:
         assert col_name in multi_params_df.columns
     multi_params_df['full_name'].equals(col_name_df['full_name'])
     multi_params_df['full_name'].equals(sample_df.ww.describe()['full_name'])
 
 
+def test_pandas_nullable_integer_quantile_fix():
+    """Should fail when https://github.com/pandas-dev/pandas/issues/42626 gets fixed"""
+    if pd.__version__ != '1.3.0':
```
thehomebrewnerd, line 620: I assume this is just to get a test failure that prompts us to implement changes once this bug is fixed?

If so, and we leave this test in, maybe we should mark this `if` block with a `# pragma: no cover` so it doesn't cause coverage issues.

woodwork/tests/schema/test_table_schema_init.py
```diff
@@ -94,7 +92,7 @@ def test_check_logical_types_errors(sample_column_names):
         'phone_number': None,
         'age': None,
     }
-    error_message = re.escape("logical_types is missing columns that are present in TableSchema: ['is_registered', 'signup_date']")
+    error_message = re.escape("logical_types is missing columns that are present in TableSchema: ['boolean', 'categorical', 'datetime_with_NaT', 'double', 'double_with_nan', 'integer', 'is_registered', 'nullable_integer', 'signup_date']")
```
thehomebrewnerd, line 95: Either add the new column names to the `bad_logical_types_keys` dict or break this message into multiple lines as it is really long.

woodwork/tests/schema/test_table_schema_init.py
```diff
@@ -487,13 +474,7 @@ def test_use_standard_tags_from_dict(sample_column_names, sample_inferred_logica
...
     assert default_schema.use_standard_tags == {col_name: False for col_name in sample_column_names}
 
-    use_standard_tags = {'id': True,
-                         'full_name': False,
-                         'email': True,
-                         'phone_number': True,
-                         'age': False,
-                         'signup_date': True,
-                         'is_registered': False}
+    use_standard_tags = {'id': True, 'full_name': False, 'email': True, 'phone_number': True, 'age': False, 'signup_date': True, 'is_registered': False, 'double': False, 'double_with_nan': False, 'integer': False, 'nullable_integer': False, 'boolean': False, 'categorical': False, 'datetime_with_NaT': False}
```
thehomebrewnerd, line 477: Break this up into multiple lines

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -1317,7 +1317,7 @@ def test_dataframe_methods_on_accessor_to_pandas(sample_df):
         pd_df = sample_df.ww.compute()
     elif _is_koalas_dataframe(sample_df):
         pd_df = sample_df.ww.to_pandas()
-
+        pd_df.ww.init(name='woodwork', index='id')
```
thehomebrewnerd, line 1320: I don't think we want to do this here. This is testing that if we call .compute or .to_panads through the Woodwork access that we get back a dataframe with Woodwork already initialized. If we aren't getting back an initialized dataframe with koalas, I think that is pointing at another issue we need to investigate.

thehomebrewnerd (commented):
There is a lot to digest in this PR, but here are a few things I've noticed so far. I'd like to take a more thorough look in the morning with a fresh mind.

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -1733,9 +1701,11 @@ def test_select_return_schema(sample_df):
      (["Datetime"], "datetime"),
      (["Unknown", "EmailAddress"], "string"),
      (["Categorical"], "category"),
-     (["BooleanNullable"], "boolean")]
+     (["Boolean", "BooleanNullable"], "boolean")]
 )
 def test_select_retains_column_order(ww_type, pandas_type, sample_df):
+    if ks and isinstance(sample_df, ks.DataFrame) and (pandas_type == "category" or pandas_type == "string"):
```
thehomebrewnerd, line 1707: We can clean up this logic a bit:
`if _is_koalas_dataframe(sample_df) and pandas_type in ["category", "string"]:`

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -2433,32 +2406,16 @@ def test_accessor_types(sample_df):
...
-        'id': Integer.primary_dtype,
-        'full_name': Unknown.primary_dtype,
-        'email': EmailAddress.primary_dtype,
-        'phone_number': Unknown.primary_dtype,
-        'age': IntegerNullable.primary_dtype,
-        'signup_date': Datetime.primary_dtype,
-        'is_registered': BooleanNullable.primary_dtype,
-    }
+    correct_physical_types = {name: ltype.primary_dtype for name, ltype in sample_inferred_logical_types.items()}
+    if ks and isinstance(sample_df, ks.DataFrame):
```
thehomebrewnerd, line 2410: We are trying to get away from this pattern with the new helper functions. You can use `if _is_koalas_dataframe(sample_df):` here instead.

woodwork/tests/schema/test_table_schema.py
```diff
@@ -59,30 +59,13 @@ def test_schema_types(sample_column_names, sample_inferred_logical_types):
...
-        'id': "['numeric']",
-        'full_name': "[]",
-        'email': "[]",
-        'phone_number': "[]",
-        'age': "['numeric']",
-        'signup_date': "[]",
-        'is_registered': "[]",
-        'formatted_date': "[]",
-    }
+    correct_semantic_tags = {'id': "['numeric']", 'full_name': '[]', 'email': '[]', 'phone_number': '[]', 'age': "['numeric']", 'signup_date': '[]', 'is_registered': '[]', 'double': "['numeric']", 'double_with_nan': "['numeric']", 'integer': "['numeric']", 'nullable_integer': "['numeric']", 'boolean': '[]', 'categorical': "['category']", 'datetime_with_NaT': '[]', 'formatted_date': '[]'}
```
thehomebrewnerd, line 68: Break into multiple lines

woodwork/tests/schema/test_table_schema.py
```diff
@@ -264,17 +248,16 @@ def test_filter_schema_cols_exclude(sample_column_names, sample_inferred_logical
 
     filtered_log_type_string = schema._filter_cols(exclude='Unknown')
     filtered_log_type = schema._filter_cols(exclude=Unknown)
-    expected = {'id', 'age', 'signup_date', 'is_registered'}
+    expected = {'boolean', 'double', 'datetime_with_NaT', 'email', 'categorical', 'double_with_nan', 'signup_date', 'age', 'integer', 'nullable_integer', 'is_registered', 'id'}
```
thehomebrewnerd, line 251: Break into multiple lines

woodwork/tests/schema/test_table_schema.py
```diff
@@ -264,17 +248,16 @@ def test_filter_schema_cols_exclude(sample_column_names, sample_inferred_logical
...
     assert set(filtered_log_type) == expected
 
     filtered_semantic_tag = schema._filter_cols(exclude='numeric')
     assert 'age' not in filtered_semantic_tag
 
     filtered_multiple_overlap = schema._filter_cols(exclude=['Unknown', 'email'], col_names=True)
-    expected = ['id', 'age', 'signup_date', 'is_registered']
-    for col in filtered_multiple_overlap:
-        assert col in expected
+    expected = {'is_registered', 'double', 'nullable_integer', 'double_with_nan', 'categorical', 'datetime_with_NaT', 'age', 'integer', 'signup_date', 'boolean', 'id'}
```
thehomebrewnerd, line 259: Break into multiple lines

woodwork/tests/utils/test_concat.py
```diff
@@ -63,10 +64,10 @@ def test_concat_cols_combo_dfs(sample_df):
 
 
 def test_concat_cols_with_series(sample_df):
-    df = sample_df[['id', 'full_name', 'email', 'phone_number', 'age']]
-    s1 = sample_df['signup_date']
+    df = sample_df[['id', 'full_name', 'email', 'phone_number', 'age', 'signup_date', 'is_registered', 'double', 'double_with_nan', 'integer', 'nullable_integer', 'boolean']]
```
thehomebrewnerd, line 67: Break into multiple lines

woodwork/tests/utils/test_concat.py
```diff
@@ -310,7 +311,7 @@ def test_concat_cols_row_order(sample_df):
 
     df1 = sample_df.ww.loc[:, ['id', 'full_name']]
     df2 = sample_df.ww.loc[[2, 3, 0, 1], ['email', 'phone_number']]
-    df3 = sample_df.ww.loc[[3, 1, 0, 2], ['age', 'signup_date', 'is_registered']]
+    df3 = sample_df.ww.loc[[3, 1, 0, 2], ['age', 'signup_date', 'is_registered', 'double', 'double_with_nan', 'integer', 'nullable_integer', 'boolean', 'categorical', 'datetime_with_NaT']]
```
thehomebrewnerd, line 314: Break into multiple lines

thehomebrewnerd (commented):
A few more suggestions after doing a deeper dive. I guess the biggest thing is that we will have to decide what to do with the copy operation when writing an orc file.

woodwork/serialize.py
tuethan1999: I had the same thoughts but Schema.from_pandas(dataframe, preserve_index=False) does not work with Koalas dataframes for some reason

thehomebrewnerd: Will this be an issue with Koalas since Koalas doesn't support the `category` dtype?

tuethan1999: Koalas lack of  `category` support is fine since orc doesn't support the `category` dtype either. The problem is that nullable dtypes are parsed differently by `pyarrow.Table.from_pandas` and `pyarrow.Schema.from_pandas`. The schema parses them as null types whereas the table parses them as boolean/integer. The following test fails:

```
from pyarrow import Table, orc, Schema, DictionaryType, field, string
import databricks.koalas as ks

df = pd.DataFrame({'is_registered': pd.Series([pd.NA, False, True, True], dtype='boolean'),
                   'nullable_integer': pd.Series([pd.NA, -36, 49, 64], dtype='Int64')})
dataframe = ks.from_pandas(df)
schema = Schema.from_pandas(dataframe, preserve_index=False)
table_schema = Table.from_pandas(dataframe, preserve_index=False).schema

assert schema == table_schema
```
The workaround for this copy operation is getting kind of tedious. Do we know how frequently people use `to_orc`?

woodwork/tests/accessor/test_table_accessor.py
tuethan1999: Yup! I brought this up in a standup last week and @tamargrey advised me to ignore it for now, and investigate it in another PR. Should I put up a new issue or fix it in this PR?

woodwork/serialize.py
gsheni: to_orc isn't heavily used right now, so we can revisit this if needed (and do a simple implementation now).

woodwork/tests/accessor/test_table_accessor.py
thehomebrewnerd: I think it's ok to create a separate issue and PR for resolving this.

In the meantime, I think we would be better off to skip this test for Koalas with a message/comment explaining why we are skipping rather than reinitializing Woodwork since reinitializing basically means we aren't testing what this test was intended for.

woodwork/serialize.py
tuethan1999: Created issue #1072

woodwork/tests/accessor/test_table_accessor.py
tuethan1999: Created issue #1071!

woodwork/tests/accessor/test_statistics.py
```diff
@@ -453,21 +453,21 @@ def test_describe_accessor_method(describe_df):
...
     for ltype in nullable_numeric_ltypes:
-        expected_vals = pd.Series({
+        expected_vals = pd.Series({  # fix when https://github.com/pandas-dev/pandas/issues/42626 gets resolved
             'physical_type': ltype.primary_dtype,
             'logical_type': ltype(),
             'semantic_tags': {'numeric', 'custom_tag'},
-            'count': 7,
-            'nunique': 6,
+            'count': 7.0,
+            'nunique': 6.0,
```
thehomebrewnerd, line 461: Are these changes and the changes to `min` and `max` necessary here and below?

woodwork/tests/accessor/test_statistics.py
```diff
@@ -453,21 +453,21 @@ def test_describe_accessor_method(describe_df):
     # Test numeric columns with nullable ltypes
     numeric_data = describe_df[['numeric_col']]
     for ltype in nullable_numeric_ltypes:
-        expected_vals = pd.Series({
+        expected_vals = pd.Series({  # fix when https://github.com/pandas-dev/pandas/issues/42626 gets resolved
```
thehomebrewnerd, line 456: Is this comment and the one below still relevant?

woodwork/tests/accessor/test_statistics.py
tuethan1999: nope!

woodwork/tests/accessor/test_statistics.py
tuethan1999: nope!

woodwork/tests/accessor/test_statistics.py
```diff
@@ -464,7 +464,7 @@ def test_describe_accessor_method(describe_df):
             'mean': 20.857142857142858,
             'mode': 10,
             'std': 18.27957486220227,
-            'min': 1,
+            'min': 1.0,
```
thehomebrewnerd, line 467: Super minor nit-pick, but does this need to be changed?

woodwork/tests/accessor/test_statistics.py
```diff
@@ -604,17 +604,28 @@ def test_describe_with_include(sample_df):
     assert len(semantic_tags_df.columns) == 2
 
     logical_types_df = sample_df.ww.describe([Datetime, BooleanNullable])
-    assert 'signup_date', 'is_registered' in logical_types_df.columns
-    assert len(logical_types_df.columns) == 2
+    for c in ['signup_date', 'is_registered', 'datetime_with_NaT']:
+        assert c in logical_types_df.columns
+    assert len(logical_types_df.columns) == 3
```
thehomebrewnerd, line 609: Not a big deal, but I think we could simplify these three lines to:
```
assert set(logical_type_df.columns) == {'signup_date', 'is_registered', 'datetime_with_NaT'}
```

thehomebrewnerd (approved):
Just a couple minor things to fix if you want, but they are not things that need to hold up merging.

Issue #1051, opened 2021-07-13
Change categorical inference to look at sample of 10K rows and 80% non-unique


gsheni:
- As a user, I wish I could use Woodwork to infer Categorical Logical Type. 
- This inference function would be based on a 10K sample, and 80% of the rows being non-unique. 
  - (Rather than the current implementation which looks at string length).

davesque:
> ...and 80% of the columns being unique.

@gsheni Did you mean rows instead of columns?  Also, do you mean that 80% of rows being unique would mean it's not categorical?  I usually think of a categorical column as having a value for each row that is chosen from a relatively small number of possible unique values.

gsheni:
@davesque Yes you are correct, I updated the issue. It should be a 10K rows sample (remove nans) + 80% of the rows being not-unique.

PR #1062, opened 2021-07-19
Add Config.with_options and fix potential issues in test suite

davesque:
Adds a new method to the `Config` object that enables setting config values temporarily inside of a with block.  Adding this functionality was necessary to fix a potential issue in the test suite.  The problem was that any test case that modifies the global config singleton with `ww.config.set_option` and then throws an `AssertionError` could cause the global config to be left in that state during the subsequent execution of further test cases.

### DIFF
--- a/docs/source/guides/setting_config_options.ipynb
+++ b/docs/source/guides/setting_config_options.ipynb
@@ -41,7 +41,9 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {},
+   "metadata": {
+    "scrolled": true
+   },
    "outputs": [],
    "source": [
     "ww.config.set_option('categorical_threshold', 25)\n",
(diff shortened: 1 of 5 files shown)


davesque:
Just saw that I forgot the release notes.  Will update.

tamargrey:
@davesque is this from a specific bug that was popping up or more a potential issue that we should take care to avoid running into?

davesque:
@tamargrey I didn't confirm that this was causing any actual failures.  It's more that I wanted to eliminate it as a possible cause for some failures I was diagnosing.  I realized that those tests would leave the config in a weird state if they failed and that could definitely cause other tests to fail if they depended on behavior governed by those settings.  And it also seemed like a nice to have feature.

davesque:
@tamargrey It's because some of the functionality I was updating to address #1051 caused failures in those tests.

tamargrey:
@davesque cool--makes sense! Thanks for adding

thehomebrewnerd (approved):
Looks good!

PR #1066, opened 2021-07-22
Add serialization guide to docs

### LINKED ISSUE #866: Add serialization/deserialization guide to documentation
- We should add a page to our docs, with a guide on how to use our serialization/deserialization.
- This Guide should cover `to_disk` function (which saves to a folder with schema JSON + data).
- It should also show how to read using `read_file` for reading in a single file into a Woodwork DataFrame

jeff-hernandez:
Closes #866


docs/source/guides/saving_and_loading_dataframes.ipynb
```diff
@@ -0,0 +1,260 @@
...
+   "id": "federal-queensland",
+   "metadata": {},
+   "source": [
+    "# Saving and Loading DataFrames\n",
+    "\n",
+    "In this guide, you will learn how to save and load Woodwork DataFrames.\n",
+    "\n",
+    "## Saving a Woodwork DataFrame\n",
+    "\n",
+    "After defining a Woodwork DataFrame with the proper logical types and semantic tags, you can save the DataFrame and  typing information by using [`DataFrame.ww.to_disk`](https://woodwork.alteryx.com/en/stable/generated/woodwork.table_accessor.WoodworkTableAccessor.to_disk.html#woodwork.table_accessor.WoodworkTableAccessor.to_disk). This method will create a directory that contains a `data` folder and a `woodwork_typing_info.json` file. To illustrate, I will use this retail DataFrame which already comes configured with Woodwork typing information."
```
gsheni, line 14: The format

ting for this doesn't look right.

docs/source/guides/saving_and_loading_dataframes.ipynb
```diff
@@ -0,0 +1,260 @@
...
+   "source": [
+    "df.head()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "designed-soccer",
+   "metadata": {},
+   "source": [
+    "From the `ww` acessor, use [`to_disk`](https://woodwork.alteryx.com/en/stable/generated/woodwork.table_accessor.WoodworkTableAccessor.to_disk.html#woodwork.table_accessor.WoodworkTableAccessor.to_disk) to save the Woodwork DataFrame."
```
gsheni, line 46: same comment as above

docs/source/guides/saving_and_loading_dataframes.ipynb
```diff
@@ -215,25 +215,37 @@
...
+    "    filepath='retail/data.feather',\n",
     "    content_type='feather',\n",
     "    **typing_information,\n",
     ")\n",
     "\n",
     "woodwork_df.ww"
    ]
+  },
+  {
+   "cell_type": "code",
```
gsheni, line 239: Let's hide this cell
- https://github.com/alteryx/featuretools/issues/1560#issuecomment-887799238

docs/source/guides/saving_and_loading_dataframes.ipynb
jeff-hernandez: Interesting, it seems to work locally in Jupyter notebooks. I removed the links.

docs/source/guides/saving_and_loading_dataframes.ipynb
jeff-hernandez: ✅ 8b68909

docs/source/guides/saving_and_loading_dataframes.ipynb
```diff
@@ -0,0 +1,275 @@
...
+    "    },\n",
+    "    'semantic_tags': {\n",
+    "        'order_id': {'category'},\n",
+    "        'product_id': {'category'},\n",
+    "        'quantity': {'numeric'},\n",
+    "        'unit_price': {'numeric'},\n",
+    "        'customer_name': {'category'},\n",
+    "        'country': {'category'},\n",
+    "        'total': {'numeric'},\n",
+    "    }\n",
```
rwedge, line 173: As part of this example, should we actualy load the JSON in from file separately instead of / in addition to defining it?

docs/source/guides/saving_and_loading_dataframes.ipynb
```diff
@@ -0,0 +1,275 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "dependent-liberty",
+   "metadata": {},
+   "source": [
+    "### Loading the DataFrame and typing information separately\n",
+    "\n",
+    "You can also load the Woodwork DataFrame and typing information separately by using `woodwork.read_file`. This function uses the `content_type` parameter to determine the file format. If `content_type` is not specified, the function will try to infer the file format from the file extension. The typing information such as the index, time index, logical types, and semantics tags are optional parameters. This approach is helpful if you want to save and load the typing information outside the specified directory or read a data file directly into a Woodwork DataFrame. To illustrate, I will write a pandas DataFrame to a single file in different formats, then load it into a Woodwork DataFrame using this typing information."
```
rwedge, line 139: Can we interweave the example with this paragraph?  We can demonstrate the basics of how it works before going over all the specifics

docs/source/guides/saving_and_loading_dataframes.ipynb
```diff
@@ -0,0 +1,275 @@
...
+   "id": "federal-queensland",
+   "metadata": {},
+   "source": [
+    "# Saving and Loading DataFrames\n",
+    "\n",
+    "In this guide, you will learn how to save and load Woodwork DataFrames.\n",
+    "\n",
+    "## Saving a Woodwork DataFrame\n",
+    "\n",
+    "After defining a Woodwork DataFrame with the proper logical types and semantic tags, you can save the DataFrame and  typing information by using `DataFrame.ww.to_disk`. This method will create a directory that contains a `data` folder and a `woodwork_typing_info.json` file. To illustrate, I will use this retail DataFrame which already comes configured with Woodwork typing information."
```
rwedge, line 14: In other docs we use "we" instead of "I"; there's also another "I' earlier in the notebook

docs/source/guides/saving_and_loading_dataframes.ipynb
jeff-hernandez: yes, updated

docs/source/guides/saving_and_loading_dataframes.ipynb
jeff-hernandez: updated

docs/source/guides/saving_and_loading_dataframes.ipynb
jeff-hernandez: The example was updated to load in the JSON, but it seems like the JSON schema doesn't directly map to the typing information parameters in `read_file`. How do we want to convert `column_typing_info` to logical types and semantic tags?

docs/source/guides/saving_and_loading_dataframes.ipynb
```diff
@@ -0,0 +1,264 @@
...
+   "execution_count": null,
+   "id": "induced-newton",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from woodwork import read_file\n",
+    "\n",
+    "woodwork_df = read_file(\n",
+    "    filepath='retail/data.csv',\n",
+    "    content_type='csv',\n",
```
gsheni, line 197:
```suggestion
    "    content_type='text/csv',\n",
```

docs/source/guides/saving_and_loading_dataframes.ipynb
```diff
@@ -0,0 +1,264 @@
...
+    "    index=typing_information['index'],\n",
+    "    time_index=typing_information['time_index'],\n",
+    ")\n",
+    "\n",
+    "woodwork_df = read_file(\n",
+    "    filepath='retail/data.parquet',\n",
+    "    content_type='parquet',\n",
```
gsheni, line 204:
```suggestion
    "    content_type='application/parquet',\n",
```

docs/source/guides/saving_and_loading_dataframes.ipynb
```diff
@@ -0,0 +1,264 @@
...
+    "    index=typing_information['index'],\n",
+    "    time_index=typing_information['time_index'],\n",
+    ")\n",
+    "\n",
+    "woodwork_df = read_file(\n",
+    "    filepath='retail/data.feather',\n",
+    "    content_type='feather',\n",
```
gsheni, line 211:
```suggestion
    "    content_type='application/feather',\n",
```

docs/source/guides/saving_and_loading_dataframes.ipynb
rwedge: Yeah that's kind of messy.  @thehomebrewnerd @tamargrey do you envision `read_file` being used without the JSON?  What scenarios do we want to document?

gsheni: **read_file** can be used with/without the JSON. It can also be used if your typing information is stored in a database and you want to pass it in with `read_file`

rwedge: @gsheni that was poorly worded, I mean if `read_file` was meant to be used both with and without the JSON, since you'd need to do some preprocessing on the JSON info first before feeding it to `read_file`

gsheni: **read_file** isn't meant to be used with the JSON generated from `to_disk`. It is meant to be used where the user wants to store typing information in their own way.

rwedge: Then my intial suggestion was a poor one and we shouldn't use the JSON here

rwedge: But perhaps we should put the sentence explaining why one might want to do this as the second sentence in the paragraph, to communicate why you might want to do this before explaining the specifics of how

thehomebrewnerd: Not sure if this helps, but I tend to think of `read_file` as an analog to `pandas.read_csv`. It's just an easy helper function to take data stored on disk in some format and read it into a Woodwork initialized dataframe. You don't need to specify any logical types or semantic tags if you don't want to, but if you want more control over how things get initialized you can pass those values as well. It just allows you to combine the read and init functions in a single call instead of having to call the library read function and then initialize Woodwork.

The `to_disk` method coupled with the `read_woodwork_table` function are what you would use to serialize and deserialize a Woodwork dataframe if you want the deserialized table to match exactly the original dataframe, including all the typing info.

rwedge: Which I see you already did, I was looking at an old build

jeff-hernandez: I reverted the change and updated the comments to help clarify when we would want to use this approach.

rwedge: looks good 👍 

what do you think of adding an example of the simple use Nate talked about -- just calling `read_file`, no typing info -- at the beginning to demonstrate a quick way to get a woodwork DF?

jeff-hernandez: Sure I'll add that

rwedge (approved):
Looks good

PR #1083, opened 2021-07-28
Make sampling for inference in woodwork more consistent

davesque:
- Removes any sampling that occurs in inference functions defined in `woodwork.type_sys.inference_functions`.
- Updates `TypeSystem.infer_logical_type` to standardize inference sampling.  Inference sampling is now done via a `.head(100000)` call that uses the appropriate API depending on the collection type (pandas, dask, or koalas).
- Changes the special case for all-null series to apply *before* sampling for pandas series and *after* sampling for dask and koalas series.  The thinking behind this is that we would ideally use the entire series to determine if there are any valid records and it is reasonably performant to do that when the series is already in-memory, but not otherwise.

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -11,6 +11,7 @@ Future Release
         * The criteria for categorical type inference have changed (:pr:`1065`)
         * The meaning of both the ``categorical_threshold`` and
           ``numeric_categorical_threshold`` settings have changed (:pr:`1065`)
+        * Make sampling for type inference more consistent (:pr:`1083`)
     * Documentation Changes
         * Fix some release notes that ended up under the wrong release (:pr:`1082`)
         * Add BooleanNullable and IntegerNullable types to the docs (:pr:`1085`)
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -43,14 +43,14 @@ Breaking Changes
       ``categorical_threshold`` or ``numeric_categorical_threshold`` settings
       will need to adjust their settings accordingly.
     * :pr:`1083`: The process of sampling series for logical type inference was
-      updated to be more consistent.  Before, only dask and koalas collections
-      were sampled in every case during type inference.  Also, further
-      randomized subsampling was performed in some cases during categorical
-      inference and in every case during email inference.  Overall, the way
-      sampling was done was inconsistent and unpredictable.  Now, the first
-      100,000 records of a series are sampled for logical type inference
-      regardless of collection type (pandas, dask, or koalas) although only
-      records from the first partition of a dask dataset will be used.
+      updated to be more consistent.  Before, initial sampling for inference
+      differed depending on collection type (pandas, dask, or koalas).  Also,
+      further randomized subsampling was performed in some cases during
+      categorical inference and in every case during email inference regardless
+      of collection type.  Overall, the way sampling was done was inconsistent
+      and unpredictable.  Now, the first 100,000 records of a column are
+      sampled for logical type inference regardless of collection type although
+      only records from the first partition of a dask dataset will be used.
       Subsampling performed by the inference functions of individual types has
       been removed.  The effect of these changes is that inferred types may now
       be different although in many cases they will be more correct.
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -42,6 +42,18 @@ Breaking Changes
       categorical.  Users who have overridden either the
       ``categorical_threshold`` or ``numeric_categorical_threshold`` settings
       will need to adjust their settings accordingly.
+    * :pr:`1083`: The process of sampling series for logical type inference was
+      updated to be more consistent.  Before, only dask and koalas collections
+      were sampled in every case during type inference.  Also, further
+      randomized subsampling was performed in some cases during categorical
+      inference and in every case during email inference.  Overall, the way
+      sampling was done was inconsistent and unpredictable.  Now, the first
+      100,000 records of a series are sampled for logical type inference
+      regardless of collection type (pandas, dask, or koalas) although only
+      records from the first partition of a dask dataset will be used.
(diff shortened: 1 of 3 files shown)


woodwork/type_sys/type_system.py
```diff
@@ -252,16 +254,25 @@ def infer_logical_type(self, series):
...
+            # Special case for series with no valid values
+            if series.count() == 0:
+                return Unknown()
 
-        # Bring Dask or Koalas data into memory for inference
-        if _is_dask_series(series):
-            series = series.get_partition(0).compute()
-        if _is_koalas_series(series):
-            series = series.head(100000).to_pandas()
+            series = series.head(100000)
```
thehomebrewnerd, line 262: Should we store 100000 in a variable so if we decide to change it in the future, we only have to remember to do it in one place?

woodwork/type_sys/inference_functions.py
```diff
@@ -4,27 +4,15 @@
...
-    f"""
-    Return a sample of ``series`` for use during type inference.  If the length
-    of ``series`` is less than ``{INFERENCE_SAMPLE_SIZE}``, use the series
-    length as the sample size.
-    """
-    return series.sample(n=min(INFERENCE_SAMPLE_SIZE, len(series)))
-
-
-def categorical_func(series):
+def categorical_func(series: pd.Series) -> bool:
```
thehomebrewnerd, line 8: Maybe this should be a bigger discussion, but we haven't been doing type annotations so far, so not sure we should include this here?

woodwork/type_sys/type_system.py
```diff
@@ -252,16 +254,25 @@ def infer_logical_type(self, series):
...
+        else:
+            if _is_dask_series(series):
+                series = series.head(100000)
+            elif _is_koalas_series(series):
+                series = series.head(100000).to_pandas()
+            else:
+                raise ValueError(f"Unexpected arg type `{type(series)}`")  # pragma: no cover
```
thehomebrewnerd, line 269: I'm not sure we need to include this else clause, since we don't claim to support any other dataframe/series types beyond pandas, Dask or Koalas.

thehomebrewnerd (commented):
Overall I think this looks pretty good. Just a few relatively minor comments that we can discuss further as necessary.

woodwork/type_sys/type_system.py
davesque: Sure, y not

woodwork/type_sys/inference_functions.py
davesque: I've just been adding them as I go along.  I'm a pretty big advocate of type annotations.  Unless we're running mypy on the codebase, adding annotations basically has no effect.  But it can still serve as useful information for developers in understanding how a function behaves.  Are you concerned that the annotations will become stale somehow if we're not using them for type checking?

I should mention that I did bring the topic up with a few other people at Inno Labs and it seemed like the general sentiment was positive for this type of change.

woodwork/type_sys/type_system.py
davesque: By the way, I know I did the same thing with the constant in `inference_functions`.  In this case, it seemed like the values were all within a few lines of each other so it didn't matter as much.  But I think I was also sort of just being lazy :).

woodwork/type_sys/type_system.py
rwedge: For user interpretability, would  "Unexpected series type" or "Unsupported series type"

woodwork/type_sys/inference_functions.py
thehomebrewnerd: Don't get me wrong, I'm not anti-annotations by any means, but it's not a practice we have adopted broadly so far. Was just thinking that if this a direction we collectively want to head, we should have that discussion so the main team is aware of how we want to do things going forward, and that seems to be happening now...

davesque: Yep, valid point.

woodwork/type_sys/type_system.py
davesque: Sure, I can make that change.

woodwork/type_sys/inference_functions.py
davesque: I just took it out for now.

woodwork/type_sys/type_system.py
thehomebrewnerd: @davesque Makes sense, and I can see the benefit of leaving this if we do add support for new types, this could help alert us to a new condition we need to handle.

thehomebrewnerd (approved):
Looks good to me!

rwedge (approved):

docs/source/release_notes.rst
```diff
@@ -11,6 +11,7 @@ Future Release
         * The criteria for categorical type inference have changed (:pr:`1065`)
         * The meaning of both the ``categorical_threshold`` and
           ``numeric_categorical_threshold`` settings have changed (:pr:`1065`)
+        * Make sampling for type inference more consistent (:pr:`1083`)
```
tamargrey, line 14: would the change in sampling size be considered a breaking change because the logical types inferred at init could be different now than they were before for the same dataframe (both because we're now using the head and also taking more entries)?

davesque: Yeah, I suppose it could be.  I'm not sure actually.  You'll potentially get a different result but you'll also probably get a more correct one.  I suppose it's worth a note at least.

davesque: Recommended update made here: https://github.com/alteryx/woodwork/pull/1083/commits/3b6661b5f4c33b09e05022cddd2c7010d0b700a4

davesque: Actually I made some edits to improve the wording.  Probably should just look at the overall changes tab.

tamargrey: breaking changes section looks good to go!

PR #1085, opened 2021-07-29
add BooleanNullable and IntegerNullable to API docs

### LINKED ISSUE #1060: Perform schema init check inside method decorator
- As a developer, I wish I could add the schema init check that calls `_raise_init_error` if there's no Woodwork schema on the Accessor via a method decorator instead of adding the following lines to almost every method:

```
        if self._schema is None:
            _raise_init_error()
```

Initial discussion [here](https://github.com/alteryx/woodwork/pull/1048#discussion_r669976532). Could be included as a part of #724 if the solution there is to include schema validation as a part of that raise init error step.

tuethan1999:
- add BooleanNullable and IntegerNullable to API docs
- closes #1060

### DIFF
--- a/docs/source/api_reference.rst
+++ b/docs/source/api_reference.rst
@@ -138,6 +138,7 @@ Logical Types
     Age
     AgeNullable
     Boolean
+    BooleanNullable
     Categorical
     CountryCode
     Datetime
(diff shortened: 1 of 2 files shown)


gsheni (approved):
looks good. Just need to update the release notes

Issue #724, opened 2021-03-18
Need better error if user assigns new column to dataframe outside of Woodwork


thehomebrewnerd:
If a user assigns a new column to a DataFrame, but does so outside of the Woodwork accessor, and subsequently attempts to access the Woodwork information (which should now be invalid), a somewhat confusing `KeyError` is raised. Catching this and raising a better error would be helpful.

There might be other similar operations on the DataFrame that could also cause this type of error.

#### Code Sample, a copy-pastable example to reproduce your bug.

```python
import pandas as pd
import woodwork as ww

df = pd.DataFrame({
    'id': [0, 1, 2],
    'val': [10, 20, 30]
})
df.ww.init()
new = pd.Series([True, False, True], dtype='boolean')
df['new'] = new
df.ww
```

```
KeyError: 'new'

table_accessor.py:586: TypingInfoMismatchWarning: Operation performed by _repr_latex_ has invalidated the Woodwork typing information:
 The following columns in the DataFrame were missing from the typing information: {'new'}.
 Please initialize Woodwork with DataFrame.ww.init
  warnings.warn(TypingInfoMismatchWarning().get_warning_message(attr, invalid_schema_message, 'DataFrame'),
```

PR #1093, opened 2021-08-02
Check for schema initialization in a decorator

### LINKED ISSUE #1060: Perform schema init check inside method decorator

rwedge:
- Many methods and attributes in ColumnAccessor and TableAccessor have checks to see if woodwork has been initialized on the dataframe / series.  This PR moves these checks into a decorator.
- Closes #1060

### DIFF
--- a/woodwork/accessor_utils.py
+++ b/woodwork/accessor_utils.py
@@ -152,17 +152,23 @@ def _is_koalas_series(data):
     return False
 
 
-def _check_schema_init(data_structure):
+def _check_schema_init(method):
     '''Decorator for Woodwork ColumnAccessor and WoodworkTableAccessor that
        checks schema initialization
     '''
-    def inner_decorator(method):
-        @wraps(method)
-        def wrapper(self, *args, **kwargs):
-            if self._schema is None:
-                msg = (f"Woodwork not initialized for this {data_structure}. "
-                       f"Initialize by calling {data_structure}.ww.init")
-                raise WoodworkNotInitError(msg)
-            return method(self, *args, **kwargs)
-        return wrapper
-    return inner_decorator
+    @wraps(method)
+    def wrapper(self, *args, **kwargs):
+        if self._schema is None:
+            from woodwork.column_accessor import WoodworkColumnAccessor
+            from woodwork.table_accessor import WoodworkTableAccessor
+            if isinstance(self, WoodworkTableAccessor):
+                data_structure = "DataFrame"
+            elif isinstance(self, WoodworkColumnAccessor):
+                data_structure = "Series"
+            else:
+                raise TypeError("decorator was applied to a non-accessor method")
+            msg = (f"Woodwork not initialized for this {data_structure}. "
+                   f"Initialize by calling {data_structure}.ww.init")
+            raise WoodworkNotInitError(msg)
+        return method(self, *args, **kwargs)
+    return wrapper
--- a/woodwork/accessor_utils.py
+++ b/woodwork/accessor_utils.py
@@ -152,23 +152,25 @@ def _is_koalas_series(data):
     return False
 
 
-def _check_schema_init(method):
-    '''Decorator for Woodwork ColumnAccessor and WoodworkTableAccessor that
-       checks schema initialization
-    '''
+def _check_column_schema(method):
+    '''Decorator for WoodworkColumnAccessor that checks schema initialization'''
     @wraps(method)
     def wrapper(self, *args, **kwargs):
         if self._schema is None:
-            from woodwork.column_accessor import WoodworkColumnAccessor
-            from woodwork.table_accessor import WoodworkTableAccessor
-            if isinstance(self, WoodworkTableAccessor):
-                data_structure = "DataFrame"
-            elif isinstance(self, WoodworkColumnAccessor):
-                data_structure = "Series"
-            else:
-                raise TypeError("decorator was applied to a non-accessor method")
-            msg = (f"Woodwork not initialized for this {data_structure}. "
-                   f"Initialize by calling {data_structure}.ww.init")
+            msg = ("Woodwork not initialized for this Series. Initialize by "
+                   "calling Series.ww.init")
+            raise WoodworkNotInitError(msg)
+        return method(self, *args, **kwargs)
+    return wrapper
+
+
+def _check_table_schema(method):
+    '''Decorator for WoodworkTableAccessor that checks schema initialization'''
+    @wraps(method)
+    def wrapper(self, *args, **kwargs):
+        if self._schema is None:
+            msg = ("Woodwork not initialized for this DataFrame. Initialize by "
+                   "calling DataFrame.ww.init")
             raise WoodworkNotInitError(msg)
         return method(self, *args, **kwargs)
     return wrapper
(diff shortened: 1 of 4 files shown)


davesque (approved):
Hey, this is awesome!

woodwork/column_accessor.py
```diff
@@ -30,6 +30,17 @@
 ks = import_or_none('databricks.koalas')
 
 
+def _check_series_schema(func):
```
tamargrey, line 33: thoughts on adding this and the dataframe wrapper function to `accessor_utils.py`?

woodwork/column_accessor.py
```diff
@@ -30,6 +30,17 @@
...
+    '''Decorator for WoodworkColumnAccessor that checks schema initialization'''
+    def wrapper(self, *args, **kwargs):
+        if self._schema is None:
```
tamargrey, line 36: Would it be worth testing the behavior where someone used this decorator on a method that's not on the Series Accessor? Definitely an edge case, and it's only used internally, so probably not necessary, but I just wanted to bring it up.

woodwork/table_accessor.py
```diff
@@ -39,6 +39,17 @@
 ks = import_or_none('databricks.koalas')
 
 
+def _check_df_schema(func):
```
thehomebrewnerd, line 42: Could/should we combine the two decorator functions into one? Or is there a reason outside of the error message to keep them as separate functions?

Not a big deal either way, just thought I'd mention as a possible option.

rwedge: The slightly different error message is the only reason they're separate.  I don't really want to make the message configurable since that would add a possible kwarg collision

rwedge: I guess it could check `self` to get that info

woodwork/column_accessor.py
davesque: Yeah I suppose people could do that in theory.  The decorator will work on any object that has a `_schema` property.  Were you thinking that we should add an `isinstance` check or something to prevent people from using it on methods from the wrong object?  I would personally say it's probably not worth doing that.  The underscore prefix I think indicate pretty clearly that it's not intended for public use.

woodwork/column_accessor.py
```diff
@@ -108,32 +108,29 @@ def schema(self):
         return copy.deepcopy(self._schema)
 
     @property
+    @_check_schema_init("Series")
```
rwedge, line 111: storing either "Series" or the whole message in a constant seems the like a good idea

woodwork/table_accessor.py
davesque: @rwedge Yeah, you could do something like `class_name = repr(type(self))`.  And then use `{class_name}` in the exception message.

woodwork/accessor_utils.py
```diff
@@ -147,3 +150,19 @@ def _is_koalas_series(data):
...
+    def inner_decorator(method):
+        @wraps(method)
+        def wrapper(self, *args, **kwargs):
+            if self._schema is None:
+                msg = (f"Woodwork not initialized for this {data_structure}. "
+                       f"Initialize by calling {data_structure}.ww.init")
+                raise WoodworkNotInitError(msg)
+            return method(self, *args, **kwargs)
+        return wrapper
+    return inner_decorator
```
davesque, line 168: Seems like this would also work here.  Then you don't need to include the explicit class name in each decorator call.
```suggestion
def _check_schema_init(method):
    '''Decorator for WoodworkColumnAccessor and WoodworkTableAccessor that
       checks schema initialization
    '''
    @wraps(method)
    def wrapper(self, *args, **kwargs):
        class_name = type(self).__name__
        if self._schema is None:
            msg = (f"Woodwork not initialized for this {class_name}. "
                   f"Initialize by calling {class_name}.ww.init")
            raise WoodworkNotInitError(msg)
        return method(self, *args, **kwargs)
    return wrapper
```

davesque (commented):
I saw some of the other discussion and thought I'd add my own suggestion here relating to it.

woodwork/accessor_utils.py
davesque: Actually, it looks like you couldn't use `type(self).__name__` since `self` is a column or table accessor.  But you could still determine the correct pandas collection type by doing something like `isinstance(self, PandasColumnAccessor)`, etc.  Seems better than putting a class name explicitly into all the decorator calls that could potentially be typoed.

davesque: And then, for that matter, maybe it would make more sense to do what @tamargrey suggested and include some kind of check for the type of `self`:
```suggestion
def _check_schema_init(method):
    '''Decorator for WoodworkColumnAccessor and WoodworkTableAccessor that
       checks schema initialization
    '''
    @wraps(method)
    def wrapper(self, *args, **kwargs):
        if isinstance(self, WoodworkTableAccessor):
            data_structure = "DataFrame"
        elif isinstance(self, WoodworkColumnAccessor):
            data_structure = "Series"
        else:
            raise TypeError("decorator was applied to a non-accessor method")

        if self._schema is None:
            msg = (f"Woodwork not initialized for this {data_structure}. "
                   f"Initialize by calling {data_structure}.ww.init")
            raise WoodworkNotInitError(msg)

        return method(self, *args, **kwargs)

    return wrapper
```

rwedge: I think we'll run into a circular import issue since this function needs to be defined before the accessor classes in order to be used.  We could import the accessors within the method to avoid the circular import.

Another approach would be to check if Table or Column are in the class name, assuming we're willing to limit our subclasses to always include TableAccessor or ColumnAccessor in the name

tamargrey: I'd be okay with importing the accessors within the method if need be. But I would also be okay with keeping these as two separate functions.

My thought process behind that is that in having separate functions we 
1. Keep the decorator function simpler and easier to understand
2. Can more easily add additional DataFrame or Series-specific checks if we decide to expand the checks performed in the decorator (for example we were considering adding a schema validity check as a part of #724, and we'd need separate validity checks for dataframes vs series)

davesque: I think the local import to avoid circular imports is fine.  They should probably go just inside of the decorator body just before the wrapper function is defined.  Then the overhead is restricted to when any module is parsed and the decorators are executed to yield the wrapper function (one-time cost).

However, I'm also fine with keeping them separate.  Really don't have a specific preference here.

The one thing I would say is that, if we keep the multi-use decorator, let's not look for a substring in the class name.  I often really try to avoid "parsing" class or function names as part of a library's API.  That makes it really likely that, at some point, someone will change the name of some resource without realizing that there's some parsing facility somewhere in the library that expects the names of things to have a certain format.  And that's not something you can grep for.  On the other hand, bare identifiers that refer to class or function objects (or string literals that refer to the same) are greppable and can be updated fairly easily if someone ever renames anything.

davesque (approved):
Awesome, looks good!

tamargrey (approved):
LGTM!

PR #1094, opened 2021-08-03
Add 'ignore' and 'passthrough' tags

### LINKED ISSUE #929: Add ignore/passthrough semantic tag to documentation page on Woodwork Logical Types and Semantic Tags
- We should show how the semantic tag `ignore` OR `passthrough` can be used to have pass through features.
- These are features a user does not want to use engineering features or modeling, but wants them when viewing their data.

tamargrey:
- Adds `'ignore'` and `'passthrough'` tags to `list_semantic_tags` and update the Understanding Logical Types and Semantic Tags guide to include explanations of them.
- Closes #929

### DIFF
--- a/docs/source/guides/logical_types_and_semantic_tags.ipynb
+++ b/docs/source/guides/logical_types_and_semantic_tags.ipynb
@@ -81,6 +81,7 @@
     "This section lists any tags that Woodwork will neither add to a DataFrame nor take direct action upon, but that are recognized to have a special meaning.\n",
     "\n",
     "- `'date_of_birth'` - Indicates that a datetime column should be parsed as a date of birth\n",
+    "- `'ignore'`/`'passthrough'` - Indicates that a column should be ignored during feature engineering or model building but should still be passed through these operations so that the column is not lost.\n",
     "\n",
     "Additional tags beyond the ones Woodwork adds at initialization may be useful for a DataFrame's interpretability, so users are encouraged to add any tags that will allow them to use their data more efficiently. \n",
     "\n",
--- a/docs/source/guides/logical_types_and_semantic_tags.ipynb
+++ b/docs/source/guides/logical_types_and_semantic_tags.ipynb
@@ -78,7 +78,7 @@
     "    - A time index column will contain either datetime or numeric data\n",
     "\n",
     "#### Other Tags\n",
-    "This section lists any tags that Woodwork will neither add to a DataFrame nor take direct action upon, but that are recognized to have a special meaning.\n",
+    "The tags listed below may be added directly to columns during or after Woodwork initialization. They are tags that have suggested meanings and that can be added to columns that will be used in the manner described below. Woodwork will neither add them automatically to a DataFrame nor take direct action upon a column if they are present. \n",
     "\n",
     "- `'date_of_birth'` - Indicates that a datetime column should be parsed as a date of birth\n",
     "- `'ignore'`/`'passthrough'` - Indicates that a column should be ignored during feature engineering or model building but should still be passed through these operations so that the column is not lost.\n",
--- a/woodwork/type_sys/utils.py
+++ b/woodwork/type_sys/utils.py
@@ -111,7 +111,7 @@ def list_semantic_tags():
          for tag in sem_tags]
     )
     tags_df = tags_df.append(
-        pd.DataFrame([['index', False, [ww.type_system.str_to_logical_type(tag) for tag in ['integer', 'double', 'categorical', 'datetime']]],
+        pd.DataFrame([['index', False, 'Any LogicalType'],
                       ['time_index', False, [ww.type_system.str_to_logical_type('datetime')]],
                       ['date_of_birth', False, [ww.type_system.str_to_logical_type('datetime')]],
                       ['ignore', False, 'Any LogicalType'],
--- a/woodwork/type_sys/utils.py
+++ b/woodwork/type_sys/utils.py
@@ -113,7 +113,9 @@ def list_semantic_tags():
     tags_df = tags_df.append(
         pd.DataFrame([['index', False, [ww.type_system.str_to_logical_type(tag) for tag in ['integer', 'double', 'categorical', 'datetime']]],
                       ['time_index', False, [ww.type_system.str_to_logical_type('datetime')]],
-                      ['date_of_birth', False, [ww.type_system.str_to_logical_type('datetime')]]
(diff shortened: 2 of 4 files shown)


woodwork/type_sys/utils.py
```diff
@@ -113,7 +113,9 @@ def list_semantic_tags():
     tags_df = tags_df.append(
         pd.DataFrame([['index', False, [ww.type_system.str_to_logical_type(tag) for tag in ['integer', 'double', 'categorical', 'datetime']]],
                       ['time_index', False, [ww.type_system.str_to_logical_type('datetime')]],
-                      ['date_of_birth', False, [ww.type_system.str_to_logical_type('datetime')]]
+                      ['date_of_birth', False, [ww.type_system.str_to_logical_type('datetime')]],
+                      ['ignore', False, 'Any LogicalType'],
+                      ['passthrough', False, 'Any LogicalType']
```
tamargrey, line 118: Adding a string makes it clearer that all logical types are valid than listing all logical types would be, but means that we couldn't use the values to, say, select by logical type for columns where these tags would be permissible (not sure that that's even an expected use case...)

woodwork/type_sys/utils.py
```diff
@@ -113,7 +113,9 @@ def list_semantic_tags():
     tags_df = tags_df.append(
         pd.DataFrame([['index', False, [ww.type_system.str_to_logical_type(tag) for tag in ['integer', 'double', 'categorical', 'datetime']]],
```
tamargrey, line 114: We don't ever enforce the valid logical types for `index` that we list here. Can we change this to `Any LogicalType` as well?

thehomebrewnerd (commented):
I guess I'm a bit confused about the purpose of this PR. If Woodwork doesn't add these tags, or otherwise do anything with them, why add them? Are these different enough compared to any other user-defined semantic tags that they require special treatment/documentation?

tamargrey:
> I guess I'm a bit confused about the purpose of this PR. If Woodwork doesn't add these tags, or otherwise do anything with them, why add them? Are these different enough compared to any other user-defined semantic tags that they require special treatment/documentation?
@thehomebrewnerd 

My understanding (@gsheni please correct me if I'm wrong) is that these are tags that Woodwork will not itself add or act on but that are recognized as having a specific meaning. We've already had `date_of_birth`, which definitely has meaning in the `Age` primitive in Featuretools. And we'll eventually want to do something similar with passthrough features.

thehomebrewnerd:
> We've already had `date_of_birth`, which definitely has meaning in the `Age` primitive in Featuretools.

I guess my question applies more broadly to the `date_of_birth` tag as well. If we wanted to just use these for illustrative purposes to show users how the could add custom tags in their applications that have special meaning, that makes some sense. I'm still wrestling with the idea of listing certain tags that are considered to have special meaning, but aren't used in any way inside the Woodwork code base. If Woodwork doesn't do anything different with them compared to any other custom tag what makes them special?

gsheni:
@thehomebrewnerd I agree that Woodwork doesn't do anything with these semantic tags currently. 
We could update our statistical functions to recognize and take action on these tags.

In addition, we could document in Featuretools the meaning of these special tags. However, putting this in Woodwork docs allows other libraries to benefit from their usage (such as EvalML). 

Even if these tags aren't used in Woodwork, they are **suggested** tags that end users may want to use in certain situations.

woodwork/type_sys/utils.py
thehomebrewnerd: Seems fine to me

thehomebrewnerd:
> @thehomebrewnerd I agree that Woodwork doesn't do anything with these semantic tags currently.
> We could update our statistical functions to recognize and take action on these tags.
> 
> In addition, we could document in Featuretools the meaning of these special tags. However, putting this in Woodwork docs allows other libraries to benefit from their usage (such as EvalML).
> 
> Even if these tags aren't used in Woodwork, they are **suggested** tags that end users may want to use in certain situations.

So with this in mind, perhaps what I'm getting hung up on is the wording "recognized to have special meaning" . Could we update the wording in the section where these are discussed to reflect that we are just suggesting that the users use these tags when they have columns that need to be used in the manner described by the tag?

docs/source/guides/logical_types_and_semantic_tags.ipynb
```diff
@@ -78,9 +78,10 @@
     "    - A time index column will contain either datetime or numeric data\n",
     "\n",
     "#### Other Tags\n",
-    "This section lists any tags that Woodwork will neither add to a DataFrame nor take direct action upon, but that are recognized to have a special meaning.\n",
+    "The tags listed below may be added directly to columns during or after Woodwork initialization. They are tags that have suggested meanings and that can be added to columns that will be used in the manner described below. Woodwork will neither add them automatically to a DataFrame nor take direct action upon a column if they are present. \n",
```
tamargrey, line 81: @thehomebrewnerd is this along the lines of what you were thinking for this section?

thehomebrewnerd: Yeah, I like that a lot better given the discussion about these tags.

tamargrey: Awesome, thanks for the direction here!

tamargrey:
@thehomebrewnerd Another thing I noticed: `time_index` shows that `Datetime` is the only valid logical type, but Woodwork allows numeric logical types to be used for the time index, so the Age, Integer, and Double logical types should all be included there as well. 

As long as we're changing the index's valid logical types, I'm thinking it makes sense to also change `time_index`'s as well. 

(`date_of_birth` I'm less sold on telling people that they can use numeric logical types for...)

thehomebrewnerd:
> @thehomebrewnerd Another thing I noticed: `time_index` shows that `Datetime` is the only valid logical type, but Woodwork allows numeric logical types to be used for the time index, so the Age, Integer, and Double logical types should all be included there as well.
> 
> As long as we're changing the index's valid logical types, I'm thinking it makes sense to also change `time_index`'s as well.
> 
> (`date_of_birth` I'm less sold on telling people that they can use numeric logical types for...)

I think it makes sense to fix the time index line. I'm also in favor of not listing numeric types for `date_of_birth`, but probably even more in favor of removing it from our list of tags...

thehomebrewnerd:
> probably even more in favor of removing it from our list of tags...

I'm ok with leaving it in though

thehomebrewnerd (approved):
LGTM!

PR #1097, opened 2021-08-04
Pass through ordering information in transform method of Ordinal logical type [WIP, Do not merge]

### LINKED ISSUE #926: Pass Ordinal Ordering to pandas categorical dtype
- As a user, I wish I could use Woodwork to sort my Ordinal columns.
- pandas supports this already. You can tell pandas that a column is ordered, and pass the ordering.
  - https://pandas.pydata.org/pandas-docs/stable/user_guide/categorical.html#sorting-and-order
- This allows you to [sort](https://pandas.pydata.org/pandas-docs/stable/user_guide/categorical.html#reordering) on that column and make [comparisons](https://pandas.pydata.org/pandas-docs/stable/user_guide/categorical.html#comparisons).

davesque:
- Updates the `transform` method on the `Ordinal` logical type to use ordering information if it's available.
- Closes #926


thehomebrewnerd:
Generally this approach seems fine to me. I did a little testing and it seems that `df.cat.set_categories` isn't implemented for Koalas yet, so we might have to have some special handling for Koalas in the interim. We might also want to check their issues to see if they are planning to add support for that and create an issue if one doesn't already exist.

woodwork/logical_types.py
```diff
@@ -407,7 +408,12 @@ def _validate_data(self, series):
     def transform(self, series):
         """Validates the series and converts the dtype to match the logical type's if it is different."""
         self._validate_data(series)
-        return super().transform(series)
+
+        typed_ser = super().transform(series)
+        if pdtypes.is_categorical_dtype(typed_ser.dtype):
```
gsheni, line 413: Will this check work for Dask and Koalas? Or do we need get the type of Series beforehand?

gsheni: I see that there is a test below for this. Just might be worth considering since we don't do `pdtypes.is_X` anywhere else in **logical_types.py**

davesque: Yeah, so it used to do that.  However, by the time we call this logic, koalas series have already been cast to the backup type and are not categorical.  So this already kind of handles that case.  We can add the explicit logic back in if you want.

gsheni: Sounds good.

woodwork/tests/logical_types/test_logical_types.py
```diff
@@ -55,6 +55,40 @@ def test_ordinal_init_with_order():
...
+    pd.testing.assert_index_equal(ser_.cat.categories, pd.Int64Index(order))
+
+
+def test_ordinal_transform_koalas(ordinal_transform_series_koalas) -> None:
+    order = [2, 1, 3]
+    typ = Ordinal(order=order)
+
+    ser_ = typ.transform(ordinal_transform_series_koalas)
+
+    assert ser_.dtype == pd.StringDtype()
```
thehomebrewnerd, line 89: What do you think about parameterizing these into a single test like we have done with most of our other tests? I think the pandas and Dask paths would be the same, but we would have different assertions for Koalas, at least until they fully add support for the functionality we need.

woodwork/tests/logical_types/test_logical_types.py
```diff
@@ -55,6 +55,40 @@ def test_ordinal_init_with_order():
     assert ordinal_from_tuple.order == order
 
 
+def test_ordinal_transform_validates(ordinal_transform_series_pandas) -> None:
+    typ = Ordinal(order=None)
+    with pytest.raises(TypeError):
```
thehomebrewnerd, line 60: Should we also check that the error message matches what we expect, so we know the right `TypeError` is getting raised?

thehomebrewnerd (commented):
Looking pretty good, just a couple comments/questions.

woodwork/tests/logical_types/test_logical_types.py
davesque: Yeah, probably a good idea.  I'll change it.

woodwork/tests/logical_types/test_logical_types.py
davesque: Yeah, things here are a bit different looking than lots of our other tests.  I guess the reason I did this was that I've always generally tried to avoid putting too much complex logic inside of test cases.  If a test case has if statements or any kind of programmatic "topology" like that, that usually indicates to me that we're beginning to mirror the structure of the code we're testing inside of the test case.  To me, that makes things a bit harder to interpret and also introduces a kind of unnecessary redundancy into the library.  In a way, it's like we have the if statements that handle different cases inside of the code we're testing and then we write them again in the tests.  They kind of "lock" into each other and don't really do anything other than just show that we happened to write a similar collection of if statements twice successfully.  Then, if anyone ever updates that logic, we end up having to make two sets of similar changes in two places.  It's kind of like a test case tautology.  We're testing that code written in the same way acts the same way.  On the other hand, if the code in our test cases has a different structure than the code its testing, that seems to provide more assurance that the functionality works properly.  It's like we've proven that the same functionality represented in two different ways is equivalent vs. showing that the same functionality represented in the same way is equivalent.

It's kind of a subtle point, but all of the above is why I generally prefer writing more "flat" tests that don't include any branching logic.  They tend to be more readable and they avoid what can begin to feel like a literal duplication of code.  In this particular situation, each test case differed slightly (pandas is the default; dask requires the `.compute()` call; and koalas is the outlier since it basically ignores the order parameter).  So I thought it made sense to "unroll" the fixtures into separate functions and also have separate test cases.

Would also be happy to do a video chat about this.

davesque: After a slack chat, we decided to keep things the same for now.

thehomebrewnerd (approved):
LGTM!

PR #1089, opened 2021-07-30
Fix to allow all np.nan values in _infer_datetime_format with Datetime ltype

thehomebrewnerd:
- Fix to allow all np.nan values in _infer_datetime_format
- Closes #1088 

Updates logic in `_infer_datetime_format` to prevent error when trying to infer formats on a series with all `np.nan` values, particularly a Koalas series.

PR #1086, opened 2021-07-29
Add in-depth logical types and tags guide

tamargrey:
- Adds a guide to Woodwork that walks through each of the logical types and semantic tags in Woodwork and explains how they're used
- Closes #1061 

To Do:
- [x] Add use-case examples under each ltype and tag
- [x] Read through and clean up/clarify
- [x] Add correct links

PR #1082, opened 2021-07-28
Fix release notes from categorical inference branch

davesque:
- Release notes from #1065 accidentally ended up in the 0.5.1 section.  This moves them back into the "Future release" section.

PR #1098, opened 2021-08-04
v0.6.0

### LINKED ISSUE #1087: release Woodwork v0.6.0

davesque:
v0.6.0 Aug 4, 2021
==================
* Fixes
  * Fix bug in ``_infer_datetime_format`` with all ``np.nan`` input (#1089)
* Changes
  * The criteria for categorical type inference have changed (#1065)
  * The meaning of both the ``categorical_threshold`` and ``numeric_categorical_threshold`` settings have changed (#1065)
  * Make sampling for type inference more consistent (#1083)
  * Accessor logic checking if Woodwork has been initialized moved to decorator (#1093)
* Documentation Changes
  * Fix some release notes that ended up under the wrong release (#1082)
  * Add BooleanNullable and IntegerNullable types to the docs (#1085)
  * Add guide for saving and loading Woodwork DataFrames (#1066)
  * Add in-depth guide on logical types and semantic tags (#1086)
* Testing Changes
  * Add additional reviewers to minimum and latest dependency checkers (#1070, #1073, #1077)
  * Update the sample_df fixture to have more logical_type coverage (#1058)

@davesque, @gsheni, @jeff-hernandez, @rwedge, @tamargrey, @thehomebrewnerd, @tuethan1999

Breaking Changes
-----------------
* #1065: The criteria for categorical type inference have changed.  Relatedly, the meaning of both the ``categorical_threshold`` and ``numeric_categorical_threshold`` settings have changed.  Now, a categorical match is signaled when a series either has the "categorical" pandas dtype *or* if the ratio of unique value count (nan excluded) and total value count (nan also excluded) is below or equal to some fraction.  The value used for this fraction is set by the ``categorical_threshold`` setting which now has a default value of ``0.2``.  If a fraction is set for the ``numeric_categorical_threshold`` setting, then series with either a float or integer dtype may be inferred as categorical by applying the same logic described above with the ``numeric_categorical_threshold`` fraction.  Otherwise, the ``numeric_categorical_threshold`` setting defaults to ``None`` which indicates that series with a numerical type should not be inferred as categorical.  Users who have overridden either the ``categorical_threshold`` or ``numeric_categorical_threshold`` settings will need to adjust their settings accordingly.
* #1083: The process of sampling series for logical type inference was updated to be more consistent.  Before, initial sampling for inference differed depending on collection type (pandas, dask, or koalas).  Also, further randomized subsampling was performed in some cases during categorical inference and in every case during email inference regardless of collection type.  Overall, the way sampling was done was inconsistent and unpredictable.  Now, the first 100,000 records of a column are sampled for logical type inference regardless of collection type although only records from the first partition of a dask dataset will be used.  Subsampling performed by the inference functions of individual types has been removed.  The effect of these changes is that inferred types may now be different although in many cases they will be more correct.


tamargrey (approved):
lgtm!

PR #1100, opened 2021-08-09
Add initialize with partial Table or Column Schema

### LINKED ISSUE #965: Add a way of initializing Woodwork with partial Table or Column Schemas

tuethan1999:
- creation of init_with_partial_schema, init_with_full_schema, and init
- Closes #965


woodwork/table_accessor.py
```diff
@@ -38,24 +38,53 @@
...
-             schema=None,
-             validate=True,
-             use_standard_tags=True,
-             **kwargs):
-        """Initializes Woodwork typing information for a DataFrame.
+    def init(self, schema=None, **kwargs):
+        """Initializes Woodwork typing information for a DataFrame with only keyword arguments.
+        logical type inference is performed on columns with unspecified logical types"""
+        if schema is not None:
+            raise ValueError("call init_with_partial_schema or init_with_full_schema to pass a schema")
```
tamargrey, line 51: Didn't really think about this till now, but I wonder we might need to keep `init(schema=schema)` to avoid adding a breaking change. We could call `self.init_with_full_schema(schema=None)` so that it matches the current behavior by just using the schema, assuming it's relevant for the whole table, and ignoring any other parameters.

We'd probably want to add a warning that tells users that they should go through the full or partial schema methods, but I'm thinking that we should avoid the breaking change here if possible.

woodwork/table_accessor.py
```diff
@@ -38,24 +38,53 @@
...
+        if validate:
+            _check_schema(self._dataframe, schema)
+            _check_unique_column_names(self._dataframe)
+        self._schema = schema
+
+    def init_with_partial_schema(self,
+                                 schema: Optional[TableSchema]=None,
+                                 index: Optional[str]=None,
+                                 time_index: Optional[str]=None,
+                                 logical_types: Optional[dict[str, LogicalType]]=None,
```
tamargrey: Might be misreading/misunderstanding how the types work, but do we want the types for logical_types to be `Optional[dict[str, Union[str, LogicalType]]]`?

woodwork/table_accessor.py
```diff
@@ -38,24 +38,53 @@
...
+                                 already_sorted: Optional[bool]=False,
+                                 name: Optional[str]=None,
+                                 semantic_tags: Optional[dict[str, Union[str, list[str], set[str]]]]=None,
+                                 table_metadata: Optional[dict]=None,
+                                 column_metadata: Optional[dict[str, dict]]=None,
```
tamargrey: These params that can take in a dictionary where they keys are column names should allow the column names to be anything that can be a pandas column name (so it allows anything hashable like numbers). Something I think could be incredibly useful for us would be to create a `column_name` type alias for these params. So we could say `Optional[dict[column_name, dict]]` and `column_name` would be something like `Union[str, number, tuple]`

@davesque is there a way with python's types to make some sort of type alias? (or maybe the question should be: Is that an advisable thing to do even if it's possible)

woodwork/table_accessor.py
tamargrey: side note: So awesome to see the types here!

woodwork/table_accessor.py
```diff
@@ -84,70 +113,81 @@ def init(self,
...
-            # TableSchema uses a different default for use_standard_tags so we need to define it here
+        def left_join(left: dict, right: dict):
+            left = left or {}
+            right = right or {}
+            combined_dict = {}
+            for key in right:
+                combined_dict[key] = right[key]
+            for key in left:
+                combined_dict[key] = left[key]
+            return combined_dict
```
tamargrey, line 127: if I'm understanding the goal for this function correctly, you should be able to do `{**right, **left}`, and that will create a new dictionary object that will contain the `right` and `left` items, with any duplicate keys giving preference to `left`

woodwork/table_accessor.py
```diff
@@ -617,10 +657,10 @@ def _get_subset_df_with_schema(self, cols_to_include, inplace=False):
         if inplace:
             cols_to_drop = [col_name for col_name in self._dataframe.columns if col_name not in cols_to_include]
             self._dataframe.drop(cols_to_drop, axis='columns', inplace=True)
-            self.init(schema=new_schema, validate=False)
+            self.init_with_full_schema(schema=new_schema, validate=False)
             return
         new_df = self._dataframe[cols_to_include]
-        new_df.ww.init(schema=new_schema, validate=False)
+        new_df.ww.init_with_partial_schema(schema=new_schema, validate=False)
```
tamargrey, line 662: I believe `new_schema` will be the full schema for `new_df` since the subset used in line 656 is the same as what was used to create `new_df`, so `init_with_full_schema` may be more clear here

woodwork/table_accessor.py
```diff
@@ -84,70 +113,81 @@ def init(self,
...
+        existing_col_metadata = {}
+        existing_use_standard_tags = {}
+        existing_semantic_tags = {}
+        existing_col_origins = {}
+        
+        if schema:  # pull schema parameters
+            name = name or schema.name
+            index = index or schema.index
+            time_index = time_index or schema.time_index
+            table_metadata = table_metadata or schema.metadata
```
tamargrey, line 150: Should these objects that may be coming from another DataFrame's schema be deepcopied before being passed into a new DataFrame at ww init?

Basically, my fear is that using the same `table_metadata` object that's coming from a `schema` means that any changes to the resulting DataFrame's woodwork metadata would get propogated back to the DataFrame where this metadata originally came from

tamargrey (commented):
Very excited about getting this functionality in Woodwork! I think it'd be good to spend a little time making sure this api can be used for the various use cases that have come up (should all be referenced by the issue) in case we need to create a follow-up issues that gets us these use cases:
- woodwork concat_columns
- featuretools' add_last_time_indexes
- woodwork's `to_frame`
- evalML's use case for retaining as much typing info as possible

woodwork/table_accessor.py
tuethan1999: I added a `test_init_with_partial_schema_metadata_deep_copy` to show that changes don't propagate

woodwork/table_accessor.py
```diff
@@ -44,18 +45,49 @@ def __init__(self, dataframe):
...
-             use_standard_tags=True,
-             **kwargs):
-        """Initializes Woodwork typing information for a DataFrame.
+    def init(self, schema=None, **kwargs):
+        """Initializes Woodwork typing information for a DataFrame with only keyword arguments.
+        logical type inference is performed on columns with unspecified logical types"""
+        if schema is not None:
+            warnings.warn("Use init_with_full_schema when passing in a full schema and init_with_partial_schema when passing in a partial schema",
+                          UseInitWithSchemaWarning)
+            self.init_with_full_schema(schema, **kwargs)
```
rwedge, line 54: How difficult would it be to determine if a passed in schema was partial or not, and call the appropriate init method accordingly?

tuethan1999: not very, just threw in a line that checks if all the dataframe columns are in the schema

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -2538,3 +2520,78 @@ def test_ltype_conversions_nullable_types():
...
+def test_init_with_partial_schema_metadata_deep_copy(sample_df):
+    test_df = sample_df.copy()
+    sample_df.ww.init(table_metadata={'test': '123'})
+    schema = sample_df.ww[['id', 'full_name']].ww.schema
+    assert schema.metadata
+    assert test_df.ww.schema is None
+    test_df.ww.init_with_partial_schema(schema)
+    assert test_df.ww.schema.metadata
+    test_df.ww._schema.metadata = {'test2': '456'}
+    assert schema.metadata != test_df.ww.schema.metadata
```
tamargrey, line 2570: I think this test wouldn't catch the deepcopy behavior that we need to avoid.

Setting `test_df.ww._schema.metadata` with a completely new dictionary object (line 2569)  will only also change `schema`'s metadata if `schema` and `test_df.ww._schema` are the same `TableSchema` object.

This test needs to update the `test_df.ww._schema.metadata` dictionary that's already there. Adding a new entry will, I believe, test if a shallow copy _of the dictionary_ was performed (if both metadata dicts are the same object). But to confirm that a deep copy has occurred, we'll need to update a mutable value in the metadata dictionary (both lists, for example, would update if the lists are the same object). 

And example of this kind of test is `test_dataframe_methods_on_accessor_new_schema_object` or `test_iloc_table_does_not_propagate_changes_to_data`

freddyaboulton (commented):
Is there a shorthand for "Init with my old schema but redo inference on the logical types for some columns?"

In EvalML, we want to eventually get to a point where our components preserve the ww information as much as possible throughout the pipeline. One of the hurdles we face is that some of our components, e.g. Imputers, StandardScaler, modify the data in existing columns as opposed to just creating new features from old columns. The transformations done by these components usually "break" the typing info, e.g. an int gets converted to  a double.

EvalML currently has a janky util function for at least preserving the logical types in the columns that did not get transformed but we'd like to preserve more info (like semantic tags and indices) using `init_with_partial_schema`. 

I'm trying to see how that would look and I think the current implementation does not work 100% for our use case. I'm not requesting changes or anything on this PR, but I'd like your guys' thoughts on how we can achieve what we're after. Maybe there's already an easier way in ww to do it.

The use case is a component to scale numeric columns. This is a minimal repro:

```python
import woodwork as ww
from sklearn.preprocessing import StandardScaler
import pandas as pd

df = pd.DataFrame({"a": [1, 2, 3],
                   "b": [4, 5, 6],
                   "c": ["a", "b", "c"]})
df.ww.init(logical_types={"a": "Integer", "b": "Integer", "c": "Categorical"},
           semantic_tags={"a": ["column a"], "b": ["column b"]})

def scale_numerics(df):
    numeric = df.ww.select(["numeric"])
    scaled = pd.DataFrame(StandardScaler().fit_transform(numeric), columns=numeric.columns)
    return ww.concat_columns([df.ww.drop(list(numeric.columns)), scaled])
```

I can't pass the old schema to `init_with_partial_schema` because then the scaled columns get converted to ints and that's not what we want:

image

I can't "subset" the schema for the non-numeric columns because then some ww info would be lost, in this case the semantic tags for the numeric columns:

image


What I can do is specify the logical types for the non-numeric columns and manually pass in as much as the schema information as possible:

image

I'm wondering if there's a short-hand for the third "solution". Having us do it manually in all our components could lead to bugs and it's possible my "solution" does not carry over all of the relevant schema information, (like the origin field?)

tuethan1999:
> I'm wondering if there's a short-hand for the third "solution". Having us do it manually in all our components could lead to bugs and it's possible my "solution" does not carry over all of the relevant schema information, (like the origin field?)
@freddyaboulton we could add a `force_infer_logical_types` argument

woodwork/tests/accessor/test_table_accessor.py
tuethan1999: you're right!  I've changed it to copy the same things that `_get_subset_schema` copies

freddyaboulton:
@tuethan1999 I think that would work! Would the api be `force_infer_logical_types=["Integer", "Boolean"]` or would it infer all logical types again

tuethan1999:
@freddyaboulton I think the api would be more along the lines `force_infer_logical_types=['b', 'c']` where the iterables arguments are column names

freddyaboulton:
@tuethan1999 Sounds good!

woodwork/table_accessor.py
```diff
@@ -84,70 +120,72 @@ def init(self,
...
+        logical_types = _infer_missing_logical_types(self._dataframe, logical_types, existing_logical_types)
+        column_descriptions = {**existing_col_descriptions, **(column_descriptions or {})}
+        column_metadata = {**existing_col_metadata, **(column_metadata or {})}
+
+        # prioritize columns in schema if use_standard_tags is True
+        normalized_tags = normalize_standard_tags(use_standard_tags, column_names)
+        if use_standard_tags is True:
+            use_standard_tags = {**normalized_tags, **existing_use_standard_tags}
+        else:
+            use_standard_tags = {**existing_use_standard_tags, **normalized_tags}
```
tamargrey, line 169: Not totally sure I'm following what's happening here... But if I understand what the behavior here is:

`use_standard_tags` will be `True` by default, so if that's the case, we assume that that's just the default (and not that the user wanted all columns to use `True`), so we overwrite those values with whatever was in the Schema. This will contradict what the docstring says about the `schema` arg and the `use_standard_tags` arg if the Schema being passed in has some columns as `False`.

What if someone does want to set all of them to True? Not sure that it'd be expected to suddenly have the schema overwrite. I think we might need the default to be None here. That way a user can specify `True` and have it mean `True`.

Can we add some tests that cover this behavior?

woodwork/table_accessor.py
```diff
@@ -690,11 +728,11 @@ def rename(self, columns, inplace=False):
             if _is_koalas_dataframe(self._dataframe):
                 raise ValueError('Rename inplace not supported for Koalas')
             self._dataframe.rename(columns=columns, inplace=True)
-            self.init(schema=new_schema)
+            self.init_with_full_schema(schema=new_schema)
             return
         new_df = self._dataframe.rename(columns=columns)
 
-        new_df.ww.init(schema=new_schema)
+        new_df.ww.init_with_full_schema(schema=new_schema)
```
tamargrey, line 735: I know we weren't doing this before, but this seems like a good candidate for `validate=False`

woodwork/table_accessor.py
```diff
@@ -38,24 +38,60 @@
...
-             index=None,
-             time_index=None,
-             logical_types=None,
-             already_sorted=False,
-             schema=None,
-             validate=True,
-             use_standard_tags=True,
-             **kwargs):
-        """Initializes Woodwork typing information for a DataFrame.
+    def init(self, schema=None, **kwargs):
```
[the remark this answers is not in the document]
tamargrey: In general, we're adding a lot of new functionality with woodwork init here, and I think it could really benefit from an in-depth explanation somewhere. Maybe it makes sense to create an issue to make a guide on the different ways to initialize woodwork.

woodwork/table_accessor.py
```diff
@@ -84,70 +120,72 @@ def init(self,
...
-                parsed_logical_types[name] = logical_type
 
-                updated_series = logical_type.transform(series)
-                if updated_series is not series:
-                    self._dataframe[name] = updated_series
-
-            column_names = list(self._dataframe.columns)
-
-            # TableSchema uses a different default for use_standard_tags so we need to define it here
+        def normalize_standard_tags(use_standard_tags: Union[bool, Dict[str, bool]], column_names: List[str]):
```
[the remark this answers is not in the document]
tuethan1999: I think it's better practice to have it separated out. I separated out the function from the init method and added a docstring to make it clear what it's purpose was

woodwork/table_accessor.py
tuethan1999: I think I meant to set the default value for unspecified use_standard_tags columns. I removed this.

docs/source/release_notes.rst
```diff
@@ -7,14 +7,15 @@ Future Release
...
+        * Add initialize with partial table schema  (:pr:`1100`)
         * Apply ordering specified by the ``Ordinal`` logical type to underlying series (:pr:`1097`)
     * Fixes
     * Changes
     * Documentation Changes
     * Testing Changes
 
     Thanks to the following people for contributing to this release:
-    :user:`davesque`, :user:`tamargrey`
+    :user:`davesque`, :user:`tamargrey`, :user:`tuethan1999`
```
gsheni, line 19: If this is a breaking change, let's document it here.

woodwork/table_accessor.py
```diff
@@ -38,24 +38,60 @@
...
-             schema=None,
-             validate=True,
-             use_standard_tags=True,
-             **kwargs):
-        """Initializes Woodwork typing information for a DataFrame.
+    def init(self, schema=None, **kwargs):
+        """Initializes Woodwork typing information for a DataFrame. Logical type inference is performed on columns with unspecified logical types"""
+        if schema is not None and set(schema.columns) == set(self._dataframe.columns):
+            self.init_with_full_schema(schema, **kwargs)
+        else:
```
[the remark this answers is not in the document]
tuethan1999: I've mapped each case above to it's corresponding test
- `init(**params)`: `test_init_detect_partial_schema`
- `init(schema=partial_schema, **overriding_params)`: `test_init_with_partial_schema_override_schema`
- `init(schema=full_schema)`: `test_init_with_full_schema`
- `init(schema=full_schema, **ignored_params)`: `test_accessor_with_schema_parameter_warning`
- `init(schema=schema_with_columns_not_in_df)`:  `test_check_partial_schema`

If you'd like I can switch all the `init` tests to be either `'init_with_partial_schema` or `init_with_full_schema` tests although keeping them the way makes sure there are no breaking changes

woodwork/table_accessor.py
tuethan1999: I've made an alias for ColumnName

docs/source/release_notes.rst
```diff
@@ -7,14 +7,20 @@ Future Release
...
+    
+Breaking Changes
+++++++++++++++++
+    * Calling `init` with a partial schema now calls `init_with_partial_schema`
```
gsheni, line 22: @tuethan1999 Can you expand on this? We need to explain the behavior before the change, and the new behavior now. See [here](https://woodwork.alteryx.com/en/latest/release_notes.html#breaking-changes) for an example

docs/source/release_notes.rst
```diff
@@ -19,8 +19,12 @@ Future Release
     
 Breaking Changes
 ++++++++++++++++
-    * Calling `init` with a partial schema now calls `init_with_partial_schema`
-      instead throwing an error.
+    * :pr:`1100`: The behavior for `init` has changed. A full schema is a
+    schema that contains all of the columns of the dataframe it describes
+    whereas a partial schema only contains a subset. Before, only a full
+    schema was permitted by the `init` method so passing a partial schema
+    would error. Now, passing a partial schema to the `init` method is calls
```
gsheni, line 26:
```suggestion
    would error. Now, passing a partial schema to the `init` method calls
```

tuethan1999:
@tamargrey now that we've already crossed the breaking change boundary, does it still make sense that we ignore arguments along with a full schema? It actually seems like the EvalML use case wants to pass in a full schema and specify which columns to reinfer. Tbh,`init_with_full_schema` seems more of an internal call to me.

woodwork/tests/accessor/test_table_accessor.py
tamargrey: @tuethan1999 This test still needs to be updated to not just add a new element to the metadata dictionary, but to update the value of an element already in the dictionary. 

```python
sample_df.ww.init(table_metadata={'test': [1,2]})

...

test_df.ww._schema.metadata['test'].append(3)
```

That way it's checking that a deep copy has occured

tamargrey (changes requested):
At this point, I think I'm with you on making `init` use partial schema init and allow overriding even with a full schema. As long as @gsheni is okay with that added breaking change, I say go for it.

> Tbh,`init_with_full_schema` seems more of an internal call to me.

I could see instances where you are using woodwork and are performing some operation that shouldn't invalidate the schema but will create a new DataFrame. Those cases might want the rigidity of `init_with_full_schema` to confirm no unexpected changes happened. I don't see much harm in leaving it public, especially since it could be the only remnant of the former init with schema behavior.

Though if we do make it private, we can also make `init_with_partial_schema` private and just expose `init`

gsheni:
@tuethan1999 @tamargrey I am okay with the breaking change.

woodwork/table_accessor.py
```diff
@@ -84,70 +125,59 @@ def init(self,
...
+                existing_col_metadata[col_name] = col_schema.metadata
+                existing_use_standard_tags[col_name] = col_schema.use_standard_tags
+
+        # overwrite schema parameters with specified kwargs
+        logical_types = _infer_missing_logical_types(self._dataframe, logical_types, existing_logical_types)
+        column_descriptions = {**existing_col_descriptions, **(column_descriptions or {})}
+        column_metadata = {**existing_col_metadata, **(column_metadata or {})}
+        column_names = list(self._dataframe.columns)
+        normalized_tags = _normalize_standard_tags(use_standard_tags, column_names)
+        use_standard_tags = {**existing_use_standard_tags, **normalized_tags}
```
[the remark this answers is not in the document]
tuethan1999: fixed!

freddyaboulton (commented):
@tuethan1999 I think this breaks `ww.to_frame()` for pandas series?

I installed this branch locally and calling `to_frame()` on any ww series throws an error:

image

I really like what's going on in this branch though. I was able to update our Imputer, TargetEncoder, PCA, LDA, and DFSTransformer with this new init implementation for partial schemas rather than the evalml util that tried to do the same thing and I can get our unit tests to pass! (Thank you @tamargrey for the help in answering my questions). The thing about `to_frame` is the only "bug" I've seen with this.

I'll continue to chug along and report back!

tamargrey:
> I installed this branch locally and calling to_frame() on any ww series throws an error:

@freddyaboulton @tuethan1999 Interestingly, there is one way where `to_frame` doesn't error for me locally: when the Series is from a DataFrame.

```
y = sample_df.ww['id']
y.ww.to_frame()
```

However `test_series_methods_returning_frame` also doesn't error.

woodwork/table_accessor.py
```diff
@@ -930,11 +998,57 @@ def _check_schema(dataframe, schema):
...
+       1. use_standard tags
+       2. existing_use_standard_tags
+       3. default_use_standard_tag
+    """
+    if isinstance(use_standard_tags, bool):
+        use_standard_tags = {col_name: use_standard_tags for col_name in column_names}
+    else:
+        use_standard_tags = {**{col_name: default_use_standard_tag for col_name in column_names},
+                             **existing_use_standard_tags,
+                             **(use_standard_tags or {})}
```
tamargrey, line 1048: Nice!

tuethan1999:
@tamargrey @freddyaboulton thanks for bringing this bug up. Turns out it breaks when the series doesn't have a name! working on a fix

woodwork/table_accessor.py
```diff
@@ -80,74 +83,140 @@ def init(self,
...
+
+        extra_params = [key for key in kwargs]
+        if extra_params:
+            warnings.warn("A schema was provided and the following parameters were ignored: " + ", ".join(extra_params), ParametersIgnoredWarning)
+
+    def init_with_partial_schema(self,
+                                 schema: Optional[TableSchema] = None,
+                                 index: Optional[str] = None,
+                                 time_index: Optional[str] = None,
+                                 logical_types: Optional[Dict[ColumnName, Union[str, LogicalType, None]]] = None,
```
davesque, line 119: By doing `from __future__ import annotations` at the top as was mentioned in a previous review comment, you can use `str | LogicalType | None` instead of `Union[str, LogicalType, None]`.  We should probably adopt that convention for all union type annotations.

woodwork/table_accessor.py
```diff
@@ -80,74 +83,140 @@ def init(self,
...
+                                 already_sorted: Optional[bool] = False,
+                                 name: Optional[str] = None,
+                                 semantic_tags: Optional[Dict[ColumnName, Union[str, List[str], Set[str]]]] = None,
```
davesque, line 122: Similar to the earlier comment about `from __future__ import annotations` and being able to use `dict[...]`, you can do the same here with `list` and set`.  So `List[str]` and `Set[str]` would become `list[str]` and `set[str]` respectively.

woodwork/table_accessor.py
```diff
@@ -80,74 +83,140 @@ def init(self,
...
+                                 table_metadata: Optional[Dict] = None,
+                                 column_metadata: Optional[Dict[ColumnName, Dict]] = None,
+                                 use_standard_tags: Optional[Union[bool, UseStandardTagsDict]] = None,
+                                 column_descriptions: Optional[Dict[ColumnName, str]] = None,
+                                 column_origins: Optional[Dict[ColumnName, str]] = None,
+                                 validate: Optional[bool] = True,
```
davesque, line 128: Relating to some of the previous comments, for annotations like this where the explicit `Optional[...]` is necessary, it is equivalent and probably more concise to do the `from __future__ import annotations` as in previous comments and then do `bool | None` instead of `Optional[bool]`.

davesque (commented):
Awesome to see some type annotations here!  Also great that you identified the proper generics to use such as `Iterable[...]`, etc.  Had some comments about some further improvements that are possible with the annotations.

woodwork/table_accessor.py
```diff
@@ -38,28 +39,30 @@
 dd = import_or_none('dask.dataframe')
 ks = import_or_none('databricks.koalas')
 
+ColumnName = Hashable
```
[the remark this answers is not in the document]
tamargrey: yes, they're anything that can be a pandas column name (or for Dask or Koalas, what can be a dask or koalas column name)

woodwork/table_accessor.py
```diff
@@ -38,28 +41,34 @@
...
+    def init(self, **kwargs):
+        """Initializes Woodwork typing information for a DataFrame with a partial schema.
+        Logical type priority:
+        1. Types specified in ``logical_types``
+        2. Types specified in ``partial_schema``
+        3. Types inferred by ``ww.type_system.infer_logical_type``
+
+        Other Info priority:
+        1. Parameter passed in
+        2. Value specified in ``partial_schema``
```
tamargrey, line 62: image

Formatting here and above isn't coming through in the docs build

woodwork/table_accessor.py
```diff
@@ -73,81 +82,151 @@ def init(self,
...
+                                 already_sorted: bool = False,
+                                 name: str = None,
+                                 semantic_tags: Dict[ColumnName, Union[str, List[str], Set[str]]] = None,
+                                 table_metadata: dict = None,
+                                 column_metadata: Dict[ColumnName, dict] = None,
+                                 use_standard_tags: Union[bool, UseStandardTagsDict] = None,
+                                 column_descriptions: Dict[ColumnName, str] = None,
+                                 column_origins: Union[str, Dict[ColumnName, str]] = None,
+                                 validate: bool = True,
+                                 **kwargs) -> None:
```
tamargrey, line 135: After reading through @davesque's comments I think these changes should be possible to do to let us remove the `Dict` and `Union` imports (unless I'm misunderstanding where we can use `dict` vs `Dict`)

```suggestion
                                 schema: TableSchema = None,
                                 index: str = None,
                                 time_index: str = None,
                                 logical_types: dict[ColumnName, str | LogicalType | None] = None,
                                 already_sorted: bool = False,
                                 name: str = None,
                                 semantic_tags: dict[ColumnName, str | list[str] | set[str]] = None,
                                 table_metadata: dict = None,
                                 column_metadata: dict[ColumnName, dict] = None,
                                 use_standard_tags: bool | UseStandardTagsDict = None,
                                 column_descriptions: dict[ColumnName, str] = None,
                                 column_origins: str | dict[ColumnName, str] = None,
                                 validate: bool = True,
                                 **kwargs) -> None:
```

(can also make similar changes in `_infer_missing_logical_types` and `_merge_use_standard_tags` )

tamargrey (commented):
Looking good! Just want to make sure we're starting out our usage of python's typing really consistently!

woodwork/table_accessor.py
davesque: @tamargrey Actually, @tuethan1999 and I realized that it might be a bit too early to adopt those next gen type annotation features.  I think `Union` and `Dict` still make sense at this point.  Otherwise, we might end up having to use python 3.9 to do the type checking with mypy.  That might be okay, but it could just as well introduce some complexities to our CI pipeline that didn't exist before.  It might also introduce subtle issues that come up as a result of the version mismatch.  Not sure how that would look but it seems possible.

tamargrey: Ah, thanks for the explanation! I agree that being safe here makes sense!

woodwork/table_accessor.py
tuethan1999: The formatting should be fixed! Is it ok if I keep the kwargs as` **kwargs`?

tamargrey: looks good! Also I'm fine keeping the kwargs as they are in `init`!

woodwork/table_accessor.py
```diff
@@ -39,11 +39,7 @@
 dd = import_or_none('dask.dataframe')
 ks = import_or_none('databricks.koalas')
 
-from woodwork.typing import (
-    AnyDataFrame,
-    ColumnName,
-    UseStandardTagsDict
-)
+from woodwork.typing import AnyDataFrame, ColumnName, UseStandardTagsDict
```
tamargrey, line 42: this needs to be moved to be right below `from woodwork.type_sys.utils import _is_numeric_series, col_is_datetime`

tamargrey (approved):

tamargrey:
lgtm! Idk if anyone else wants to take a last look before merging, though

PR #1112, opened 2021-08-24
Add `AgeFractional` logical type 

tamargrey:
- Creates an `AgeFractional` logical type that has the dtype `float64` and adds to relevant locations in Woodwork
- Closes #876, closes #1076 

Note: went with the name `AgeFractional` instead of `FractionalAge` to keep the age ltypes next to one another in alphabetical order

PR #1113, opened 2021-08-25
Release v0.7.0

jeff-hernandez:
v0.7.0 Aug 25, 2021
===================
  * Enhancements
    * Add ``'passthrough'`` and ``'ignore'`` to tags in ``list_semantic_tags`` (#1094)
    * Add initialize with partial table schema  (#1100)
    * Apply ordering specified by the ``Ordinal`` logical type to underlying series (#1097)
    * Add ``AgeFractional`` logical type (#1112)

   @davesque,  @jeff-hernandez,  @tamargrey,  @tuethan1999

Breaking Changes
----------------
  * #1100:
    The behavior for ``init`` has changed. A full schema is a
    schema that contains all of the columns of the dataframe it describes
    whereas a partial schema only contains a subset. A full schema will also
    require that the schema is valid without having to make any changes to 
    the DataFrame. Before, only a full schema was permitted by the ``init`` 
    method so passing a partial schema would error. Additionally, any
    parameters like ``logical_types`` would be ignored if passing in a schema.
    Now, passing a partial schema to the ``init`` method calls the 
    ``init_with_partial_schema`` method instead of throwing an error. 
    Information from keyword arguments will override information from the
    partial schema. For example, if column ``a`` has the Integer Logical Type
    in the partial schema, it's possible to use the ``logical_type`` argument
    to reinfer it's logical type by passing ``{'a': None}`` or force a type by
    passing in ``{'a': Double}``. These changes mean that Woodwork init is less
    restrictive. If no type inference takes place and no changes are required
    of the DataFrame at initialization, ``init_with_full_schema`` should be
    used instead of ``init``. ``init_with_full_schema`` maintains the same
    functionality as when a schema was passed to the old ``init``.


tamargrey (approved):
lgtm!

docs/source/release_notes.rst
```diff
@@ -2,21 +2,25 @@
 
 Release Notes
 -------------
+.. **Future Release**
+..     * Enhancements
+..     * Fixes
+..     * Changes
+..     * Documentation Changes
+..     * Testing Changes
```
thehomebrewnerd, line 10: The styling for this commented out section should match what is shown in `release.md`:

```
.. Future Release
  ==============
    * Enhancements
    * Fixes
    * Changes
    * Documentation Changes
    * Testing Changes

```

PR #1122, opened 2021-08-31
Add Infer URL inference function

### LINKED ISSUE #1067: Add inference for URL logical type
- As a user, I wish I could use Woodwork to automatically infer if a feature is a url logical type. This would make it possible to automate the process of collecting new datasets that have url features as well as reduce the work the user has to do to specify correct types for their data. The latter is helpful for datasets with a large number of features

#### Code Example

```python
import pandas as pd
import woodwork as ww

X = pd.DataFrame({'url_feature': ["https://evalml.alteryx.com/en/stable/",
                                  "https://woodwork.alteryx.com/en/stable/guides/statistical_insights.html",
                                  "https://twitter.com/AlteryxOSS"]})
X.ww.init()

assert X.ww.logical_types['url_feature'].type_string == 'url'
```

ajaypallekonda:
- Added a URL inference function that compares the first 100k rows to see if they match the URL regex to infer URL type. Updated schema and tests.
- Closes #1067


docs/source/guides/setting_config_options.ipynb
```diff
@@ -138,29 +138,27 @@
...
-    "name": "ipython",
-    "version": 3
-   },
-   "file_extension": ".py",
-   "mimetype": "text/x-python",
    "name": "python",
-   "nbconvert_exporter": "python",
-   "pygments_lexer": "ipython3",
-   "version": "3.7.11"
+   "version": ""
```
davesque, line 159: Were these changes automatically included by jupyter?  If so, it might make sense to just leave them out.  Thoughts on this @gsheni @tuethan1999 ?

woodwork/type_sys/inference_functions.py
```diff
@@ -94,3 +94,23 @@ def email_address_func(series: pd.Series) -> bool:
...
+        series_match_method = series.str.match
+    except (AttributeError, TypeError):
+        # This can happen either when the inferred dtype for a series is not
+        # compatible with the pandas string API (AttributeError) *or* when the
+        # inferred dtype is not compatible with the string API `match` method
+        # (TypeError)
+        return False
+    matches = series_match_method(pat=regex)
+
+    return matches.sum() == matches.count()
```
davesque, line 116: Since this inference func is really just a copy of the email inference func, I wonder if it makes sense to abstract this out into a class.  So something like this:
```python
class InferWithRegex:
    def __init__(self, get_regex: Callable[[], str]):
        self.get_regex = get_regex

    def __call__(self, series: pd.Series) -> bool:
        regex = self.get_regex()

        # Includes a check for object dtypes
        if not pdtypes.is_string_dtype(series.dtype):
            return False

        try:
            series_match_method = series.str.match
        except (AttributeError, TypeError):
            # This can happen either when the inferred dtype for a series is not
            # compatible with the pandas string API (AttributeError) *or* when the
            # inferred dtype is not compatible with the string API `match` method
            # (TypeError)
            return False
        matches = series_match_method(pat=regex)

        return matches.sum() == matches.count()

email_address_func = InferWithRegex(lambda: ww.config.get_option('email_inference_regex'))
url_func = InferWithRegex(lambda: ww.config.get_option('url_inference_regex'))
```
You would probably want to use the lambdas (or some kind of deferred config lookup) to avoid having to worry about what order things are done in.  For example, a user might happen to import the module that these funcs are defined in before they have a chance to override the config setting.

davesque (commented):
Nice, awesome work.  Mostly I just have one thought about code duplication.

docs/source/guides/setting_config_options.ipynb
ajaypallekonda: Yeah I think so, I only added the text part.

davesque: Yeah, I mean they probably don't matter that much.  If anything, I'm mostly just curious what the effect of the "hash" part is.

woodwork/type_sys/inference_functions.py
ajaypallekonda: It does make more sense to abstract it into a class. I used your suggestion and pushed new changes.

docs/source/guides/setting_config_options.ipynb
davesque: Actually, these things probably don't matter.  We can just include them.

davesque (approved):
Yeah, looks good to me!  Thanks for adding that!

docs/source/release_notes.rst
```diff
@@ -5,12 +5,14 @@ Release Notes
 .. Future Release
```
gsheni, line 5: Uncomment this
```suggestion
Future Release
```

docs/source/release_notes.rst
```diff
@@ -5,12 +5,14 @@ Release Notes
...
   ===============
     * Enhancements
+        * Add support for automatically inferring the ``URL`` logical type (:pr:`1122`)
     * Fixes
     * Changes
     * Documentation Changes
     * Testing Changes
 
 .. Thanks to the following people for contributing to this release:
```
gsheni, line 14:
```suggestion
    Thanks to the following people for contributing to this release:
```

PR #1127, opened 2021-09-03
Add column accessor attribute to check for nullable logical type

### LINKED ISSUE #1110: Add nullable as a first-level attribute to WoodworkColumnAccessor and ColumnSchema
- We use pandas dtypes for the underlying representation of our Logical Types. Some of these are nullable and some of them are not.
- We can add a `nullable` attribute to further make this property clear. We can add it to our WoodworkColumnAccessor and ColumnSchema
- We can also add an `is_nullable` function on the ColumnSchema
- This may be a breaking change might break serialization and schema. It should be documented properly in the changelog.
- We should add an example to our docs on how to use this new property and function:
  - https://woodwork.alteryx.com/en/latest/guides/logical_types_and_semantic_tags.html?highlight=ColumnSchema#ColumnSchema-objects

jeff-hernandez:
Closes #1110


woodwork/column_schema.py
```diff
@@ -141,6 +146,15 @@ def is_boolean(self):
...
         return ltype_class == Boolean or ltype_class == BooleanNullable
 
+    def is_nullable(self, series_type):
+        """Whether the logical type contains a nullable physical type.
+
+            Args:
+                series_type (Series): Type of series instance
+        """
+        dtype = self.logical_type._get_valid_dtype(series_type)
+        return dtype in _NULLABLE_PHYSICAL_TYPES
```
jeff-hernandez, line 156: > We can also add an `is_nullable` function on the ColumnSchema

@gsheni  do we want this to be a function of a series type?

jeff-hernandez:
Is there a particular case where these changes can break serialization or the column schema?

woodwork/logical_types.py
```diff
@@ -503,3 +503,15 @@ class PostalCode(LogicalType):
     primary_dtype = 'category'
     backup_dtype = 'string'
     standard_tags = {'category'}
+
+
+_NULLABLE_PHYSICAL_TYPES = {
+    'Int64',
```
gsheni, line 509: You can add more nullable types:

```suggestion
    'Int8',
    'Int16',
    'Int32',
    'Int64',
    'Float8'
    'Float16'
    'Float32'
    'Float64'
```

woodwork/logical_types.py
```diff
@@ -503,3 +503,15 @@ class PostalCode(LogicalType):
...
+    'boolean',
+    'category',
+    'datetime64[ns]',
+    'float64',
```
gsheni, line 513:
```suggestion
    'float8',
    'float16',
    'float32',
    'float64',
    'float128',
```

woodwork/column_schema.py
gsheni: @rwedge @tamargrey @thehomebrewnerd @davesque can better answer this. I am not sure.

thehomebrewnerd: Just glancing through this quickly, so I might be missing something, but generally we are expecting that a schema won't have any knowledge of the data to which it is attached (in some cases it won't have data - like in FT). So if we need to use the series type to determine if a column is nullable, I'm thinking this shouldn't be on the schema, but only on the accessor.

jeff-hernandez: If we're all on the same page here, I'll go ahead and not add  `is_nullable` on the ColumnSchema

davesque:
@thehomebrewnerd @gsheni Unless I'm mistaken, after looking at the original issue and the discussion on this PR, it doesn't look like there's anything left to do on this other than to just get approvals.  Please let me know if there's anything else that needs to happen and I'll get it taken care of.

thehomebrewnerd (commented):
I just have a couple questions about the approach used here for now.

Is there a reason to use the physical types for determining whether a column is nullable or not? Many of the dtypes in `_NULLABLE_PHYSICAL_TYPES` are not used by Woodwork and would not be present on a series that has Woodwork initialized. Should we base this check off the LogicalType instead? If we did that this could also be a property of the `ColumnSchema`.

We could also define `nullable` as a property of the LogicalType which could be propagated to the schema and accessor if needed.

davesque:
@thehomebrewnerd I kinda feel like nullability would ideally not be related to data typing.  However, since we're wed to pandas at this point I think looking at the pandas dtype is the most direct way of doing it.  The pandas dtype feels more like the source of truth since it has the final word on supporting a missing value sentinel.

davesque:
Part of the reason I say this is because of how arrow handles missing data, which I've mentioned a few times.  An array's type doesn't affect nullability in arrow.  That's governed by the presence of a null bitmap that indicates which values are valid vs. invalid.  This feels more right to me since it acknowledges that data may always be missing for this or that reason.  That has more to do with the context that some array is being used in and less to do with the way the underlying data is encoded.

In any case, it just goes to show how nullability may not have much to do with typing.  So I feel better about having our column accessor API just defer to the underlying storage lib.

It's maybe not super relevant to the general discussion here, but there's a pretty well-known article written by Wes McKinney that goes into some of the design trade-offs with Pandas vs. Arrow: https://wesmckinney.com/blog/apache-arrow-pandas-internals/.  Section 4 discusses the nullability thing.  If anything, it's useful supplemental reading for some of these topics.

thehomebrewnerd:
@davesque That's fair. I guess this could always change, but at the moment we have a 1-to-1 relationship between LogicalTypes and pandas dtypes, so if we know the LogicalType we also know what the series dtype will be (without needing to access the series directly), so either could be used  for this purpose I think.

In any case, I'm fine with leaving this as it is since it doesn't seem we have a need to have this property available on the LogicalType or the ColumnSchema at the moment.

As a follow up question, do we know if pandas has a way to determine if a given dtype supports missing values? That would be even better than creating and maintaining our own list I think, but if the don't have anything then we need to define this on our own.

davesque:
> As a follow up question, do we know if pandas has a way to determine if a given dtype supports missing values? That would be even better than creating and maintaining our own list I think, but if the don't have anything then we need to define this on our own.

Yeah, I wondered about this myself.  I sort of assumed, since it hadn't been mentioned yet in the history of the PR, that pandas must not have such an API.  But I should check to be sure.

davesque:
@thehomebrewnerd Some cursory googling and code browsing in the pandas repo seems to indicate that there is no such API.  I'm not seeing anything like it in the implementation of the `pandas.api.types` module.  Googling for things like "check is pandas series takes null values" also doesn't turn anything up.  I'm surprised, actually.

thehomebrewnerd:
@davesque I did a quick search as well and didn't turn up anything. Also looked at the properties on the dtype object but didn't see anything there that looked like what we want.

davesque:
@gsheni @thehomebrewnerd I determined that a few of those types in the nullability set are not supported
```python
In [1]: from woodwork.logical_types import _NULLABLE_PHYSICAL_TYPES
WARNING:root:'PYARROW_IGNORE_TIMEZONE' environment variable was not set. It is required to set this environment variable to '1' in both driver and executor sides if you use pyarrow>=2.0.0. Koalas will set it for you but it does not work if there is a Spark context already launched.

In [2]: import pandas as pd

In [3]: for dtype in _NULLABLE_PHYSICAL_TYPES:
   ...:     try:
   ...:         pd.Series([None], dtype=dtype)
   ...:     except Exception as e:
   ...:         print(dtype)
   ...:         print(repr(e))
   ...:
Float8
TypeError("data type 'Float8' not understood")
Float16
TypeError("data type 'Float16' not understood")
float8
TypeError("data type 'float8' not understood")
```

docs/source/guides/logical_types_and_semantic_tags.ipynb
```diff
@@ -676,6 +676,51 @@
...
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "sexual-adoption",
+   "metadata": {},
+   "source": [
+    "## Checking for nullable logical types\n",
+    "\n",
+    "Some logical types support having null values in the underlying data while others do not. The logical types that do support null values are listed below:\n",
```
thehomebrewnerd, line 687: Just wondering if we should list this separately here to document which logical types support null values, or should we include this information in the discussion of each individual logical type that happens earlier in the document? Having the list here means users would have to look in two different places to find all the info about a logical type. Also, if we left this list, maybe it would be better to list the types that do not support null values as that would be significantly shorter.

We could leave the cell demonstrating how to access the property, but not sure this long list is necessary?

thehomebrewnerd (commented):
Code looks fine - just one though/comment related to the documentation update. 

Also - we should add this new property to the API reference for `WoodworkColumnAccessor`

docs/source/guides/logical_types_and_semantic_tags.ipynb
tamargrey: By putting the nullable property on the column accessor, I'd argue that we're specifically avoiding associating nullability with Logical Types. We're saying that nullability is only a relevant concept for a Woodwork column (series + column schema) and not any one of the pieces that make one up (logical type, physical type, etc).

It just so happens that there _is_ a relationship between nullability and logical type, but I think it's misleading to only explain `series.ww.nullable` in the concept of Logical Types when that's not how users will use it. 

The use case for telling users which logical types are nullable could be when a user is picking which logical types they use at init based on which are nullable or not. If that's the main use case, having nullability exist on an already created woodwork column is not useful for that purpose. 

The implementation here is useful for determining what you can do with your data once you know whether its nullable or not. For example: you'd check a column's nullability when deciding whether you need preprocessing to impute nulls. But for that, a user wouldn't need to know the logical type (the logical type is important for determining how to impute, but not whether to impute).

So I don't know that this concept of a column's nullability needs to be explained in the context of logical types. What makes sense to me as a way of explaining nullability in Woodwork with the implementation here is to focus less on how one determines whether a column is nullable, but what you do with that information. 

So maybe the docs entry doesn't belong in this guide, since it might look something like 

> A column in woodwork may contain null values. This can have an impact on how the column is used. Instead of checking whether a column has any null values, the `nullable` property of a woodwork column can be used to check if a column can contain null values in the first place. 
... example showing series.ww.nulalble usage ...

davesque: Hmm...I just did a couple things.  I did add nullability information to each type's description in the docs.  If we go with what @tamargrey is saying and try to de-emphasize the relationship between logical types and nullability (which I'm in favor of doing), I can take all those individual nullability flags out.

I also updated the description of how nullability is determined to more transparently indicate the importance of the underlying pandas dtype.  There's no longer an exhaustive list of the nullable logical types but a description of how nullability is determined.

davesque: @thehomebrewnerd I also updated the API docs to include the `WoodworkColumnAccessor.nullable` property.

woodwork/column_accessor.py
```diff
@@ -106,6 +106,13 @@ def _series(self):
     def schema(self):
         return copy.deepcopy(self._schema)
 
+    @property
+    @_check_column_schema
+    def nullable(self):
+        """Whether the column schema contains a nullable logical type."""
```
tamargrey, line 112: This might be better as 

> Whether the column can contain null values 

Or something that focuses less on the column schema and logical types.

tamargrey (approved):
Looks good, just a small comment about the docstring

woodwork/column_accessor.py
davesque: Yep, I agree that's better.

gsheni (approved):

PR #1129, opened 2021-09-03
Added method to get valid mutual information columns

### LINKED ISSUE #254: Add method to get valid mutual info column
Now that we skip rows for mutual information between a column and itself, we should have a way for users to determine which columns are valid for mutual info.

`dt.get_valid_mi_columns` should return a list of the columns in the datatable that are not fully null and are Boolean, Categorical, or Numeric

ajaypallekonda:
- Added `get_valid_mi_columns` method to list columns that have valid logical types for mutual information calculation
- Closes #254 

-----


woodwork/statistics_utils.py
```diff
@@ -300,6 +300,30 @@ def _get_mutual_information_dict(dataframe, num_bins=10, nrows=None, include_ind
     return mutual_info
 
 
+def _get_valid_mi_columns(dataframe, include_index=False):
+    """Returns a list of column names a list of columns from the DataFrame with valid
+        Logical Types that support mutual information.
```
davesque, line 305: I think there's a typo here.
```suggestion
    """Returns a list of columns from the DataFrame with valid
        Logical Types that support mutual information.
```

woodwork/statistics_utils.py
```diff
@@ -300,6 +300,30 @@ def _get_mutual_information_dict(dataframe, num_bins=10, nrows=None, include_ind
...
+            included as long as its LogicalType is valid for mutual information calculations.
+            If False, the index column will not have mutual information calculated for it.
+            Defaults to False.
+
+    Returns:
+        list: A list of column names that have valid Logical Types that support
+        mutual information.
+    """
+    valid_types = get_valid_mi_types()
+    valid_columns = [col_name for col_name, col in dataframe.ww.columns.items() if type(col.logical_type) in valid_types]
```
davesque, line 320: This might be a more canonical way to do this check:
```suggestion
    valid_columns = [col_name for col_name, col in dataframe.ww.columns.items() if isinstance(col.logical_type, valid_types)]
```

woodwork/table_accessor.py
```diff
@@ -841,6 +842,23 @@ def mutual_information(self, num_bins=10, nrows=None, include_index=False, callb
         mutual_info = self.mutual_information_dict(num_bins, nrows, include_index, callback)
         return pd.DataFrame(mutual_info)
 
+    def get_valid_mi_columns(self, include_index=False):
+        """Retrieves a list of columns from the DataFrame with valid Logical Types that support mutual
+        information. Logical Types that support mutual information are as follows:  Age, AgeFractional,
+        AgeNullable, Boolean, BooleanNullable, Categorical, CountryCode, Datetime, Double, Integer,
+        IntegerNullable, Ordinal, PostalCode, and SubRegionCode.
```
davesque, line 849: @gsheni @tuethan1999 Should we be explicitly listing out the types here in the docstring?  Seems like something that could potentially become out of date.

woodwork/tests/accessor/test_statistics.py
```diff
@@ -245,6 +245,25 @@ def __call__(self, update, progress, total, unit, time_elapsed):
...
+    valid_columns = df_mi.ww.get_valid_mi_columns()
+    mi = df_mi.ww.mutual_information()
+    valid_mi_columns = (mi.column_1.append(mi.column_2)).unique()
+
+    assert (valid_columns == valid_mi_columns).all()
+
+
+def test_get_valid_mi_columns_with_index(sample_df):
+    sample_df.ww.init(index='id')
+    mi = sample_df.ww.get_valid_mi_columns()
```
davesque, line 259: If this line is specifically supposed to produce an `mi` result where the index is not included, we should probably do this:
```suggestion
    mi = sample_df.ww.get_valid_mi_columns(include_index=False)
```
Even if that is already the default value, it makes the intentions of the test case clear.

woodwork/tests/accessor/test_statistics.py
```diff
@@ -245,6 +245,25 @@ def __call__(self, update, progress, total, unit, time_elapsed):
...
+
+    assert not ('id' in mi)
```
davesque, line 261: This is a bit more typical:
```suggestion
    assert 'id' not in mi
```

davesque (commented):
Just a few things I noticed.

woodwork/table_accessor.py
gsheni: I think it's a minor thing. We can keep updating the docstring. 
We also have a [get_valid_mi_types](https://woodwork.alteryx.com/en/latest/generated/woodwork.utils.get_valid_mi_types.html#woodwork.utils.get_valid_mi_types) that we could potentially reference in the docstring.

tuethan1999: I also think we should probably reference [get_valid_mi_types](https://woodwork.alteryx.com/en/latest/generated/woodwork.utils.get_valid_mi_types.html#woodwork.utils.get_valid_mi_types) instead of listing out the types

woodwork/statistics_utils.py
ajaypallekonda: I get this error with this change "TypeError: isinstance() arg 2 must be a type or tuple of types". I think its because `valid_types` is list of valid types and not just a tuple. I'll keep this line of the code the same for now

davesque: You can just do `tuple(valid_types)`.  Or you can do `valid_types = tuple(get_valid_mi_types())`.

ajaypallekonda: Gotcha! I just fixed that.

woodwork/statistics_utils.py
```diff
@@ -300,6 +300,30 @@ def _get_mutual_information_dict(dataframe, num_bins=10, nrows=None, include_ind
...
+            included as long as its LogicalType is valid for mutual information calculations.
+            If False, the index column will not have mutual information calculated for it.
+            Defaults to False.
+
+    Returns:
+        list: A list of column names that have valid Logical Types that support
+        mutual information.
+    """
+    valid_types = tuple(get_valid_mi_types())
+    valid_columns = [col_name for col_name, col in dataframe.ww.columns.items() if type(col.logical_type) in valid_types]
```
davesque, line 320: Thanks for making the `valid_types = tuple(get_valid_mi_types())` change.  Were you also going to make the `isinstance(col.logical_type, valid_types)` change?

davesque (approved):
Nice, LGTM!  Thanks for making all those changes I mentioned!

woodwork/table_accessor.py
```diff
@@ -778,9 +779,8 @@ def rename(self, columns, inplace=False):
     def mutual_information_dict(self, num_bins=10, nrows=None, include_index=False, callback=None):
         """
         Calculates mutual information between all pairs of columns in the DataFrame that
-        support mutual information. Logical Types that support mutual information are as
-        follows:  Age, AgeFractional, AgeNullable, Boolean, BooleanNullable, Categorical, CountryCode, Datetime, Double,
-        Integer, IntegerNullable, Ordinal, PostalCode, and SubRegionCode
+        support mutual information. Use get_valid_mi_types to see which Logical Types support
```
gsheni, line 782: I think you can put in docstrings with code like this. You can actually run them as tests

```suggestion
        support mutual information. Use get_valid_mi_types to see which Logical Types support
        >>> from woodwork.utils import get_valid_mi_types
        >>> get_valid_mi_types()
        [Age, AgeFractional, AgeNullable, Boolean, BooleanNullable, 
         Categorical, CountryCode,Datetime, Double, Integer, IntegerNullable, 
         Ordinal, PostalCode, SubRegionCode]
```

- You will need to make sure setup.cfg has this
```
addopts = --ignore "boilerplates/" --doctest-modules
```

- We do this in premium primitives
- https://github.com/alteryx/premium-primitives/blob/main/setup.cfg#L4
- https://github.com/alteryx/premium-primitives/blob/main/premium_primitives/age_over_18.py#L22

gsheni (approved):

PR #1132, opened 2021-09-09
Update get_invalid_schema_message to prevent nested looping

thehomebrewnerd:
- Update get_invalid_schema_message to prevent nested looping

Updates the `get_invalid_schema_message` function to remove nested looping that was being caused by calling `schema.logical_types` inside of a for loop looping over all the columns in a dataframe. Moved `schema.logical` types outside of the for loop so it is now only called once.


tamargrey:
This change looks good and makes sense to me. It also makes me think that we should probably look for other situations where we might be needlessly recreating `schema.logical_types` or `schema.semantic_tags` across Woodwork and the places that have started using it.

gsheni (approved):

PR #1137, opened 2021-09-10
Update Datetime.transform to use default number rows in call to infer_datetime_format

### LINKED ISSUE #1135: Datetime inference takes too long on large dataframes
Type inference on datetime columns takes a very long time. This can be especially problematic for large dataframes that contain several datetime columns.

As an example, initializing Woodwork on this single column dataframe takes over 15 seconds based on my local testing.

```python
import pandas as pd
import woodwork as ww

df = pd.DataFrame({'dates': ['2017-01-01']*1000000})
df.ww.init()
```

I believe a majority of this time is in the `_infer_datetime_format` utility function, but I have not confirmed for sure.

For comparison, type inference on a column of integers of the same length takes about 3 ms and inference on a column of strings takes about 275 ms.

We should further investigate why datetime inference is so slow and implement improvements to speed it up as much as possible to prevent long inference times on large datasets containing datetime columns.

thehomebrewnerd:
- Update Datetime.transform to use default number rows in call to infer_datetime_format
- Closes #1135 

Update `Datetime.transform` to use default number of rows (100) when inferring the datetime format, instead of the previous approach of using 10% of the rows. This is being done to improve inference performance for datasets with a large number of rows.

### DIFF
--- a/woodwork/logical_types.py
+++ b/woodwork/logical_types.py
@@ -199,8 +199,7 @@ class Datetime(LogicalType):
         """Converts the series data to a formatted datetime. Datetime format will be inferred if datetime_format is None."""
         new_dtype = self._get_valid_dtype(type(series))
         if new_dtype != str(series.dtype):
-            sample_length = len(series.index) // 10 or len(series.index)
-            self.datetime_format = self.datetime_format or _infer_datetime_format(series, n=sample_length)
+            self.datetime_format = self.datetime_format or _infer_datetime_format(series)
             try:
                 if _is_dask_series(series):
                     name = series.name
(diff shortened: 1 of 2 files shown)


woodwork/logical_types.py
```diff
@@ -199,8 +199,7 @@ def transform(self, series):
         """Converts the series data to a formatted datetime. Datetime format will be inferred if datetime_format is None."""
         new_dtype = self._get_valid_dtype(type(series))
         if new_dtype != str(series.dtype):
```
tamargrey, line 201: Noticed that the fact that the `_infer_datetime_format` call happens after this check means that the datetime format is not always saved. Opened an issue #1138

davesque (approved):
LGTM!

tamargrey (approved):
lgtm!

PR #1153, opened 2021-09-23
Adding Jupyter notebook cleaner

bchen1116:
Add in 2 additional commands for `make lint` and `make lint-fix` to remove any executed code from the jupyter notebooks

Running two example notebooks leave these changes:
git status
Example diff:
index b576f60 5bccdee 100644

We then run `make lint`:
make lint

Which tells us to run `make lint-fix`:
make lint-fix

**Note:** The remaining change is the change of python version. Since there is no current python version standardizer, this should be ok
Pasted Graphic 6

**Another Note**: The linter will remove the `id` value from each cell. Not all notebooks have these ID values, but I don't think removing these will be detrimental
image

### DIFF
--- a/docs/notebook_cleaner.py
+++ b/docs/notebook_cleaner.py
@@ -17,11 +17,27 @@ def _get_ipython_notebooks(docs_source):
     return notebooks
 
 
+def _check_delete_empty_cell(notebook, delete=True):
+    with open(notebook, "r") as f:
+        source = json.load(f)
+        cell = source["cells"][-1]
+        if cell["cell_type"] == "code" and cell["source"] == []:
+            # this is an empty cell, which we should delete
+            if delete:
+                source["cells"] = source["cells"][:-1]
+            else:
+                return False
+    if delete:
+        json.dump(source, open(notebook, "w"), ensure_ascii=False, indent=1)
+    else:
+        return True
+
+
 def _check_execution_and_output(notebook):
     with open(notebook, "r") as f:
         source = json.load(f)
         for cells in source["cells"]:
-            if cells["cell_type"] == "code" and cells["execution_count"] != None:
+            if cells["cell_type"] == "code" and (cells["execution_count"] is not None or cells['outputs']!= []):
                 return False
     return True
 
@@ -30,7 +46,7 @@ def _fix_execution_and_output(notebook):
     with open(notebook, "r") as f:
         source = json.load(f)
         for cells in source["cells"]:
-            if cells["cell_type"] == "code" and cells["execution_count"] != None:
+            if cells["cell_type"] == "code" and cells["execution_count"] is not None:
                 cells["execution_count"] = None
                 cells["outputs"] = []
         source["metadata"]["kernelspec"]["display_name"] = "Python 3"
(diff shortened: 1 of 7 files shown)


requirements.txt
```diff
@@ -1,2 +1,3 @@
 pandas>=1.2.5
 scikit-learn>=0.22
+click>=7.1.2
```
gsheni, line 3: Can we make this a dev-requirement?

docs/source/release_notes.rst
```diff
@@ -7,12 +7,13 @@ Future Release
     * Enhancements
     * Fixes
     * Changes
+        * Added notebook standardizer to remove executed outputs (:pr:`1153`)
```
thehomebrewnerd, line 10: I'm thinking we should include this in the `Testing Changes` section instead of here since it will really be used as part of the CI checks to make sure the documents are in good order.

docs/notebook_cleaner.py
```diff
@@ -0,0 +1,86 @@
...
+    for root, _, filenames in os.walk(docs_source):
+        if any(dir_ in root for dir_ in directories_to_skip):
+            continue
+        for filename in filenames:
+            if filename.endswith(".ipynb"):
+                notebooks.append(os.path.join(root, filename))
+    return notebooks
+
+
+def _check_execution_and_output(notebook):
```
thehomebrewnerd, line 20: This function name mentions outputs, but I don't see that we are checking outputs anywhere. Is it ever possible to get something in the cell outputs with an execution count of None?

docs/notebook_cleaner.py
```diff
@@ -0,0 +1,86 @@
...
+    with open(notebook, "r") as f:
+        source = json.load(f)
+        for cells in source["cells"]:
+            if cells["cell_type"] == "code" and cells["execution_count"] != None:
```
thehomebrewnerd, line 24: Should we be using `is not None` here and below instead of `!= None`?

thehomebrewnerd (commented):
Overall looks pretty good. Just a couple comments/questions.

Also, one other common issue we run into is having blank cells in the notebook - especially at the end of the file. Is that a check we could/should add here as well?

docs/notebook_cleaner.py
```diff
@@ -0,0 +1,86 @@
...
+                return False
+    return True
+
+
+def _fix_execution_and_output(notebook):
+    with open(notebook, "r") as f:
+        source = json.load(f)
+        for cells in source["cells"]:
+            if cells["cell_type"] == "code" and cells["execution_count"] != None:
```
tamargrey, line 33: same comment as nate's above--can we do `is not None`?

docs/notebook_cleaner.py
bchen1116: Good call! I don't think it would be possible unless a user goes in and edits the notebooks via text-editor. I can add a check for outputs, however

bchen1116:
@thehomebrewnerd Added in support of removing the last cell if it's empty!

Example usage here:
diff --git adocssourceguidesfeature_descriptions ipynb bdocssourceguidesfeature_descriptions ipynb
make lint
make lint-fix
index 3fd8d6d cce52a8 100644

docs/notebook_cleaner.py
```diff
@@ -60,26 +89,42 @@ def cli():
...
         executed_notebooks = ["\t" + notebook for notebook in executed_notebooks]
         executed_notebooks = "\n".join(executed_notebooks)
         raise SystemExit(
             f"The following notebooks have executed outputs:\n {executed_notebooks}\n"
             "Please run make lint-fix to fix this."
         )
+    if empty_cells:
+        empty_cells = ["\t" + notebook for notebook in empty_cells]
+        empty_cells = "\n".join(empty_cells)
+        click.echo(
```
thehomebrewnerd, line 124: Should this also be a `raise SystemExit` to generate a lint error?

thehomebrewnerd (approved):
Looks good! This will be a nice addition!

PR #1184, opened 2021-11-03
Make box plot low/high indices/values optional to return

### LINKED ISSUE #1168: Make box plot low/high indices/values optional to return
In a series with many outliers, returning all of the low and high indices and the low and high values can make the results of `box_plot_dict` difficult to read, as the low bound, the high bound, and the quantiles are such small parts of the output in that case.

We should add a parameter to `series.ww.box_plot_dict` that makes the inclusion of `low_indices`, `high_indices`, `low_values`, and `high_values` in the resulting dictionary optional, defaulting to including them in the returned dictionary.

```
# To avoid getting the indices and values returned set the parameter to False
series.ww.box_plot_dict(include_indices_and_values=False)
# defaults to True - if no parameter is passed in, returns the indices and values
series.ww.box_plot_dict()
```

tamargrey:
- Adds a parameter to `box_plot_dict` that determines whether the lists of outlier values and their indices are included in the returned dictionary
- Closes #1168

### DIFF
--- a/woodwork/tests/accessor/test_statistics.py
+++ b/woodwork/tests/accessor/test_statistics.py
@@ -1256,6 +1256,13 @@ def test_box_plot_optional_return_values(outliers_df):
     has_outliers_series = outliers_df["has_outliers"]
     has_outliers_series.ww.init()
 
-    box_plot_info = has_outliers_series.ww.box_plot_dict(include_indices_and_values=False)
+    has_outliers_box_plot_info = has_outliers_series.ww.box_plot_dict(include_indices_and_values=False)
 
-    assert {"low_bound", "high_bound", "quantiles"} == set(box_plot_info.keys())
+    assert {"low_bound", "high_bound", "quantiles"} == set(has_outliers_box_plot_info.keys())
+
+    no_outliers_series = outliers_df["no_outliers"]
+    no_outliers_series.ww.init()
+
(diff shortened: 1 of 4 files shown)


woodwork/tests/accessor/test_statistics.py
```diff
@@ -1250,3 +1250,19 @@ def test_box_plot_quantiles_errors(outliers_df):
     error = "quantiles must be a dictionary."
     with pytest.raises(TypeError, match=error):
         series.ww.box_plot_dict(quantiles=1)
+
+
+def test_box_plot_optional_return_values(outliers_df):
```
thehomebrewnerd, line 1255: Do you think we should add an explicit condition to this test to check that the extra keys are included when `include_indices_and_values=True`? We sort of implicitly test that in other tests, but if the purpose of this is to check that the right keys are returned maybe it would be good to check both the `True` and the `False` case.

thehomebrewnerd (commented):
Looks fine to me overall. One test suggestion, and looks like some linting fixes are needed.

woodwork/tests/accessor/test_statistics.py
tamargrey: Added checks with the explicit `include_indices_and_values=True`!

thehomebrewnerd (approved):
Looks good to me!

davesque (approved):
LGTM after Nate's comments.

gsheni (approved):

PR #1193, opened 2021-11-16
Update install page with correct min optional dependencies

gsheni:


rwedge:
Looks good

I wonder if in the future we could generate this at build time from our requirements files or from the library, if we stored info about these option min versions in the library code

rwedge (approved):

Issue #1165, opened 2021-10-14
Datetime format inference may not recognize day/month/year format


freddyaboulton:
#### Code Sample, a copy-pastable example to reproduce your bug.

On ww 0.8.2:

```python
import pandas as pd
import woodwork as ww

df = pd.DataFrame({"date": ["01/10/2021", "01/10/2021"] * 2000 + ["31/10/2021", "30/10/2021"] * 2000})
df.ww.init()
assert df['date'].isna().any()
```

This warning will also come up
```
TypeConversionWarning: Some rows in series 'date' are incompatible with datetime format '%m/%d/%Y' and have been replaced with null values.
```

Doing the following  fixes the issue:

```python
df.ww.init(logical_types={"date": ww.logical_types.Datetime(datetime_format="%d/%m/%Y")})
```

But I still think it's weird that it originally inferred the format as `%m/%d/%Y` when the right format is `%d/%m/%Y`

gsheni:
Related issue: https://github.com/alteryx/woodwork/issues/1152

Would a temporary patch for this be to sample more rows to infer the datetime format?

thehomebrewnerd:
@freddyaboulton Created a spike to investigate this further, likely during our next sprint, and determine a path forward.

https://github.com/alteryx/woodwork/issues/1167

PR #1202, opened 2021-11-24
Allow frequency inference on temporal (Datetime, Timedelta) columns of Woodwork DataFrame

### LINKED ISSUE #1155: Store interval frequency of data points in Datetime column
In certain instances, it is useful to know the frequency of observations in a Datetime column (1 min, 1 day, 5 days) where the interval of time between data points in the column is consistent. These frequencies may be known by users or inferred from the data. 

As a user, it could be useful for down-the-line operations that will try to infer that frequency or otherwise require it be passed in as an argument (for example: pandas' [to_period](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.dt.to_period.html)) if a Woodwork DataFrame had knowlege of a Datetime column's frequency. 

We should consider storing a column's frequency as an [offset string](https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#timeseries-offset-aliases) or offset object.

#### Code Example

Storing the frequency in the `Datetime` logical type:

```python
df.ww.init(logical_types={'dates': Datetime(frequency='1D')})
assert df.ww.logical_types['dates'].frequency == '1D'
```

This may be a relevant concept only for time index, so it could also be a value that's inferred via `pd.infer_freq` just for the time index column:

```python
df.ww.init(time_index='dates')})
assert df.ww.logical_types['dates'].frequency == '1D'
```

Additionally, since Woodwork allows numeric time indexes, is this a concept that can or should exist for a numeric time index?

Questions to answer:
- What does it look like if a column has no discernable frequency and one is assigned or attempted to be inferred? 
- What if a few columns are not compliant with the inferred or assigned frequency?
- Does it matter if the data is sorted or not?
- Can users assign a frequency after woodwork initialization?

tamargrey:
- Adds an accessor method `df.ww.infer_temporal_frequencies` that calls `pd.infer_freq` on all datetime or timedelta columns. A `temporal_columns` parameter can also be used to specify specific columns for which to infer frequency (for example, we can just call it on the time index).
- Closes #1155

To Be Decided:
- [x] Should we sort each column before passing it to `infer_freq`? Or add a note that users should make sure data is sorted (which could be hard with multiple datetime columns)
- [x] Do we limit the number of rows passed to `infer_freq`? We'd need to take a sequential slice from the dataframe, but it may be more beneficial to take a subset to determine the frequency to minimize the chance of an outlier missing row. 
  - [x] An alternative would be to take multiple subsets of the data and infer the frequency for each and use the result that shows up the most as the column's frequency. 
- [x] Neither dask nor koalas have an `infer_freq` implementation, and calling `pd.infer_freq` on koalas or dask dataframes raises errors. Should we be converting the data to pandas to allow frequency inference for them, or just note that this is only for pandas dataframes?

### DIFF
--- a/woodwork/statistics_utils.py
+++ b/woodwork/statistics_utils.py
@@ -649,21 +649,31 @@ def _get_histogram_values(series, bins=10):
     return results
 
 
-def _infer_datetime_frequencies(dataframe, datetime_columns=None):
-    """
-    note that for frequency to work there needs to be more than 3 elements
-    changing the data may mean you need to rerun this
+def _infer_temporal_frequencies(dataframe, temporal_columns=None):
+    """Infers the observation frequency (daily, biweekly, yearly, etc) of each temporal column
+            in the DataFrame. Temporal columns are ones with the logical type Datetime or Timedelta.
+
+    Args:
+        dataframe (DataFrame): The DataFrame for which frequncies should be inferred.
+        temporal_columns (list[str], optional): Columns for which frequencies should be inferred. Must be columns
+            that are present in the DataFrame and are temporal in nature. Defaults to None. If not
+            specified, all temporal columns will have their frequencies inferred.
+
+    Returns:
+        (dict): A dictionary where each key is a temporal column from the DataFrame, and the
+            value is its observation frequency represented as a pandas offset alias string (D, M, Y, etc.)
+            or None if no uniform frequency was present in the data.
     """
     logical_types = dataframe.ww.logical_types
 
-    if datetime_columns is None:
-        datetime_columns = [
+    if temporal_columns is None:
+        temporal_columns = [
             col
             for col, ltype in logical_types.items()
             if isinstance(ltype, (Datetime, Timedelta))
         ]
     else:
-        for col in datetime_columns:
+        for col in temporal_columns:
             if col not in dataframe:
                 raise ValueError(f"Column {col} not found in dataframe.")
             ltype = logical_types[col]
@@ -672,4 +682,4 @@ def _infer_datetime_frequencies(dataframe, datetime_columns=None):
                     f"Cannot determine frequency for column {col} with logical type {ltype}"
                 )
 
-    return {col: pd.infer_freq(dataframe[col]) for col in datetime_columns}
+    return {col: pd.infer_freq(dataframe[col]) for col in temporal_columns}
--- a/woodwork/statistics_utils.py
+++ b/woodwork/statistics_utils.py
@@ -654,15 +654,22 @@ def _infer_datetime_frequencies(dataframe, datetime_columns=None):
     note that for frequency to work there needs to be more than 3 elements
     changing the data may mean you need to rerun this
     """
+    logical_types = dataframe.ww.logical_types
+
     if datetime_columns is None:
-        logical_types = dataframe.ww.logical_types
         datetime_columns = [
             col
             for col, ltype in logical_types.items()
             if isinstance(ltype, (Datetime, Timedelta))
         ]
     else:
-        # confirm all columns are present in the dataframe and are datetime in nature
-        pass
+        for col in datetime_columns:
+            if col not in dataframe:
+                raise ValueError(f"Column {col} not found in dataframe.")
+            ltype = logical_types[col]
+            if not isinstance(ltype, (Datetime, Timedelta)):
+                raise TypeError(
+                    f"Cannot determine frequency for column {col} with logical type {ltype}"
+                )
 
     return {col: pd.infer_freq(dataframe[col]) for col in datetime_columns}
--- a/woodwork/statistics_utils.py
+++ b/woodwork/statistics_utils.py
@@ -5,7 +5,14 @@ import pandas as pd
 from sklearn.metrics.cluster import normalized_mutual_info_score
 
 from woodwork.accessor_utils import _is_dask_dataframe, _is_koalas_dataframe
-from woodwork.logical_types import Datetime, Double, Integer, IntegerNullable, LatLong
+from woodwork.logical_types import (
+    Datetime,
+    Double,
+    Integer,
+    IntegerNullable,
+    LatLong,
+    Timedelta,
+)
 from woodwork.utils import _update_progress, get_valid_mi_types, import_or_none
 
 dd = import_or_none("dask.dataframe")
@@ -642,12 +649,20 @@ def _get_histogram_values(series, bins=10):
     return results
 
 
-def _infer_datetime_frequencies(dataframe, columns=None):
+def _infer_datetime_frequencies(dataframe, datetime_columns=None):
     """
     note that for frequency to work there needs to be more than 3 elements
+    changing the data may mean you need to rerun this
     """
-    # if columns is not none confirm all columns are present and datetimes
-    # otherwise just select all datetime columns
+    if datetime_columns is None:
+        logical_types = dataframe.ww.logical_types
+        datetime_columns = [
+            col
+            for col, ltype in logical_types.items()
+            if isinstance(ltype, (Datetime, Timedelta))
+        ]
+    else:
+        # confirm all columns are present in the dataframe and are datetime in nature
+        pass
 
-    # loop over datetime columns and get frequencies, returning them as strings or None
-    return {}
+    return {col: pd.infer_freq(dataframe[col]) for col in datetime_columns}
--- a/woodwork/statistics_utils.py
+++ b/woodwork/statistics_utils.py
@@ -640,3 +640,14 @@ def _get_histogram_values(series, bins=10):
         )
 
     return results
+
+
+def _infer_datetime_frequencies(dataframe, columns=None):
+    """
+    note that for frequency to work there needs to be more than 3 elements
+    """
+    # if columns is not none confirm all columns are present and datetimes
+    # otherwise just select all datetime columns
+
+    # loop over datetime columns and get frequencies, returning them as strings or None
+    return {}
--- a/woodwork/table_accessor.py
+++ b/woodwork/table_accessor.py
@@ -27,7 +27,7 @@ from woodwork.statistics_utils import (
     _get_mutual_information_dict,
     _get_valid_mi_columns,
     _get_value_counts,
-    _infer_datetime_frequencies,
+    _infer_temporal_frequencies,
 )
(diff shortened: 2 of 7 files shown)


woodwork/table_accessor.py
```diff
@@ -1081,6 +1082,25 @@ def value_counts(self, ascending=False, top_n=10, dropna=False):
         """
         return _get_value_counts(self._dataframe, ascending, top_n, dropna)
 
+    @_check_table_schema
+    def infer_temporal_frequencies(self, temporal_columns=None):
```
tamargrey, line 1086: Went with "temporal" across this feature instead of "datetime" because it will work for timedelta or datetime. If that's confusing, I'd be willing to switch to just `infer_frequencies` or something else that makes it clear that this is done on the table-level with all qualifying columns.

thehomebrewnerd (commented):
This looks pretty good to me. Do you think it would be worth testing with a broader range of offset alias values? Just curious if this also works with things like nanosecods or more "atypical" offsets like business day or month start: https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#offset-aliases

Certainly don't need to go overboard with our testing though, but might be good to have some broader representation of offset types in our test dataframe.

tamargrey:
> This looks pretty good to me. Do you think it would be worth testing with a broader range of offset alias values? Just curious if this also works with things like nanosecods or more "atypical" offsets like business day or month start: https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#offset-aliases

I'm happy to include a more broad set of offset aliases, but I wanted to avoid testing the actual values returned by `infer_freq`, since that seems mostly like testing the pandas function itself. So I'm not totally sure what we'd go about testing.

thehomebrewnerd:
> > Do you think it would be worth testing with a broader range of offset alias values? Just curious if this also works with things like nanosecods or more "atypical" offsets like business day or month start: https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#offset-aliases
> 
> I'm happy to include a more broad set of offset aliases, but I wanted to avoid testing the actual values returned by `infer_freq`, since that seems mostly like testing the pandas function itself. So I'm not totally sure what we'd go about testing.

Yeah, I get what you are saying. I guess I'm wondering a little bit about the limitations of the pandas function, since those will also be our limitations. It's not clear to me what those limitations are from the pandas function documentation. Maybe it's not worth worrying about though.

freddyaboulton (commented):
@tamargrey I think this looks good to me! Regarding your questions:

> Should we sort each column before passing it to `infer_freq`?

I would say no until we figure out how often/useful it is to infer frequencies on multiple datetime columns. My hunch is that most users will only care about the `time_index` which will already be sorted so let's not double sort for now. Of course, users can have a non-sorted time_index but I don't think there's anything we can do about that.

> Do we limit the number of rows passed to infer_freq?

I am on the fence about this. I think we would need to do some experimentation to see what results we get with taking a sample vs not. My guess is that it's ok to do so. If we can't infer a frequency on the first k rows, we're not going to be able to infer a frequency on the whole column. In my opinion the right thing to do is to take the first-order diff of all rows and find the frequency that works the best for the most rows (like what's mentioned in https://github.com/alteryx/woodwork/issues/1165 for datetime format inference) but that seems like a v2 improvement? 

Regarding @thehomebrewnerd 's comment:

>  I guess I'm wondering a little bit about the limitations of the pandas function

I think it might be useful to document what those limitations are. I don't know what all of those are so maybe we don't have to do this prior to merge but I think it will be useful to document the cases in which this will fails. For example, if the datetime column has any NaN (because ww can't infer a datetime format, for example) `infer_freq` will also fail. Or if there is a single row not matching the frequency of the other rows.

tamargrey:
@freddyaboulton thanks for the input! Sounds like we have a decent number of things to add into a V2 issue, though from your and @thehomebrewnerd's comments, I'll put some time right now into looking a bit further into `infer_freq`'s limitations so we can document them.

woodwork/table_accessor.py
```diff
@@ -1081,6 +1082,47 @@ def value_counts(self, ascending=False, top_n=10, dropna=False):
...
+            temporal_columns (list[str], optional): Columns for which frequencies should be inferred. Must be columns
+                that are present in the DataFrame and are temporal in nature. Defaults to None. If not
+                specified, all temporal columns will have their frequencies inferred.
+
+        Returns:
+            (dict): A dictionary where each key is a temporal column from the DataFrame, and the
+                value is its observation frequency represented as a pandas offset alias string (D, M, Y, etc.)
+                or None if no uniform frequency was present in the data.
+
+        Note:
```
tamargrey, line 1100: @freddyaboulton @thehomebrewnerd Tried to cover all of the behaviors that I could find for `infer_freq` in the bullet points below. Let me know if there's anything you thing I should add or if anything is confusingly written!

woodwork/table_accessor.py
```diff
@@ -1096,6 +1096,28 @@ def infer_temporal_frequencies(self, temporal_columns=None):
             (dict): A dictionary where each key is a temporal column from the DataFrame, and the
                 value is its observation frequency represented as a pandas offset alias string (D, M, Y, etc.)
                 or None if no uniform frequency was present in the data.
+
+        Note:
+            The pandas util ``pd.infer_freq``, which is used in this method, has the following behaviors:
+                - If even one row in a column does not follow the frequency seen in the remaining columns,
```
thehomebrewnerd, line 1102: I think this should be `...does not follow the frequency seen in the remaining rows,`

woodwork/table_accessor.py
```diff
@@ -1096,6 +1096,28 @@ def infer_temporal_frequencies(self, temporal_columns=None):
...
+                    no frequency will be able to be inferred. Example of otherwise daily data that skips one day:
```
thehomebrewnerd, line 1103: Maybe just say `no frequency will be inferred.` here and remove the `able to be` phrase?

woodwork/table_accessor.py
```diff
@@ -1096,6 +1096,28 @@ def infer_temporal_frequencies(self, temporal_columns=None):
...
+                    need to combine aliases.
+                - Offset strings can contain more information than just the offset alias. For example, a date range
+                    ``pd.date_range(start="2020-01-01", freq="w", periods=10)`` will be inferred to have frequency ``W-SUN``.
+                    That string is an offset alias with an anchoring suffix that indicates that the data is not only
+                    observed at a weekly frequency, but that all the dates are on Sundays. More anchored offsets
+                    can be seen here: https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#anchored-offsets
+                - Some frequencies that can be defined for a ``pd.date_range`` cannot then be re-inferred by pandas' ``pd.infer_freq``.
+                    One example of this can be seen when using the business day offset alias ``B``
+                    ``pd.date_range(start="2020-01-01", freq="4b", periods=10)``, which is a valid ``freq``
+                    parameter when building the date range, but is not then re-inferrable.
```
thehomebrewnerd, line 1120: I think we could just say `inferrable` here instead of `re-inferrable`?

thehomebrewnerd (commented):
The additional detail in the note looks good - just a couple small wording suggestions to consider.

woodwork/table_accessor.py
freddyaboulton: This is awesome! Thanks for adding this.

woodwork/table_accessor.py
tamargrey: good catch!

thehomebrewnerd (approved):
LGTM!

thehomebrewnerd (approved):
LGTM!

woodwork/statistics_utils.py
```diff
@@ -640,3 +647,39 @@ def _get_histogram_values(series, bins=10):
...
+            col
+            for col, ltype in logical_types.items()
+            if isinstance(ltype, (Datetime, Timedelta))
+        ]
+    else:
+        for col in temporal_columns:
+            if col not in dataframe:
+                raise ValueError(f"Column {col} not found in dataframe.")
+            ltype = logical_types[col]
+            if not isinstance(ltype, (Datetime, Timedelta)):
```
ParthivNaresh, line 680: I'm curious why we're not including PeriodIndex here?

`period_range = pd.period_range(start='1/1/21', freq="3D", periods=4)`
`print(period_range)`
`print(period_range.freq)`
`>>> PeriodIndex(['2021-01-01', '2021-01-04', '2021-01-07', '2021-01-10'], dtype='period[3D]')`
`>>> <3 * Days>`

woodwork/table_accessor.py
```diff
@@ -1081,6 +1082,48 @@ def value_counts(self, ascending=False, top_n=10, dropna=False):
...
+                - Offset aliases can be combined to create something like ``2d1H``, which could also be expressed as '49H'.
+                    Pandas' frequency inference will return the lower common alias, ``49H``, in situations when it'd otherwise
+                    need to combine aliases.
+                - Offset strings can contain more information than just the offset alias. For example, a date range
+                    ``pd.date_range(start="2020-01-01", freq="w", periods=10)`` will be inferred to have frequency ``W-SUN``.
+                    That string is an offset alias with an anchoring suffix that indicates that the data is not only
+                    observed at a weekly frequency, but that all the dates are on Sundays. More anchored offsets
+                    can be seen here: https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#anchored-offsets
+                - Some frequencies that can be defined for a ``pd.date_range`` cannot then be re-inferred by pandas' ``pd.infer_freq``.
+                    One example of this can be seen when using the business day offset alias ``B``
```
ParthivNaresh, line 1119: Great addition!

ParthivNaresh (approved):
Looks great! Just curious if we want to include PeriodIndex here as well.

woodwork/statistics_utils.py
tamargrey: These are Woodwork logical types! Any column that's in a woodwork dataframe will be a pandas Series, and the only pandas dtypes Woodwork allows that are valid inputs to `infer_freq` are the ones attached to these logical types

ParthivNaresh: Ahh completely misread those, awesome looks perfect!

PR #1206, opened 2021-11-29
Handle ints represented as doubles in describe_dict with extra stats

### LINKED ISSUE #1203: Update describe_dict extra stats to return top values for Double columns
- As a user, I wish I could use Woodwork to compute `top_values` for `Double` columns when using `df.ww.describe_dict(extra_stats=True). Currently top values are only returned for `Integer` and `IntegerNullable`.

thehomebrewnerd:
- Handle ints represented as doubles in describe extra stats
- Closes #1203 

Updates `df.ww.describe_dict` to compute `top_values` for integers stored as doubles. Top values will be computed for double columns if the range of values is less than or equal to the number of histogram bins and all of the numbers are actually integers, just stored as floats.

### DIFF
--- a/woodwork/statistics_utils.py
+++ b/woodwork/statistics_utils.py
@@ -2,10 +2,11 @@ from timeit import default_timer as timer
 
 import numpy as np
 import pandas as pd
+from pandas.core.dtypes.common import is_integer_dtype
 from sklearn.metrics.cluster import normalized_mutual_info_score
 
 from woodwork.accessor_utils import _is_dask_dataframe, _is_koalas_dataframe
-from woodwork.logical_types import Datetime, Double, Integer, IntegerNullable, LatLong
+from woodwork.logical_types import Datetime, Double, LatLong
 from woodwork.utils import _update_progress, get_valid_mi_types, import_or_none
 
 dd = import_or_none("dask.dataframe")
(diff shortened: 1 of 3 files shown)


woodwork/statistics_utils.py
```diff
@@ -147,14 +148,13 @@ def _get_describe_dict(
...
-                    _range = range(int(values["min"]), int(values["max"]) + 1)
-                    # Calculate top numeric values if range of values present
-                    # is less than or equal number of histogram bins
-                    if len(_range) <= bins:
-                        values["top_values"] = _get_numeric_value_counts_in_range(
-                            series, _range
-                        )
+                _range = range(int(values["min"]), int(values["max"]) + 1)
+                # Calculate top numeric values if range of values present
+                # is less than or equal number of histogram bins
```
tamargrey, line 153: This can probably be updated to explain what the `series % 1 == 0` check is doing.

thehomebrewnerd: Updated!

tamargrey (approved):
one small comment about testing with IntegerNullable that's not a blocker, but otherwise LGTM!

thehomebrewnerd:
> one small comment about testing with IntegerNullable that's not a blocker, but otherwise LGTM!

@tamargrey Did your comment get saved? Not seeing it now...

woodwork/tests/accessor/test_statistics.py
```diff
@@ -763,6 +763,20 @@ def test_describe_with_include(sample_df):
     multi_params_df["full_name"].equals(sample_df.ww.describe()["full_name"])
 
 
+def test_describe_numeric_all_nans():
+    df = pd.DataFrame({"float": [np.nan] * 5})
+    df.ww.init(logical_types={"float": "double"})
```
tamargrey, line 768: @thehomebrewnerd sorry about that. Was wondering if it would make sense to also test with IntegerNullable here

thehomebrewnerd: Should be simple to add. I will update!

tamargrey (approved):
lgtm!

PR #1210, opened 2021-12-02
Add Infer Natural Language to Woodwork

tuethan1999:
- Fixes https://github.com/alteryx/woodwork/issues/1211


docs/source/release_notes.rst
```diff
@@ -5,12 +5,14 @@ Release Notes
 .. Future Release
```
gsheni, line 5:
```suggestion
Future Release
```

docs/source/release_notes.rst
```diff
@@ -5,12 +5,14 @@ Release Notes
...
   ==============
     * Enhancements
+        * Add type inference for natural language (:pr:`1210`) 
     * Fixes
     * Changes
     * Documentation Changes
     * Testing Changes
 
 .. Thanks to the following people for contributing to this release:
```
gsheni, line 14:
```suggestion
   Thanks to the following people for contributing to this release:
```

docs/source/release_notes.rst
```diff
@@ -5,12 +5,14 @@ Release Notes
 .. Future Release
   ==============
```
gsheni, line 6:
```suggestion
==============
```

woodwork/type_sys/inference_functions.py
```diff
@@ -82,6 +93,23 @@ def timedelta_func(series):
     return False
 
 
+def num_common_words(wordlist: Union[Tokens, Any]) -> float:
+    if not isinstance(wordlist, Iterable):
+        return np.nan
+    num_common_words = 0
+    for x in wordlist:
```
thehomebrewnerd, line 100: I don't know if a list comprehension would be faster than a loop here, but I think we could get the number of common words like this:
```
num_common_words = len([x for x in wordlist if x.lower() in COMMON_WORDS_SET])
```

thehomebrewnerd (commented):
Not looking too bad. Just have a few suggestion after an initial read-through.

woodwork/type_sys/inference_functions.py
tuethan1999: I wrote it this way for performance but I don't feel strongly one way or another. Guess the question is if we want to save 3 lines for 14% runtime

```python
from woodwork.type_sys.inference_functions import COMMON_WORDS_SET, NL_delimiters
from woodwork.demo import load_retail

df = load_retail()
series = df['description']
tokens = series.astype("string").str.split(NL_delimiters)

def loop(wordlist):
    num_common_words = 0
    for x in wordlist:
        if x.lower() in COMMON_WORDS_SET:
            num_common_words += 1
    return num_common_words

def list_comprehension(wordlist):
    num_common_words = len([x for x in wordlist if x.lower() in COMMON_WORDS_SET])
    return num_common_words

%timeit -n 100 tokens.map(loop)
%timeit -n 100 tokens.map(list_comprehension)
```
327 ms ± 2.13 ms per loop (mean ± std. dev. of 7 runs, 100 loops each)
373 ms ± 1.36 ms per loop (mean ± std. dev. of 7 runs, 100 loops each)

woodwork/tests/type_system/conftest.py
```diff
@@ -235,6 +235,41 @@ def timedeltas(request):
     return request.getfixturevalue(request.param)
 
 
+# Natural Language Fixtures
+@pytest.fixture
+def pandas_natural_language():
```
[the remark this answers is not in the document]
tuethan1999: I'm not really sure what the benefit would be other than to show people that we also consider long strings to be natural language

woodwork/type_sys/inference_functions.py
thehomebrewnerd: Go with the option that has the best performance. I wasn't trying to reduce lines of code with my suggestion, but I was trying to find the option with the best performance and wasn't sure which it would be. I think generally for our inference functions, performance should always trump a reduction in lines of code.

woodwork/tests/type_system/conftest.py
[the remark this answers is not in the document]
thehomebrewnerd: I guess I was thinking maybe the benefit is that we make sure our inference works as expected with a variety of natural language inputs, and if we ever make a change to the inference function we have a test that helps us make sure we didn't break something unexpectedly for certain types of inputs. Might not be all that important now though.

woodwork/type_sys/inference_functions.py
```diff
@@ -82,6 +93,25 @@ def timedelta_func(series):
...
+            num_common_words += 1
+    return num_common_words
+
+
+def natural_language_func(series):
+    tokens = series.astype("string").str.split(NL_delimiters)
+    mean_num_common_words = np.nanmean(tokens.map(num_common_words))
+
+    return (
+        mean_num_common_words > 1.14
```
gsheni, line 111: Does this exactly `1.1377166509628296`?

woodwork/tests/type_system/conftest.py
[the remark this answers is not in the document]
gsheni: Resolving this comment for now.

woodwork/type_sys/inference_functions.py
gsheni: Resolving this comment

woodwork/type_sys/inference_functions.py
gsheni: Or does it not matter?

thehomebrewnerd (approved):

PR #1218, opened 2021-12-21
Make public method to get subset schema

jeff-hernandez:


woodwork/table_schema.py
```diff
@@ -552,7 +552,12 @@ def _filter_cols(self, include=None, exclude=None, col_names=False):
 
         return cols_to_return
 
-    def _get_subset_schema(self, subset_cols):
+    @property
+    def _get_subset_schema(self):
+        # removing this property might cause breaking changes in EvalML
+        return self.get_subset_schema
```
jeff-hernandez, line 558: Do we want to file an issue to track when this might be safe to remove?

davesque (commented):
I assume we grepped for occurrences of references to this method in woodwork.  Have we done the same for featuretools and other projects in our org?  My intuition is that a grep for `_get_subset_schema` would be pretty specific and would generally only yield matches that are relevant, even for projects like featuretools, tempo client, etc.  And it's probably worth grepping across projects like that if we haven't already.  If we don't find any other occurrences, seems pretty safe to remove.  Unless I'm forgetting something.

*Update:* What I was implying is that, if we'd been thorough about making sure none of our internal projects don't refer to this method, I don't think we should be obligated to maintain backwards compatibility by providing the `_get_subset_schema` property method.  After all, the underscore prefix should have implied that the method wasn't public API and was subject to change without notice.

jeff-hernandez:
@davesque you raise a good point about the private method being subject to change. It does look like EvalML uses this method in a few places.
Screen Shot 2021-12-21 at 1 21 59 PM

davesque:
Cool, yeah I guess then actually maybe the wrapper method makes sense and also the tracking issue for removing it.  Then EvalML will still work at least until we can correct those usages of the method.  Once that is done, we can get rid of the wrapper and close the issue.  Unless there are any other occurrences of the method anywhere.

woodwork/table_schema.py
```diff
@@ -552,7 +552,12 @@ def _filter_cols(self, include=None, exclude=None, col_names=False):
 
         return cols_to_return
 
-    def _get_subset_schema(self, subset_cols):
+    @property
```
thehomebrewnerd, line 555: I think if you change this to a property like this, EvalML will break since they will lose the ability to pass in `subset_cols`, no?

jeff-hernandez: The property serves as an alias that returns the actual function, so it wouldn't cause any breaking changes.

```python
class A:
    def function(self, x):
        return x
    
    @property
    def aliase(self):
        return self.function

assert A().function(True) is True
```

thehomebrewnerd: Yeah, when I first looked, I thought you were calling the other method instead of returning it. Ignore my original comment.

woodwork/table_schema.py
thehomebrewnerd: I think creating an issue makes sense - as far as I know once EvalML upgrades to this version and switches to using the public method, we can remove. Might actually want to create two issues - one in this repo to remove the private method and one in EvalML to upgrade WW and switch to using the public method.

thehomebrewnerd (approved):
Looks fine to me. There is one more location where we use the private method - in `test_get_subset_schema`. Did you leave that for test coverage?

jeff-hernandez:
@thehomebrewnerd yep just for coverage

PR #1234, opened 2022-01-03
Improve natural language inference performance

thehomebrewnerd:
- Improve natural language inference performance

Implements changes to only call natural language inference function if no other type matches are found first, eliminating costly function calls.


woodwork/type_sys/type_system.py
```diff
@@ -315,11 +315,22 @@ def get_inference_matches(types_to_check, series, type_matches=[]):
                 get_inference_matches(check_next, series, type_matches)
             return type_matches
 
-        type_matches = get_inference_matches(self.root_types, series)
+        # Don't include NaturalLanguage as we only want to check that if
+        # no other matches are found
+        types_to_check = [
+            ltype for ltype in self.root_types if ltype != NaturalLanguage
```
gsheni, line 321: Can we change the order of `self.root_types` to have NaturalLanguage be the last value?

woodwork/tests/type_system/test_ltype_inference.py
```diff
@@ -125,6 +133,32 @@ def test_natural_language_inference(natural_language):
             assert isinstance(inferred_type, NaturalLanguage)
 
 
+@patch("woodwork.type_sys.inference_functions.natural_language_func")
+def test_nl_inference_called_on_no_other_matches(nl_mock, pandas_natural_language):
+    new_type_sys = TypeSystem(
```
thehomebrewnerd, line 138: Ended up creating and implementing the mocked function in this way, otherwise the non-mocked version was getting used. Maybe there is a better way?

thehomebrewnerd: Basically the type system object gets created with the non-mocked version of the function. Didn't want to change the global `type_system` object in this test, so created a new type system object in the test and updated it's natural language inference function to use the mocked version of the function instead.

tamargrey: This might be related to something I learned the other day when writing [this test](https://github.com/alteryx/featuretools/pull/1809/files#diff-3b68209906d2ec21024d95fad1426e6c39e3e67370ba26686f178cfbb0c00c03R288). I'm not 100% sure I'm understanding this, though, so could be totally off base here. 

When using a mock on a function that's not being called in the file in which it's defined, the path to reference in `@patch` should be the path to the place where your mocked function is being used, not the place where that function was defined. So in the test linked to above, it's `featuretools.primitives.rolling_transform_primitive._apply_roll_with_offset_gap` and not `featuretools.primitives.utils._apply_roll_with_offset_gap`. In this case, it might be something like `woodwork.type_sys.type_system.TypeSystem.infer_logical_type.natural_language_func`

thehomebrewnerd:
Did a simple test to confirm the performance change.

**This Branch**
image


**Current Release - Woodwork 0.11.0**
image

woodwork/tests/type_system/test_ltype_inference.py
thehomebrewnerd: @tamargrey Interesting. Let me take a look at that closer.

tamargrey:
> Did a simple test to confirm the performance change.

@thehomebrewnerd Nice! Would it be difficult to also see the performance for woodwork 0.10.0 or the commit prior to the one that added nl inference (2e54e42db9d5d8d2a593b2df43eadec61b05db81)? Just to get an idea of what kind of performance hit these changes now are to before we had nl inference?

thehomebrewnerd:
> > Did a simple test to confirm the performance change.
> 
> @thehomebrewnerd Nice! Would it be difficult to also see the performance for woodwork 0.10.0 or the commit prior to the one that added nl inference ([2e54e42](https://github.com/alteryx/woodwork/commit/2e54e42db9d5d8d2a593b2df43eadec61b05db81))? Just to get an idea of what kind of performance hit these changes now are to before we had nl inference?

@tamargrey Here are the results for WW 0.10.0:
image

tamargrey:
> I'm not entirely convinced these changes are an improvement over 0.10.0, but the difference might be due to normal variation (and other stuff I have running on my machine?).

Yep, that seems like a fair assumption to make. But at least its the same order of magnitude, so I'd be willing to say that this new implementation is probably similar enough to 0.10.0 in runtime that we wouldn't have noticed it in the same way!

thehomebrewnerd:
> > I'm not entirely convinced these changes are an improvement over 0.10.0, but the difference might be due to normal variation (and other stuff I have running on my machine?).
> 
> Yep, that seems like a fair assumption to make. But at least its the same order of magnitude, so I'd be willing to say that this new implementation is probably similar enough to 0.10.0 in runtime that we wouldn't have noticed it in the same way!

Agreed!

woodwork/tests/type_system/test_ltype_inference.py
tamargrey: the stack overflow question that lead me to that understanding, in case it's useful: https://stackoverflow.com/questions/32461465/python-patch-not-working

woodwork/type_sys/type_system.py
```diff
@@ -315,11 +315,22 @@ def get_inference_matches(types_to_check, series, type_matches=[]):
...
+        ]
+        type_matches = get_inference_matches(types_to_check, series)
 
         if len(type_matches) == 0:
-            # If no matches, set type to default type (Unknown)
-            logical_type = self.default_type
+            # Check if this is NaturalLanguage, otherwise set
```
tuethan1999, line 326: should we leave a comment explaining that we do this because the inference function was created on these assumptions and for performance reasons?

tuethan1999 (approved):

woodwork/tests/type_system/test_ltype_inference.py
```diff
@@ -125,6 +133,32 @@ def test_natural_language_inference(natural_language):
...
+        default_type=DEFAULT_TYPE,
+    )
+    new_type_sys.inference_functions[NaturalLanguage] = nl_mock
+    inferred_type = new_type_sys.infer_logical_type(pandas_natural_language[0])
+    assert isinstance(inferred_type, NaturalLanguage)
+    assert nl_mock.called
+
+
+@patch("woodwork.type_sys.inference_functions.natural_language_func")
+def test_nl_inference_not_called_with_other_matches(nl_mock, pandas_integers):
```
tamargrey, line 150: a suggestion for a third version of this test would be to confirm that nl inference function had been called on a series whose logical type is inferred as Unknown

woodwork/type_sys/type_system.py
thehomebrewnerd: Updated the comment to better reflect this.

woodwork/tests/type_system/test_ltype_inference.py
thehomebrewnerd: Added this test. Also had to rework the test a little bit to remove some assertions since the mocked function doesn't return the right value. Just was a happy accident that things worked with the tests as they were before.

tamargrey (approved):
lgtm after lint fix and tests pass!

PR #1239, opened 2022-01-05
Prevent setting index that contains null values

### LINKED ISSUE #1191: Prevent NaN values in when setting index
- Going forward, we will not allow Woodwork users to set an a column as the index if there is 1 or more NaN values

jeff-hernandez:
- Closes #1191


woodwork/table_accessor.py
```diff
@@ -1167,14 +1167,14 @@ def _check_index(dataframe, index):
...
-        # Does not check for Dask as Dask does not support is_unique
-        raise IndexError("Index column must be unique")
+    if index is not None and isinstance(dataframe, pd.DataFrame):
+        if not dataframe[index].is_unique:
+            # User specifies an index that is in the dataframe but not unique
+            # Does not check for Dask as Dask does not support is_unique
+            raise IndexError("Index column must be unique")
+
+        if dataframe[index].notnull().any():
+            raise IndexError("Index contains null values")
```
jeff-hernandez, line 1177: @alteryx/ml-tools we should confirm whether or not this validation should happen here since it can possibly introduce multiple breaking changes in the unit tests

thehomebrewnerd: Based on a quick glance, this feels like the right place for this check to happen to me.

woodwork/table_accessor.py
```diff
@@ -1167,14 +1167,14 @@ def _check_index(dataframe, index):
...
-        # User specifies an index that is in the dataframe but not unique
-        # Does not check for Dask as Dask does not support is_unique
-        raise IndexError("Index column must be unique")
+    if index is not None and isinstance(dataframe, pd.DataFrame):
+        if not dataframe[index].is_unique:
+            # User specifies an index that is in the dataframe but not unique
+            # Does not check for Dask as Dask does not support is_unique
+            raise IndexError("Index column must be unique")
+
+        if dataframe[index].notnull().any():
```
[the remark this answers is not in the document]
jeff-hernandez: Yeah, either `.isnull().any()` or `.notnull().all()` should work but haven't compared performance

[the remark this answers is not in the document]
jeff-hernandez: nvm, disregard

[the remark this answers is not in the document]
thehomebrewnerd: Seems like `.isnull().any()` could be faster by a bit because the `.any()` check could stop as soon as it finds a single null value, while the `.all()` check would have to go through all the rows every time to make sure everything is not null. Probably worth confirming though.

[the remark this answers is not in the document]
jeff-hernandez: Yeah I think `.isnull().any()` makes more sense

woodwork/tests/utils/test_concat.py
```diff
@@ -386,28 +386,6 @@ def test_concat_cols_validate_schema(mock_validate_accessor_params, sample_df):
     assert mock_validate_accessor_params.called
 
 
-def test_concat_cols_mismatched_index_adds_single_nan(sample_df):
-    sample_df.ww.init(index="id")
```
tamargrey, line 390: If I'm understanding correctly, this test would now not pass because introducing a nan into the resulting dataframe that has an index is not allowed. 

Can we remove this `index='id'` and keep the test? I think doing that shouldn't change the usefulness of this test, which is mostly about the behavior when `concat_columns` introduces nans in non nullable logical types

jeff-hernandez: Yep sounds good

woodwork/table_accessor.py
```diff
@@ -1167,14 +1167,14 @@ def _check_index(dataframe, index):
...
-        and isinstance(dataframe, pd.DataFrame)
-        and not dataframe[index].is_unique
-    ):
-        # User specifies an index that is in the dataframe but not unique
-        # Does not check for Dask as Dask does not support is_unique
-        raise IndexError("Index column must be unique")
+    if index is not None and isinstance(dataframe, pd.DataFrame):
+        if not dataframe[index].is_unique:
+            # User specifies an index that is in the dataframe but not unique
+            # Does not check for Dask as Dask does not support is_unique
```
thehomebrewnerd, line 1173: Minor detail, but I think we should update this comment a bit. We are now skipping both the unique check and the null check for Dask and Koalas, with the primary reason being that we don't want to pull the data into memory to perform the check on initialization. Maybe move it above the `if not dataframe[index].is_unique:` line and update the text to better match the new behavior?

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -2961,3 +2964,8 @@ def test_falsy_columns_in_partial_schema(falsy_names_df):
     assert new_df.ww.index == 0
     assert new_df.ww.time_index == ""
     assert new_df.ww.name == 0
+
+
+def test_nan_index_error(sample_df_pandas):
+    with pytest.raises(IndexError, match="Index contains null values"):
+        sample_df_pandas.ww.init(index="email")
```
thehomebrewnerd, line 2971: What do you think about including this in `test_check_index_errors`? I think that would be a more targeted test to confirm we get the right behavior from the `_check_index` function where this error actually gets raised, without needing to call the entire init process.

We actually should probably also have a test that confirms that `_check_index` function gets called on both `.init` and `.set_index` calls, if we don't already. That's a bit of a separate issue though.

woodwork/tests/utils/test_accessor_utils.py
```diff
@@ -255,7 +255,7 @@ def test_get_invalid_schema_message_index_checks(sample_df):
 
     nan_df = pd.DataFrame(
         {
-            "id": pd.Series([5, 4, None, 2], dtype="float64"),
+            "id": pd.Series([5, 4, 3, 2], dtype="float64"),
```
thehomebrewnerd, line 258: I'm not entirely clear now on what we were testing here, but I think perhaps this test was intended to confirm that if a single null value is present that we don't generate an invalid schema message? @tamargrey Do you recall?

If we were testing that introducing a single null value was ok before, I think maybe with this new behavior of not allowing any nulls in the index, we may need to update `get_invalid_schema_message` to return a message if some operation causes null values to be introduced in the index that invalidates a schema (like was happening in `test_concat_cols_mismatched_index_adds_single_nan` before the change you made to not set the index to `id`). This would be similar to the assertion that precedes this.

tamargrey: I don't fully remember, but given the fact that the dataframe is called `nan_df` and that the other two index-related schema vilidity checks were already made in this test, I'd guess that the goals was to confirm that an index column containing nans was acceptable to `get_invalid_schema_message`, in which case the check with this dataframe isn't necessary

jeff-hernandez: It sounds like this case is safe to remove then, right?

thehomebrewnerd: I guess it depends on if we want to handle the situation where a schema is invalid because an index contains a single null value in `get_invalid_schema_message`. The `.is_unique` check will cover a scenario where we have multiple null values, I think, but we wouldn't have anything in that function to catch a single missing value. I'm not sure how important that really is though.

tamargrey: If the following works, then we need to add a case to `get_invalid_schema_message` to catch it and raise an error and update this test to confirm that that message is produced 

```
df = pd.DataFrame(
      {
          "id": pd.Series([5, 4, 3, 2], dtype="float64"),
          "col": pd.Series(["b", "b", "b", "d"], dtype="category"),
      }
  )
  df.ww.init(index="id")
  df_schema = df.ww.schema

nan_df = df.replace({3:None})
nan_df.ww.init_with_full_schema(df_schema)

```

thehomebrewnerd: @tamargrey It works, but we first have to do a couple other things - we need to make sure the dtypes are consistent and set the pandas index:
```
df = pd.DataFrame(
      {
          "id": pd.Series([5, 4, 3, 2], dtype="float64"),
          "col": pd.Series(["b", "b", "b", "d"], dtype="category"),
      }
  )
df.ww.init(index="id")
df_schema = df.ww.schema

nan_df = df.replace({3:None})
nan_df["id"] = nan_df["id"].astype("float64")
nan_df = nan_df.set_index("id", drop=False)
nan_df.ww.init_with_full_schema(df_schema)
```
Without those other two steps we get a dtype mismatch error and a index mismatch error, but once those changes are made, the init works fine without issue.

tamargrey: oops, yeah forgot about the other requirements. In that case, we shoud add a check like `if dataframe[schema.index].isnull().any(): return "Index contains null values"`

jeff-hernandez: I was looking through the validation process and had thoughts/questions. If I'm understanding correctly, `init_with_full_schema` assigns or replaces the underlying schema whereas `init_with_partial_schema` creates and builds the schema. As a result, these two places (possibly more?) where the same validation (more or less) needs to happen.

jeff-hernandez: Would it be possible to take schema in `init_with_full_schema` and convert it back to parameters needed for `init_with_partial_schema`?

tamargrey: It'd definitely be possible, and when you pass a full schema into `df.ww.init` that's now what happens. But when you pass parameters into `init_with_partial_schema`, it'll change the dataframe to fit those parameters if it has to.

We specifically kept `init_with_full_schema` with its current behavior that won't make any changes to the dataframe, since it's useful in cases where we expect the dataframe to be valid and wouldn't want the partial schema init to, say, change dtypes or reset the underlying index. We use it a couple of places around woodwork and featuretools and rely on that behavior. 

I definitely agree that having this logic twice over doesn't feel the best, though I don't think we have anywhere where the validation is actually being performed twice in a row. It'd be great if we could have shared utils for this, but the reason we dont, if I'm rmembering correctly, is that we want `get_invalid_schema_message` to return the explanation without yet raising an error.

jeff-hernandez: If `init_with_partial_schema` only changes the dataframe when necessary, it sounds like this method shouldn't change the dataframe at all in some cases where it's not necessary. I'm not entirely sure that I understand the need for both, but not against leaving this as is.

tamargrey: This is definitely a valid line of questioning. I think there's definitely good reason to have `init_with_partial_schema` when setting `validate=False`--we use that a lot across woodwork when we know the schema will be valid. There's not much usage with `validate=True`, but the utilities that make up the validation are independently important. `get_invalid_schema_message` and `is_schema_valid` are both used by evalml.

So we still need to be able to define what is and is not a valid schema, and with the changes in this PR, we should add a check that a dataframe passed into `get_invalid_schema_message` or `is_schema_valid` does not have any nans in its index column.

woodwork/table_accessor.py
jeff-hernandez: updated

woodwork/tests/accessor/test_table_accessor.py
jeff-hernandez: updated `test_check_index_errors` also added tests for `init` and `set_index`

woodwork/tests/utils/test_accessor_utils.py
jeff-hernandez: updated

thehomebrewnerd (approved):
LGTM!

tamargrey (approved):
lgtm!

PR #1243, opened 2022-01-10
[DO NOT MERGE] Add df.ww.to_parquet method to serialize to a single file

### LINKED ISSUE #1223: Add df.ww.to_parquet method to save data and typing info in specified file
As a user, I wish I could use Woodwork to save my Woodwork dataframe as a parquet file with a filename/path that I specify. Currently, Woodwork serialization accepts only a path and not a file name, and stores the data in a subdirectory named `data` and sets the filename based on the name of the dataframe.  I would like to have the ability to specify the path/filename directly and have Woodwork save the data there. 

The filename could be an optional parameter, and Woodwork could name the file based on the datafame name as it does today, if the user does not specify the filename.

Related to #968 

#### Code Example

```python
df.ww.to_parquet(path=my_directory, filename="my_file.parquet")
```

### LINKED ISSUE #1225: Read parquet files created with df.ww.to_parquet
As a user, I wish I could read a Woodwork initialized dataframe from a parquet or arrow file created with the `df.ww.to_parquet` function. The Woodwork typing information should be read in from the metadata in the file header.

Related Issues: #1223 and #1224

thehomebrewnerd:
Add df.ww.to_parquet method to serialize to a single file
- Closes #1223 
- Closes #1225 

Adds a new accessor method to serialize a Woodwork dataframe to the specified parquet file. Saves all Woodwork typing information and other metadata in the parquet file metadata. Also adds a new method `read_parquet` to deserialize the saved information back into a Woodwork initialized dataframe.

### DIFF
--- a/woodwork/deserialize.py
+++ b/woodwork/deserialize.py
@@ -231,15 +231,15 @@ def _typing_info_to_init_params(typing_info):
 
 def read_parquet(path, filename, lib="pandas"):
     """
-    Read data from the parquet file at the location specified by `path`
-    with the filename specified by `filename`. Will initialize Woodwork using
+    Read data from the parquet file at the location specified by ``path``
+    with the filename specified by ``filename``. Will initialize Woodwork using
     the typing information stored in the parquet file metadata. If typing info
     is not present, will initialize Woodwork without any arguments.
 
     Args:
         path (str): Location on disk to read from.
         filename (str): Name of file to read. Will be ignored if reading dataframe with Dask.
-        lib (str): Name of library to use for reading data. Defaults to pandas.
+        lib (str): Name of library to use for reading data. Defaults to ``pandas``.
             Should be one of ["pandas", "dask"]
     """
     import pyarrow as pa
--- a/woodwork/deserialize.py
+++ b/woodwork/deserialize.py
@@ -217,7 +217,7 @@ def _check_schema_version(saved_version_str):
         )
 
 
-def read_parquet_file(filepath):
+def read_parquet(filepath):
     """Read data from the specified parquet file and initialize Woodwork using
     the typing information stored in the parquet file metadata."""
     import pyarrow as pa
--- a/woodwork/deserialize.py
+++ b/woodwork/deserialize.py
@@ -215,3 +215,73 @@ def _check_schema_version(saved_version_str):
             OutdatedSchemaWarning().get_warning_message(saved_version_str),
             OutdatedSchemaWarning,
         )
+
+
+def read_parquet_file(filepath):
+    """Read data from the specified parquet file and initialize Woodwork using
+    the typing information stored in the parquet file metadata."""
+    import pyarrow as pa
+
+    dataframe = pd.read_parquet(filepath)
+    file_metadata = pa.parquet.read_metadata(filepath)
+    table_typing_info = json.loads(file_metadata.metadata[b"woodwork_metadata"])
+
+    loading_info = table_typing_info["loading_info"]
+    table_type = loading_info.get("table_type", "pandas")
+
+    logical_types = {}
+    semantic_tags = {}
+    column_descriptions = {}
+    column_origins = {}
+    column_metadata = {}
+    use_standard_tags = {}
+    column_dtypes = {}
+    for col in table_typing_info["column_typing_info"]:
+        col_name = col["name"]
+
+        ltype_metadata = col["logical_type"]
+        ltype = ww.type_system.str_to_logical_type(
+            ltype_metadata["type"], params=ltype_metadata["parameters"]
+        )
+
+        tags = col["semantic_tags"]
+
+        if "index" in tags:
+            tags.remove("index")
+        elif "time_index" in tags:
+            tags.remove("time_index")
+
+        logical_types[col_name] = ltype
+        semantic_tags[col_name] = tags
+        column_descriptions[col_name] = col["description"]
+        column_origins[col_name] = col["origin"]
+        column_metadata[col_name] = col["metadata"]
+        use_standard_tags[col_name] = col["use_standard_tags"]
+
+        col_type = col["physical_type"]["type"]
+        if col_type == "category":
+            # Make sure categories are recreated properly
+            cat_values = col["physical_type"]["cat_values"]
+            cat_dtype = col["physical_type"]["cat_dtype"]
+            if table_type == "pandas":
+                cat_object = pd.CategoricalDtype(pd.Index(cat_values, dtype=cat_dtype))
+            else:
+                cat_object = pd.CategoricalDtype(pd.Series(cat_values))
+            col_type = cat_object
+        column_dtypes[col_name] = col_type
+
+    dataframe.ww.init(
+        name=table_typing_info.get("name"),
+        index=table_typing_info.get("index"),
+        time_index=table_typing_info.get("time_index"),
+        logical_types=logical_types,
+        semantic_tags=semantic_tags,
+        use_standard_tags=use_standard_tags,
+        table_metadata=table_typing_info.get("table_metadata"),
+        column_metadata=column_metadata,
+        column_descriptions=column_descriptions,
+        column_origins=column_origins,
+        validate=False,
+    )
+
+    return dataframe
--- a/woodwork/deserialize.py
+++ b/woodwork/deserialize.py
@@ -265,7 +265,7 @@ def read_parquet(path, filename, lib="pandas", profile_name=None):
         dataframe = pd.read_parquet(filepath)
         file_metadata = pa.parquet.read_metadata(filepath)
 
-    ww_metadata = file_metadata.metadata.get(str.encode(METADATA_KEY))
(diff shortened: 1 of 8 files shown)


woodwork/deserialize.py
```diff
@@ -215,3 +182,98 @@ def _check_schema_version(saved_version_str):
...
+    """
+    import pyarrow as pa
+
+    if lib == "dask":
+        dataframe = dd.read_parquet(path)
+        # Read the metadata from the first parquet file in the directory
+        meta_filepath = glob.glob(os.path.join(path, "*.parquet"))[0]
+        file_metadata = pa.parquet.read_metadata(meta_filepath)
+    else:
+        if _is_s3(path):
```
tamargrey, line 254: nitpick: could you do `elif _is_s3(path)` for this block and have the `else` be at the same level

woodwork/deserialize.py
```diff
@@ -215,3 +182,98 @@ def _check_schema_version(saved_version_str):
...
+
+                use_smartopen(local_filepath, s3_path, transport_params)
+                dataframe = pd.read_parquet(local_filepath)
+                file_metadata = pa.parquet.read_metadata(local_filepath)
+        else:
+            filepath = os.path.join(path, filename)
+            dataframe = pd.read_parquet(filepath)
+            file_metadata = pa.parquet.read_metadata(filepath)
+
+    ww_metadata = file_metadata.metadata.get(b"woodwork_metadata")
```
tamargrey, line 269: Should the string `"woodwork_metadata"` be a constant somewhere that gets used at `serialize` as well?

woodwork/deserialize.py
```diff
@@ -215,3 +182,98 @@ def _check_schema_version(saved_version_str):
...
+        "column_metadata": column_metadata,
+        "use_standard_tags": use_standard_tags,
+    }
+
+
+def read_parquet(path, filename, lib="pandas", profile_name=None):
+    """
+    Read data from the parquet file at the location specified by ``path``
+    with the filename specified by ``filename``. Will initialize Woodwork using
+    the typing information stored in the parquet file metadata. If typing info
```
tamargrey, line 236: Would it be worth indicating that it looks for. the woodwork typing info under the `"woodwork_metadata"` string? If someone has serialized and deserialized through woodwork, there's no real reason to have to know this, but it could be useful if there's some further manipulation a user wanted to do.  (Though maybe we specifically would want to avoid that, in which case, we shouldn't overly advertise location where the typing info is stored within the metadata

thehomebrewnerd: I'm a bit on the fence here, but generally I am leaning toward not advertising this too much, as I don't think the intent would be for the users to do anything with the Woodwork metadata that gets stored. One reason I could see to list this would be to help advanced users avoid accidentally corrupting the WW data if they modify the parquet file metadata.

I'm thinking it might be best to update the key to include an underscore to better indicate that it isn't supposed to be modified by the users. Changes there, if not done properly, could impact the ability to deserialize the file, and isn't something I would expect casual users to do. 

Maybe we can revisit in the future if we find a situation where users need to know about the details, but for now I think I'd leave this out of the documentation. You ok with this?

woodwork/deserialize.py
thehomebrewnerd: Good idea!

woodwork/deserialize.py
thehomebrewnerd: Yeah, good call. I will clean this up.

woodwork/deserialize.py
tamargrey: I like adding an underscore to the key a lot! And maybe a more genera note that users should be careful to not modify the parquet file metadata, as it may impact deserialization.

tamargrey (approved):
This looks good to me, but probably worth also getting another set of eyes on this before merging.

thehomebrewnerd:
> Why treat this is as a separate serialization method instead of updating the parquet approach of the `to_disk` method?

We *could* do that, but I think that pretty much means a ground-up re-write of `to_disk` and all the other utility functions that are used along the way (both read and write) to handle this type of flow. Currently `to_disk` does not support users specifying the output filename, and I think that is something we want here. Handling parquet files in `to_disk` like they are here would mean that in all the methods and functions we would need unique paths for parquet files because they would work differently than the other formats.

As a bit of a tangent, I'm not sure having a monolithic `to_disk` method is a great approach long term. Because of all the logic branches, it does make it a bit difficult to follow what happens for a given file format, compared to having individual `to_format` methods. I guess that is probably a conversation for a different day though...

woodwork/serialize.py
```diff
@@ -223,3 +224,43 @@ def save_orc_file(dataframe, filepath):
...
+    else:
+        if filename is None:
+            filename = "data.parquet"
+        table = pa.Table.from_pandas(dataframe)
+        table_meta = table.schema.metadata
+        combined_metadata = {
+            METADATA_KEY.encode(): json.dumps(ww_typing).encode(),
+            **table_meta,
+        }
+        table = table.replace_schema_metadata(combined_metadata)
```
davesque, line 249: Something we might want to do here is this:
```suggestion
        table_meta = table.schema.metadata
        table_meta[METADATA_KEY.encode()] = json.dumps(ww_typing).encode()
        table = table.replace_schema_metadata(table_meta)
```
The advantage of doing it this way is that we would always overwrite the `METADATA_KEY` key which I believe is what's intended here.  If that's not the intention, the old approach is also fine I think.

woodwork/deserialize.py
```diff
@@ -215,3 +182,97 @@ def _check_schema_version(saved_version_str):
...
+
+            use_smartopen(local_filepath, s3_path, transport_params)
+            dataframe = pd.read_parquet(local_filepath)
+            file_metadata = pa.parquet.read_metadata(local_filepath)
+    else:
+        filepath = os.path.join(path, filename)
+        dataframe = pd.read_parquet(filepath)
+        file_metadata = pa.parquet.read_metadata(filepath)
+
+    ww_metadata = file_metadata.metadata.get(str.encode(METADATA_KEY))
```
davesque, line 268: Any reason we wouldn't just do this here to keep things consistent?
```suggestion
    ww_metadata = file_metadata.metadata.get(METADATA_KEY.encode())
```
Or maybe another thing we could do is just something like this:
```python
METADATA_KEY = b"_woodwork_metadata"
```
Since we always pass it to arrow which requires keys as bytes.

davesque (commented):
Mostly looks good.  Just had a couple minor thoughts/questions.

woodwork/deserialize.py
```diff
@@ -135,19 +113,8 @@ def _typing_information_to_woodwork_table(table_typing_info, validate, **kwargs)
...
-        logical_types=logical_types,
-        semantic_tags=semantic_tags,
-        use_standard_tags=use_standard_tags,
-        table_metadata=table_typing_info.get("table_metadata"),
-        column_metadata=column_metadata,
-        column_descriptions=column_descriptions,
-        column_origins=column_origins,
-        validate=validate,
-    )
+    init_params = _typing_info_to_init_params(table_typing_info)
```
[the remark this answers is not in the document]
thehomebrewnerd: As I have been going through this update, I have also come to a similar conclusion that our entire serialization/deserialization approach might benefit from some refactoring. Basically, what led me to this conclusion, is that it's really not possible to change the way we serialize to parquet files inside of our `ww.df.to_disk` method, without changing a lot of other helper functions that are also used for serializing to other file types.

To me, it feels like we should be able to change the way we serialize to parquet without potentially impacting/breaking the way we serialize to other formats - possibly by just updating a `ParquetSerializer` class or something like that. The fact that we can't do that right now leads me to believe there might be a better design for our serialization code.

I suggest we give things a bit more thought before we merge in this PR and decide how we want to proceed.

woodwork/serialize.py
thehomebrewnerd: Yeah, that's fair and probably a little cleaner. I mostly copied this code from another example on how to update parquet metadata and didn't think through that detail enough.

woodwork/deserialize.py
thehomebrewnerd: No reason, that was just an oversight. I like the idea of updating `METADATA_KEY` to be bites from the start so we can avoid calling `.encode()` in several places.

woodwork/deserialize.py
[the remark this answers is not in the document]
davesque: Yeah, interesting thoughts.  And actually what you describe is I think a classic use-case for OO programming.  You have a method that does some piece of a process or algorithm and then you want to have another thing like that but just have it do something slightly different.  So you make a subclass of the class that contains the method and override it with your extra behavior.  That could help with the situation you describe where it's hard to change one helper function without impacting a lot of other things.  Having that helper function in a class would allow changing it in a subclass without impacting other things that depend on the base class.

thehomebrewnerd:
Closing this MR as we have decided to refactor the serialization approach in Woodwork. That refactor will have a significant impact on the changes needed to serialize to a single parquet file, and will be better to handle in a new PR after the serialization refactor is complete.

PR #1247, opened 2022-01-14
Validate email address logical types

### LINKED ISSUE #1080: Add validation to Email Address Logical Type
- As a user, I wish I could use Woodwork with to set columns as Email Address Logical Type, and have Woodwork validate my data (make sure the column(s) are actually email address.)
- This validation would make sure that **all non-nan values are actually Email Address**

jeff-hernandez:
Closes #1080


woodwork/logical_types.py
```diff
@@ -64,6 +64,9 @@ def transform(self, series):
                 raise TypeConversionError(series, new_dtype, type(self))
         return series
 
+    def validate(self, series):
+        return
```
jeff-hernandez, line 68: Not sure if it matters to use `return` or `pass` here

gsheni: raise a `NotImplementedError`?

woodwork/table_accessor.py
```diff
@@ -1247,7 +1248,9 @@ def _infer_missing_logical_types(
         parsed_logical_types[name] = _get_column_logical_type(
             series, logical_type, name
         )
-        updated_series = parsed_logical_types[name].transform(series)
+        parsed_logical_type = parsed_logical_types[name]
+        updated_series = parsed_logical_type.transform(series)
+        parsed_logical_type.validate(updated_series)
```
gsheni, line 1253: I don't think we want to call this here. Validation is only when the user defines/sets a specific column to be a Logical Type.

woodwork/table_accessor.py
```diff
@@ -541,6 +541,7 @@ def set_types(self, logical_types=None, semantic_tags=None, retain_index_tags=Tr
         for col_name, logical_type in logical_types.items():
             series = self._dataframe[col_name]
             updated_series = logical_type.transform(series)
+            logical_type.validate(updated_series)
```
gsheni, line 544: Is there sampling/head happening somewhere?

woodwork/logical_types.py
jeff-hernandez: Do we want to require every logical type to implement a validate function? I guess it would make sense to at least confirm that every logical type has been converted to the proper dtype as a result of the transform -- unless that's already being done elsewhere.

gsheni: Oh, right. No I don't (at this time).

woodwork/type_sys/inference_functions.py
```diff
@@ -120,7 +120,8 @@ def __call__(self, series: pd.Series) -> bool:
         regex = self.get_regex()
 
         # Includes a check for object dtypes
-        if not pdtypes.is_string_dtype(series.dtype):
+        is_object = pdtypes.is_object_dtype(series.dtype)
+        if is_object or not pdtypes.is_string_dtype(series.dtype):
             return False
```
jeff-hernandez, line 125: `is_string_dtype` can evaluate an object without strings to be true. I'm assuming we strictly want string dtypes

```
>>> pd.api.types.is_string_dtype(pd.Series([(3, 2)]))
True
```

woodwork/table_accessor.py
jeff-hernandez: No, I can add that

woodwork/table_accessor.py
jeff-hernandez: makes sense -- will remove

woodwork/logical_types.py
```diff
@@ -64,6 +64,10 @@ def transform(self, series):
                 raise TypeConversionError(series, new_dtype, type(self))
         return series
 
+    def validate(self, series, return_indices=False):
+        if return_indices:
+            return pd.Series(name=series.name, dtype="boolean")
```
jeff-hernandez, line 69: Alternatively, the default mask could be non-null values

woodwork/column_accessor.py
```diff
@@ -435,6 +435,12 @@ def box_plot_dict(self, quantiles=None, include_indices_and_values=True):
             include_indices_and_values=include_indices_and_values,
         )
 
+    @_check_column_schema
+    def validate(self, return_indices=False):
```
jeff-hernandez, line 439: At the column level, I think this should be `return_index` instead of `return_indices`

woodwork/logical_types.py
```diff
@@ -306,6 +310,16 @@ class EmailAddress(LogicalType):
...
 
+    def validate(self, series, return_indices=False):
+        regex = r"\b[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Z|a-z]{2,}\b"
+        valid = series.str.fullmatch(regex).astype("boolean")
+
+        if return_indices:
+            return valid.rename(series.name)
+
+        elif (~valid).any():
+            raise ValueError("email address not understood")
```
thehomebrewnerd, line 321: What do you think about update this message to be something like "Series `{series.name}` contains invalid email addresses"?

woodwork/column_accessor.py
thehomebrewnerd: I think the `return_indices` or `return_index` names make me think we are returning the values of the index where invalid values are found, not necessarily a series of True/False/nan values that indicate the validation status for each row.

I guess I'm not sure what we want here - do we want a list of index values for rows that contain invalid values or a mask column that can be used to select the invalid rows?

woodwork/table_accessor.py
```diff
@@ -1124,6 +1124,19 @@ def infer_temporal_frequencies(self, temporal_columns=None):
             self._dataframe, temporal_columns=temporal_columns
         )
 
+    @_check_table_schema
+    def validate(self, return_indices=False):
+        if return_indices:
+            indices = [
```
thehomebrewnerd, line 1130: Should this be the results for all columns or only the columns that contain invalid values?

woodwork/tests/accessor/test_column_accessor.py
```diff
@@ -965,3 +965,18 @@ def test_nullable_attribute(sample_df_pandas):
...
+    series = init_series(series, logical_type="EmailAddress")
+    assert series.ww.validate() is None
+
+    series.loc[4] = "bad_email"
+    series = init_series(series, logical_type="EmailAddress")
+    with pytest.raises(ValueError, match="email address not understood"):
+        series.ww.validate()
+
+    actual = series.ww.validate(return_indices=True)
+    expected = pd.Series([True, pd.NA, True, True, False], dtype="boolean")
```
thehomebrewnerd, line 981: Just wondering if the return value is a mask that indicates which rows are valid (True) and invalid (False), should the value for the `nan` input row actually be `True` since that row is valid? 

I don't think we can directly use this output as a mask to find invalid data if it contains `pd.NA` values, but I could be wrong.

woodwork/logical_types.py
tamargrey: Maybe we can note that the email address regex can be changed in the config if needed?

woodwork/logical_types.py
```diff
@@ -306,6 +310,16 @@ class EmailAddress(LogicalType):
 
     primary_dtype = "string"
 
+    def validate(self, series, return_indices=False):
+        regex = r"\b[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Z|a-z]{2,}\b"
```
tamargrey, line 314: should this reference the config `email_inference_regex` value?

woodwork/logical_types.py
tamargrey, line 321: I think this could be a good time to create a custom exception. If users want to catch the validation error, having a try/except on a `ValueError` could end up falling into the except if some part of validation raises an unrelated exception.

woodwork/column_accessor.py
tamargrey: My only problem with returning a mask is that for users to determine whether they actually have any invalid columns, they need to do a `series.all()` call or something. Returning just the invalid values makes it immediately obvious to users whether they have any invalid rows. We could return `None` if it's valid even. 

If we stick with a mask, I agree that the parameter name might want to use the word mask. If we return just invalid indices, I might want to make it `return_invalid_indices` just to make it super clear what's being returned.

woodwork/table_accessor.py
```diff
@@ -1124,6 +1124,19 @@ def infer_temporal_frequencies(self, temporal_columns=None):
...
+                self.ww[column].ww.validate(return_indices=True)
+                for column in self.columns
+            ]
+            return pd.concat(indices, axis=1)
+
+        else:
+            for column in self.columns:
+                self.ww[column].ww.validate(return_indices=False)
```
tamargrey, line 1138: this will only catch the first invalid column--maybe we want to offer an option that just returns the message or a True/False value that can be used like `is_valid`

woodwork/column_accessor.py
```diff
@@ -435,6 +435,12 @@ def box_plot_dict(self, quantiles=None, include_indices_and_values=True):
...
+        return self.logical_type.validate(
+            self._series_weakref(), return_indices=return_indices
```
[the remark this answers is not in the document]
jeff-hernandez: updated

woodwork/logical_types.py
jeff-hernandez: updated

woodwork/column_accessor.py
jeff-hernandez: @tamargrey I think that's a good idea. Returning invalid values instead of a mask is more helpful and easier to understand. With dask and koalas dataframes, it'd be better to let users decide whether or not to compute or write invalid values to disk. With that being said, I think it'd be more consistent across dataframes to return an empty series if the column is valid.

woodwork/logical_types.py
jeff-hernandez: updated

woodwork/tests/accessor/test_column_accessor.py
jeff-hernandez: When dtype is boolean, `nan` becomes `NA` and remains `NA` in the mask. You could still use a negated mask to find invalid values.

woodwork/table_accessor.py
jeff-hernandez: I think it should be results for all columns to keep data out of memory for dask and koalas

woodwork/tests/accessor/test_column_accessor.py
```diff
@@ -965,3 +965,18 @@ def test_nullable_attribute(sample_df_pandas):
         expected = EXPECTED_COLUMN_NULLABILITIES[key]
 
         assert actual is expected
+
+
+def test_validate(sample_df_pandas):
```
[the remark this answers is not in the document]
jeff-hernandez: We'll need special handling. When returning invalid values, we can return the dataframe and let users decide whether or not to compute or write to disk. Otherwise, we can just compute the scalar boolean value before raising an error if necessary.

woodwork/table_accessor.py
jeff-hernandez: I don't think it's necessarily crucial for this initial pass. I see this more like a nice to have that probably needs more design. Maybe we can circle back if it becomes necessary?

woodwork/logical_types.py
jeff-hernandez: Will we want custom exceptions for every logical type?

woodwork/logical_types.py
jeff-hernandez: Maybe the default behavior should return None

woodwork/logical_types.py
tamargrey: Maybe not an exception that's specific to each logical type, but one that indicates the data is invalid for its specified logical type?

jeff-hernandez: I see, so maybe we can use `TypeValidationError` for validation errors across logical types?

tamargrey: that sounds good to me!

woodwork/tests/logical_types/test_logical_types.py
```diff
@@ -187,3 +197,31 @@ def test_ordinal_transform(sample_series):
...
+    with pytest.raises(ValueError, match=match):
+        email_address.validate(series)
+
+    actual = email_address.validate(series, return_invalid_values=True)
+
+    if _is_dask_series(actual):
+        actual = actual.compute()
+
+    if _is_koalas_series(actual):
+        actual = actual.to_frame().to_pandas()["email"]
```
thehomebrewnerd, line 224: I think you can use the `to_pandas` testing utility function here and avoid checking the series type.

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -2986,3 +2976,33 @@ def test_nan_index_error(sample_df_pandas):
...
+    with pytest.raises(ValueError, match=match):
+        df.ww.validate()
+
+    actual = df.ww.validate(return_invalid_values=True)
+
+    if _is_dask_dataframe(actual):
+        actual = actual.compute()
+
+    if _is_koalas_dataframe(actual):
+        actual = actual.to_pandas()
```
thehomebrewnerd, line 3004: Use the `to_pandas` utility function here?

woodwork/tests/accessor/test_column_accessor.py
```diff
@@ -965,3 +963,32 @@ def test_nullable_attribute(sample_df_pandas):
...
+    with pytest.raises(ValueError, match=match):
+        series.ww.validate()
+
+    actual = series.ww.validate(return_invalid_values=True)
+
+    if _is_dask_series(actual):
+        actual = actual.compute()
+
+    if _is_koalas_series(actual):
+        actual = actual.to_frame().to_pandas()["email"]
```
thehomebrewnerd, line 991: Use the `to_pandas` utility function here?

woodwork/logical_types.py
```diff
@@ -306,6 +311,34 @@ class EmailAddress(LogicalType):
 
     primary_dtype = "string"
 
+    def validate(self, series, return_invalid_values=False):
```
thehomebrewnerd, line 314: We should add a docstring here to explain what this method does.

woodwork/column_accessor.py
```diff
@@ -435,6 +435,13 @@ def box_plot_dict(self, quantiles=None, include_indices_and_values=True):
             include_indices_and_values=include_indices_and_values,
         )
 
+    @_check_column_schema
+    def validate(self, return_invalid_values=False):
```
thehomebrewnerd, line 439: We should add a docstring here to explain what this method does.

woodwork/table_accessor.py
```diff
@@ -1124,6 +1124,29 @@ def infer_temporal_frequencies(self, temporal_columns=None):
...
+                concat = dd.concat
+            if _is_koalas_dataframe(self._dataframe):
+                concat = ks.concat
+
+            return concat(invalid_values, axis=1)
+
+        else:
+            for column in self.columns:
+                series = self.ww[column]
+                series.ww.validate(return_invalid_values=False)
```
thehomebrewnerd, line 1148: I think we can simplify this a little bit like this:
```
        invalid_values = []
        for column in self.columns:
            series = self.ww[column]
            values = series.ww.validate(return_invalid_values=return_invalid_values)
            if values is not None:
                invalid_values.append(values)

        if return_invalid_values:
            concat = pd.concat
            if _is_dask_dataframe(self._dataframe):
                concat = dd.concat
            if _is_koalas_dataframe(self._dataframe):
                concat = ks.concat
        
            return concat(invalid_values, axis=1)
```

woodwork/tests/logical_types/test_logical_types.py
jeff-hernandez: updated

woodwork/tests/accessor/test_column_accessor.py
jeff-hernandez: updated

woodwork/tests/accessor/test_table_accessor.py
jeff-hernandez: updated

woodwork/table_accessor.py
```diff
@@ -1124,6 +1124,32 @@ def infer_temporal_frequencies(self, temporal_columns=None):
...
+                invalid_values.append(values)
+
+        if return_invalid_values:
+            concat = pd.concat
+            if _is_dask_dataframe(self._dataframe):
+                concat = dd.concat
+            if _is_koalas_dataframe(self._dataframe):
+                concat = ks.concat
+
+            return concat(invalid_values, axis=1)
```
tamargrey, line 1151: I think that the concat call will be fine if different columns have different invalid values, but it's probably worth confirming that in a test. Maybe with a second EmailAddress column added to sample_df.

thehomebrewnerd (approved):

woodwork/table_accessor.py
jeff-hernandez: updated

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -2990,13 +2990,19 @@ def test_validate(sample_df):
         invalid_row = ks.from_pandas(invalid_row)
 
     df = df.append(invalid_row)
-    df.ww.init(logical_types={"email": "EmailAddress"})
-    match = "Series email contains invalid email addresses."
+    df["email_2"] = df["email"].fillna("bad_email_2")
+    types = dict(email="EmailAddress", email_2="EmailAddress")
+    df.ww.init(logical_types=types)
 
+    match = "Series email contains invalid email addresses."
```
tamargrey, line 2997: can the bad email addresses be at different index values between the two columns? the concat call should not have a problem with this, and it should still match the expected dataframe--there would just be some additional nan values.
image

thehomebrewnerd (commented):
Just randomly thought of one more thing we need to do here - we should add all the new methods to the API Reference document.

woodwork/tests/accessor/test_table_accessor.py
jeff-hernandez: updated

jeff-hernandez:
@tamargrey  I don't mind renaming to `validate_logical_type`

PR #1253, opened 2022-01-20
Tuples incorrectly inferred as EmailAddress

### LINKED ISSUE #1248: Tuple type columns incorrectly inferred as EmailAddress logical types
Woodwork's default inference currently identifies columns that have latlong like values stored in tuples as `EmailAddress`. We should update the EmailAddress inference so that these columns are identified as `Unknown` until proper LatLong type inference is added.

```python
import pandas as pd
import woodwork as ww

df = pd.DataFrame({"tuple_ints": pd.Series([(1, 2), (3, 4)]),
                   "tuple_strings": pd.Series([("1", "2"), ("3", "4")]),
                   "string_tuple": pd.Series(["(1, 2)", "(3, 4)"])})
df
```

```
  tuple_ints tuple_strings string_tuple
0     (1, 2)        (1, 2)       (1, 2)
1     (3, 4)        (3, 4)       (3, 4)
```

```python
df.ww.init()
df.ww
```
```
              Physical Type  Logical Type Semantic Tag(s)
Column                                                   
tuple_ints           string  EmailAddress              []
tuple_strings        string  EmailAddress              []
string_tuple         string       Unknown              []
```

dvreed77:
- This PR fixes the issue where tuples are incorrectly inferred as an EmailAddress
- Closes #1248

### DIFF
--- a/woodwork/tests/type_system/conftest.py
+++ b/woodwork/tests/type_system/conftest.py
@@ -3,24 +3,13 @@ import pandas as pd
 import pytest
 
 from woodwork.logical_types import Categorical, CountryCode, Double, Integer, Unknown
+from woodwork.tests.testing_utils import pd_to_dask, pd_to_koalas
 from woodwork.type_sys.inference_functions import (
     categorical_func,
     double_func,
     integer_func,
 )
 from woodwork.type_sys.type_system import TypeSystem
-
-
-def pd_to_dask(series):
-    dd = pytest.importorskip("dask.dataframe", reason="Dask not installed, skipping")
-    return dd.from_pandas(series, npartitions=2)
-
-
-def pd_to_koalas(series):
-    ks = pytest.importorskip(
-        "databricks.koalas", reason="Koalas not installed, skipping"
-    )
-    return ks.from_pandas(series)
 
 
 # Integer Inference Fixtures
@@ -163,6 +152,7 @@ def bad_dask_emails(bad_pandas_emails):
 
 @pytest.fixture
 def bad_koalas_emails(bad_pandas_emails):
+
     return [pd_to_koalas(series) for series in bad_pandas_emails]
 
 
--- a/woodwork/tests/type_system/conftest.py
+++ b/woodwork/tests/type_system/conftest.py
@@ -150,6 +150,9 @@ def bad_pandas_emails():
         pd.Series(["fl@alteryx.com", "b☃d@email.com", "good@email.com", np.nan]),
         pd.Series(["fl@alteryx.com", "@email.com", "good@email.com", "foo@bar.com"]),
         pd.Series(["fl@alteryx.com", "bad@email", "good@email.com", np.nan]),
+        pd.Series([1, 2, 3, 4]).astype("int"),
+        pd.Series([{"key": "value"}]).astype("O"),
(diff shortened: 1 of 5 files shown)


woodwork/tests/type_system/conftest.py
```diff
@@ -150,6 +150,9 @@ def bad_pandas_emails():
         pd.Series(["fl@alteryx.com", "b☃d@email.com", "good@email.com", np.nan]),
         pd.Series(["fl@alteryx.com", "@email.com", "good@email.com", "foo@bar.com"]),
         pd.Series(["fl@alteryx.com", "bad@email", "good@email.com", np.nan]),
+        pd.Series([1, 2, 3, 4]).astype("int"),
```
gsheni, line 153:
```suggestion
        pd.Series([1, 2, 3, 4]).astype("int"),
```

woodwork/tests/type_system/conftest.py
```diff
@@ -150,6 +139,9 @@ def bad_pandas_emails():
...
+        pd.Series([{"key": "value"}]).astype("O"),
+        pd.Series([(1, 2), (3, 4)]).astype("O"),
```
thehomebrewnerd, line 144: Minor point, but can we update these new series so that these are similar to the series we use above, where we have some valid email addresses, and only a single bad value? Our inference should be setup to fail if only a single value fails to match the regex, and having only a single bad value might make that a little more obvious.

woodwork/tests/type_system/conftest.py
```diff
@@ -160,6 +152,7 @@ def bad_dask_emails(bad_pandas_emails):
 
 @pytest.fixture
 def bad_koalas_emails(bad_pandas_emails):
+
```
thehomebrewnerd, line 155: Super minor nit-pick...remove extra blank line.

woodwork/type_sys/inference_functions.py
```diff
@@ -117,6 +117,7 @@ def __init__(self, get_regex: Callable[[], str]):
         self.get_regex = get_regex
 
     def __call__(self, series: pd.Series) -> bool:
+        series = series.dropna()
```
thehomebrewnerd, line 120: Since we are adding this dropna, maybe we should add a test to make sure in our bad emails test that a series of all `nan` values does not get inferred as EmailAddress?

woodwork/tests/type_system/conftest.py
dvreed77: I can add those cases as well, but the original issue was a series of tuples-only and that was getting inferred as an email address, so it seems like we would want to check that explicitly maybe?

thehomebrewnerd: When the issue was discovered, it just happened to be with a series of all tuples, but there wasn't really anything special about that input series other than the fact that it contained tuples. We can check both situations, but generally I don't think we need to go too far with creating all kinds of different combinations of inputs here.

thehomebrewnerd: After discussion with @dvreed77 mixing tuples and strings creates issues for the Koalas tests, so will leave this as it is. The changes in this PR should still prevent a column with valid email addresses and a single tuple from being inferred as an email address.

thehomebrewnerd (approved):

PR #1255, opened 2022-01-21
Allow Tuple NaN LatLong Values

### LINKED ISSUE #1226: Require that all LatLong have 1 NaN representation

dvreed77:
- Allow Tuple NaN LatLong Values
- Closes #1226


woodwork/tests/accessor/test_statistics.py
```diff
@@ -673,8 +673,8 @@ def test_describe_accessor_method(describe_df):
                 "physical_type": expected_dtype,
                 "logical_type": ltype(),
                 "semantic_tags": {"custom_tag"},
-                "count": 6,
-                "nan_count": 2,
+                "count": 7,
+                "nan_count": 1,
```
thehomebrewnerd, line 677: This behavior change is the only thing that gives me a little hesitation. We are now saying that `(np.nan, np.nan)` is not a null value, which is technically true I suppose. Before we would have replaced it with a single `nan` and recognized it as null.

Might not be worth the trouble it's worth to update the stats to handle this edge case, but now if a column has a lot of `(nan, nan)` tuples, we won't really identify that for the user as a potential highly null column.

Just wanted to mention in case anyone has a strong opinion on this behavior change.

dvreed77: is there a way we can overload nan checking?

dvreed77: added a check in describe to handle this case

woodwork/utils.py
```diff
@@ -543,3 +539,12 @@ def _replace_nan_strings(df):
         df[col] = replaced_series
 
     return df
+
+
+def _is_tuple_nan(value):
```
tamargrey, line 544: Not sure I'd expect `_is_tuple_nan(np.nan)` to be True based off the name. There's a somewhat similar util `_is_null_latlong`, but that just focuses on individual values within the tuple and not the tuple iteslf. It does make me wonder if somehow they could be combined or both used for this purpose.

tamargrey (commented):
I think the docstring of `_reformat_to_latlong` and the LatLong class docstring now need to be updated to reflect the new null values that are possible in LatLong columns.

woodwork/utils.py
dvreed77: @tamargrey good catch. I think I should add another condition there to see if its a tuple

woodwork/utils.py
```diff
@@ -291,12 +291,23 @@ def _is_valid_latlong_value(val, bracket_type=tuple):
...
+    if isinstance(value, str):
+        return value in ww.config.get_option("nan_values")
+    if isinstance(value, list):
+        return False
+    return pd.isnull(value)
+
+
+def _is_latlong_nan(value):
+    """Checks if a LatLong value is NaN"""
+    if isinstance(value, tuple):
```
tamargrey, line 307: Koala stores the latlongs as lists, because it can't handle tuples. You may need to include `list` here as well--the fact that it's not currently failing makes me wonder if I'm missing something here.

woodwork/tests/utils/test_utils.py
```diff
@@ -326,18 +326,22 @@ def test_to_latlong_float():
...
+    assert _is_nan("NaN")
+    assert _is_nan("<NA>")
+    assert _is_nan("")
+    assert _is_nan("null")
 
-    assert not _is_null_latlong([None, 1, 3])
-    assert not _is_null_latlong("none")
-    assert not _is_null_latlong(0)
-    assert not _is_null_latlong(False)
+    assert not _is_nan([None, 1, 3])
```
tamargrey, line 340: I like the inclusion of all the falsy values here. Thoughts on also having the empty list `[]` or some other data structure like a set or dict?

woodwork/utils.py
dvreed77: @tamargrey it falls through and gets fixed down here: https://github.com/alteryx/woodwork/pull/1255/files#diff-05318c9c3bf0f63e471074de1ccab68b2c818c14ba9ae3a0220e510264467d44L248

tamargrey: I was more thinking that something like `[np.nan, np.nan]` would get passed in and not get recognized as needing to have its individual values checked so `_is_nan([np.nan, np.nan])` will get called and return False, which won't be doing the check we want. 

in the latlong reformatting, this doesn't end up making a difference, because no transformation was needed, but in `describe`, wouldn't it not get counted as a nan?

dvreed77: Ah, I see what you mean. Will fix

tamargrey (approved):
LGTM!

PR #1265, opened 2022-01-24
WIP: Expand correlation measures

### LINKED ISSUE #1157: Add Pearson correlation coefficient calculation to Woodwork DataFrames
- As a user, I wish I could use Woodwork to calculate the pearson correlation coefficient for pairs of columns in my Woodwork DataFrame in addition to the mutual info calculation and have a more general correlation calculation that returns `max(abs(Pearson), m_i)` to utilize pearson in situations where it is better at picking up relationships between columns than mutual info is.

### Implementation details
The only correlation calculation currently in Woodwork is through `df.ww.mutual_information_dict`/`df.ww.mutual_information`. If we want to add a pearson calculation that's used if it's greater than mutual info, it may not make sense to just add that into a method that is called "mutual information".

There also may be users who would only want mutual information or pearson separately. There are a few options for how we could achieve this:
1. Three separate methods: `df.ww.mutual_information` and `df.ww.pearson_correlation` calculate and return only the correlations named, and `df.ww.correlation` which calculates both and returns the max of the two. We'd need to determine if each method needs a separate method that returns a dataframe from the one that returns a dictionary.
2. One new method: `df.ww.correlation` that, by default, gets the maximum of the two but can have a parameter `correlation_type='mutual_info'/'pearson'/'all'` that determines the type of correlation calculation used. Note - this would be a breaking change if we remove the mutual info methods
3. Expand the capabilities of `df.ww.mutual_information` to be able to calculate Pearson and return the max of pearson and mutual info or either individually if specified. This seems misleading to me (even if pearson isn't used by default).

Note: We should handle the categorization and nan replacement the same for both correlation calculations, and `df.ww.get_valid_mi_columns` should return the same for an equivalent method for pearson calculations.

rwedge:
- Reworks how Mutual Information is calculated
- Adds support for Pearson Correlation
- Adds general method for calculating one or all of MI, Pearson, and the absolute max of both
- Closes #1157


gsheni:
@rwedge Given that all the tests pass, is this ready for review?

rwedge:
@gsheni yes

woodwork/statistics_utils/_get_dependence_dict.py
```diff
@@ -0,0 +1,201 @@
...
+    nrows=None,
+    include_index=False,
+    callback=None,
+    extra_stats=False,
+    min_shared=25,
+    random_seed=0,
+):
+    """Calculates dependence measures between all pairs of columns in the DataFrame that
+    support measuring dependence. Logical Types that are supported are as
+    follows:  Boolean, Categorical, CountryCode, Datetime, Double, Integer, Ordinal,
```
thehomebrewnerd, line 28: Wondering if we should make this supported types list more generic instead of listing specific types to make it less likely to get outdated as we change/add types? Are the nullable types supported? What about Age? If we just listed boolean, categorical, datetime and numeric columns instead, would that cover it?

Same comment applies to other docstrings where this is mentioned.

woodwork/statistics_utils/_get_dependence_dict.py
```diff
@@ -0,0 +1,201 @@
...
+                    result.update({measure_type: np.nan for measure_type in measure})
+                else:
+                    num_union = (notna_mask[a_col] | notna_mask[b_col]).sum()
+                    if calc_mutual:
+                        mi_score = adjusted_mutual_info_score(data[a_col], data[b_col])
+                        mi_score = mi_score * num_intersect / num_union
+                        if "mutual" in measure:
+                            result["mutual"] = mi_score
+                    if calc_pearson:
+                        pearson_score = np.corrcoef(data[a_col], data[b_col])[0, 1]
```
thehomebrewnerd, line 167: I'm not too familiar with how these metrics work, so this may be a dumb question, but do we want to use the binned numeric data for Pearson?

thehomebrewnerd (commented):
Still working my way through this, but in the meantime were you planning to add tests for the uncovered lines?

woodwork/statistics_utils/_get_dependence_dict.py
```diff
@@ -0,0 +1,278 @@
...
+    nrows=None,
+    include_index=False,
+    callback=None,
+    extra_stats=False,
+    min_shared=25,
+    random_seed=0,
+):
+    """Calculates dependence measures between all pairs of columns in the DataFrame that
+    support measuring dependence. Supports boolean, categorical, datetime, and numeric data.
+    Call woodwork.utils.get_valid_dependence_types for a complete list of supported Logical Types.
```
rwedge, line 30: Should probably change to

Call woodwork.utils.get_valid_mi_types  and woodwork.utils.get_valid_pearson_types for complete lists of supported Logical Types.

(and update wording similarly in table_accessor.py)

woodwork/statistics_utils/_get_dependence_dict.py
```diff
@@ -0,0 +1,278 @@
...
+
+    unit = "calculations"
+    if calc_order[0] == "pearson":
+        pearson_types = get_valid_pearson_types()
+        pearson_columns = _get_valid_columns(dataframe, pearson_types)
+        valid_columns = pearson_columns
+    if "mutual" in calc_order:
+        mi_types = get_valid_mi_types()
+        mutual_columns = _get_valid_columns(dataframe, mi_types)
+        valid_columns = mutual_columns
```
tamargrey, line 89: might be worth a comment that the valid pearson columns are a subset of the valid mutual info columns, which is why this isn't a problem

woodwork/statistics_utils/_get_dependence_dict.py
```diff
@@ -0,0 +1,278 @@
...
+    # Assume 1 unit for preprocessing, n for handling null, m for make categorical
+    # (p*p+p)/2 for pearson and (m*m+m)/2 for mutual
+    total_loops = 1 + n + m + (p * p + p) / 2 + (m * m + m) / 2
+    callback_caller = CallbackCaller(callback, unit, total_loops, start_time=start_time)
+    callback_caller.update(1)
+
+    data = data.dropna()
+    # cast nullable type to non-nullable
+    for col_name in data:
+        column = dataframe.ww.schema.columns[col_name]
```
tamargrey, line 136: woodwork nitpick: should be able to do `dataframe.ww.columns[col_name]`

rwedge:
Open question: add `get_valid_pearson_columns` to match `get_valid_mi_columns`?

tamargrey (dismissed):
lgtm!

bchen1116 (commented):
Just left a few nit picks for now!

rwedge:
@bchen1116 thanks for the review, updated

bchen1116 (dismissed):
LGTM! I like the thoroughness of the tests you included

tamargrey (approved):
lgtm!

woodwork/statistics_utils/_bin_numeric_cols_into_categories.py
```diff
@@ -1,33 +1,24 @@
...
-def _make_categorical_for_mutual_info(schema, data, num_bins):
+def _bin_numeric_cols_into_categories(schema, data, num_bins):
     """Transforms dataframe columns into numeric categories so that
-    mutual information can be calculated
+    dependence can be calculated
 
     Args:
         schema (woodwork.TableSchema): Woodwork typing info for the data
-        data (pd.DataFrame): dataframe to use for calculating mutual information
+        data (dict[pd.Series]): dictionary of Pandas series to use for
```
bchen1116, line 10: Nit: Periods for these descriptions

bchen1116 (approved):
Looks good Roy! Great work getting this out. I left a few comments on nits with the documentation and a few questions, but nothing blocking. Thanks for breaking the large `dependence_dict` file into multiple smaller files. Definitely helps with readability imo!

woodwork/statistics_utils/_calculate_max_dependence_for_pair.py
```diff
@@ -0,0 +1,30 @@
...
+    result, returned_measures, min_shared, extra_stats
+):
+    """Calculates the max dependence measure for a pair of columns. While
+    the magnitude of a negatively correlated value is used to determine the max,
+    the original (negative) value will be used in the "max" column.
+    """
+    # if pearson was not measured, mutual info must be max
+    if "pearson" in result:
+        score = pd.Series(
+            [result["mutual_info"], abs(result["pearson"])],
```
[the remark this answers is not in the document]
rwedge: Since we are determining the max, we are in a scenario where both mutual and Pearson were calculated.  Since the columns we can use for Pearson are a subset of the columns we can use for Mutual Info, we'll always have a mutual info score but might not have a Pearson score.

woodwork/statistics_utils/_cast_nullable_int_and_datetime_to_int.py
```diff
@@ -0,0 +1,12 @@
+from woodwork.logical_types import IntegerNullable
+
+
+def _cast_nullable_int_and_datetime_to_int(data_dict, woodwork_columns):
+    for col_name in data_dict:
+        column = woodwork_columns[col_name]
+        if isinstance(column.logical_type, IntegerNullable):
+            cur_dtype = data_dict[col_name].dtype
+            new_dtype = cur_dtype.name.lower()  # e.g. Int64 -> int64
+            data_dict[col_name] = data_dict[col_name].astype(new_dtype)
```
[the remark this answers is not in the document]
rwedge: since `IntegerNullable`'s physical type is always "Int64" that would be sufficient, but I think this if statement should be re-written for a more generic check for nullable int physical types instead of `IntegerNullable` specifically

PR #1269, opened 2022-01-27
Update box plot bounds to min and max if necessary

### LINKED ISSUE #1267: Box plot dictionary has misleading values for high and low bounds in some cases.
As a Woodwork user, I want to be able to use the output of `series.ww.box_plot_dict` to build a box plot that makes sense. 

Currently, Woodwork's use of IQR to determine the high and low outlier bounds can end up in situations where the high or low bound can be above or below the max or min values of the series. If a box plot is then built using these bounds, it's misleading, as it points to the range of the data being wider than it actually is. 

An example using the following quantiles: `{0.0: 1 , 0.25: 42, 0.5: 145, 0.75: 284, 1.0: 977}` and our equation for calulating the bounds
```
IQR = q_3 - q_1
low_bound = q_1 - 1.5IQR
high_bound = q_3 + 1.5IQR
```
leaves us with a high bound of 647, which makes sense, but a low bound of -321, which is way below the min value of 1.

We should update our high and low bounds to be

```python
"low_bound": max(low_bound, quantiles['0.0'])
"high_bound": min(high_bound, quantiles['0.75'])
```

tamargrey:
- Requires min and max values in quantiles so they can be used to set the high and low bounds when no outliers are present
- Closes #1267

### DIFF
--- a/woodwork/column_accessor.py
+++ b/woodwork/column_accessor.py
@@ -406,9 +406,9 @@ class WoodworkColumnAccessor:
                 Defaults to True.
 
         Note:
-            The minimum quantiles necessary for building a box ploy using the IQR method are the
+            The minimum quantiles necessary for building a box plot using the IQR method are the
             minimum value (0.0 in the quantiles dict), first quartile (0.25), third quartile (0.75), and maximum value (1.0).
-            If no quantiles are providex, the following quantiles will be calculated:
+            If no quantiles are provided, the following quantiles will be calculated:
             {0.0, 0.25, 0.5, 0.75, 1.0}, which correspond to {min, first quantile, median, third quantile, max}.
 
         Returns:
--- a/woodwork/column_accessor.py
+++ b/woodwork/column_accessor.py
@@ -415,8 +415,8 @@ class WoodworkColumnAccessor:
             (dict[str -> float,list[number]]): Returns a dictionary containing box plot information for the Series.
                 The following elements will be found in the dictionary:
 
-                - low_bound (float): the lower bound below which outliers lay - to be used as a whisker
-                - high_bound (float): the high bound above which outliers lay - to be used as a whisker
+                - low_bound (float): the lowest data point in the dataset excluding any outliers - to be used as a whisker
+                - high_bound (float): the highest point in the dataset excluding any outliers - to be used as a whisker
                 - quantiles (list[float]): the quantiles used to determine the bounds.
                     If quantiles were passed in, will contain all quantiles passed in. Otherwise, contains the five
                     quantiles {0.0, 0.25, 0.5, 0.75, 1.0}.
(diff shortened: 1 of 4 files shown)


woodwork/column_accessor.py
```diff
@@ -406,10 +406,10 @@ def box_plot_dict(self, quantiles=None, include_indices_and_values=True):
                 Defaults to True.
 
         Note:
-            The minimum quantiles necessary for outlier detection using the IQR method are the
-            first quartile (0.25) and third quartile (0.75). If these keys are missing from the quantiles dictionary,
-            the following quantiles will be calculated: {0.0, 0.25, 0.5, 0.75, 1.0}, which correspond to
-            {min, first quantile, median, third quantile, max}.
+            The minimum quantiles necessary for building a box ploy using the IQR method are the
```
thehomebrewnerd, line 409: `ploy` -> `plot`

woodwork/column_accessor.py
```diff
@@ -406,10 +406,10 @@ def box_plot_dict(self, quantiles=None, include_indices_and_values=True):
...
+            minimum value (0.0 in the quantiles dict), first quartile (0.25), third quartile (0.75), and maximum value (1.0).
+            If no quantiles are providex, the following quantiles will be calculated:
```
thehomebrewnerd, line 411: `providex` -> `provided`

woodwork/statistics_utils.py
```diff
@@ -463,10 +463,10 @@ def _get_box_plot_info_for_column(
             Defaults to True.
 
     Note:
-        The minimum quantiles necessary for outlier detection using the IQR method are the
-        first quartile (0.25) and third quartile (0.75). If these keys are missing from the quantiles dictionary,
-        the following quantiles will be calculated: {0.0, 0.25, 0.5, 0.75, 1.0}, which correspond to
-        {min, first quantile, median, third quantile, max}.
+        The minimum quantiles necessary for building a box ploy using the IQR method are the
```
thehomebrewnerd, line 466: `ploy` -> `plot`

woodwork/statistics_utils.py
```diff
@@ -463,10 +463,10 @@ def _get_box_plot_info_for_column(
...
+        minimum value (0.0 in the quantiles dict), first quartile (0.25), third quartile (0.75), and maximum value (1.0).
+        If no quantiles are providex, the following quantiles will be calculated:
```
thehomebrewnerd, line 468: `providex` -> `provided`

thehomebrewnerd (commented):
Just a couple docstring fixes, but otherwise looks good to me.

thehomebrewnerd (approved):
LGTM!

dvreed77 (approved):
LGTM!

PR #1281, opened 2022-02-03
Improve Datetime Frequency Inference

### LINKED ISSUE #1208: Datetime column frequency inference V2
In #1202, we added the ability to infer frequencies for temporal (datetime and timedelta) columns in a Woodwork dataframe. This inference is called for by the user and is not stored anywhere on the DataFrame. It also only works for pandas DataFrames. 

There are a number of improvements we could make. They are listed here, and we can create individual issues if any of these become necessary to implement.

- Store inferred frequency in the Woodwork typing info
  - Should likely be stored on the Datetime logical type object
  - If stored on Datetime logical type object, we need to implement validation in `Datetime.transform` (or the validation method if #1192 has implemented a `LogicalType.validate` method)
- If we can store the frequency, automatically infer the frequency for time index columns and store it
- Add frequency inference as column accessor method as well - can automatically store on logical type
- Sort columns passed to `infer_temporal_frequencies` before trying to infer frequencies
  - from @freddyaboulton : I would say no until we figure out how often/useful it is to infer frequencies on multiple datetime columns. My hunch is that most users will only care about the time_index which will already be sorted so let's not double sort for now. Of course, users can have a non-sorted time_index but I don't think there's anything we can do about that.
- Only pass a single subset (must be taken consecutively from series) to use for frequency inference
  - from @freddyaboulton: I am on the fence about this. I think we would need to do some experimentation to see what results we get with taking a sample vs not. My guess is that it's ok to do so. If we can't infer a frequency on the first k rows, we're not going to be able to infer a frequency on the whole column. In my opinion the right thing to do is to take the first-order diff of all rows and find the frequency that works the best for the most rows (like what's mentioned in #1165 for datetime format inference) but that seems like a v2 improvement?

- Use multiple subsets of data (that's long enough) for frequency inference and take the mode of non-null inferred frequencies 
- Implement for Dask and Koalas dataframes as well (if using just a subset of data, this will be doable)

dvreed77:
- Improves Datetime frequency inference when some datetime values are slightly misaligned.
- Closes #1208


woodwork/statistics_utils/frequency_inference/_determine_most_likely_freq.py
```diff
@@ -0,0 +1,18 @@
+from ._constants import FREQ_INFERENCE_THRESHOLD, NON_INFERABLE_FREQ
+
+
+def _determine_most_likely_freq(alias_dict):
+    n_total = sum(alias_dict.values())
+    sorted_freqs = sorted(alias_dict.items(), key=lambda item: item[1], reverse=True)
+
+    most_likely_alias, most_likely_count = sorted_freqs[0]
+
+    if most_likely_count / n_total < FREQ_INFERENCE_THRESHOLD:
```
freddyaboulton, line 10: This might be a useful case to capture in an error code?

woodwork/statistics_utils/frequency_inference/_build_freq_dataframe.py
```diff
@@ -0,0 +1,13 @@
+from ._constants import CANDIDATE_COLUMN_NAME
+
+
+def _build_freq_dataframe(df):
+    def to_series(row):
+        for alias in row[CANDIDATE_COLUMN_NAME]:
```
freddyaboulton, line 6: We only care about the most likely frequency right? I think we can trim down on the size of this dataframe (and maybe total run time) if we only determine which rows belong to a window where the most likely frequency was found.

freddyaboulton: Also, I think this is only created to determine the "true" start and end dates according to the most likely frequency. I wonder if we can do that by just iterating over `candidate_df` and not having to create another dataframe.

freddyaboulton (commented):
@dvreed77 Thanks for your work on this! I think the overall approach looks good but will take longer to evaluate all the possible cases. Hopefully the commented-out unit tests can help with that! 

I realize this is still a WIP but I wanted to leave some thoughts ahead of your meeting since I won't be able to attend.

woodwork/statistics_utils/frequency_inference/_build_freq_dataframe.py
dvreed77: @freddyaboulton very good points! thanks

dvreed77: I was able to remove this completely and generate bounds for a given alias during the windowing.

woodwork/statistics_utils/frequency_inference/_infer_frequency.py
```diff
@@ -0,0 +1,135 @@
...
+            debug_obj=InferDebug(
+                message = DataCheckMessageCode.DATETIME_SERIES_IS_EMPTY,
+            ),
+        )
+
+    nan_values = _determine_nan_values(observed_ts)
+    duplicate_values = _determine_duplicate_values(observed_ts)
+
+    actual_range_start = observed_ts_clean.min().isoformat()
+    actual_range_end = observed_ts_clean.max().isoformat()
```
tamargrey, line 60: Since we check that the series is monotonic increasing, can we just take the value at the 0th and nth index?

tamargrey: Or does that stop them from being "Actual" min and max values?

woodwork/tests/fixtures/datetime_freq.py
```diff
@@ -0,0 +1,987 @@
...
+    return {
+        "dates": dates,
+        "expected_debug_obj": expected_debug_obj
+    }
+
+
+
+
+
+# def case3():
```
rwedge, line 459: lots of commented out code here; plan to keep?

rwedge (commented):
should we split out the test_statistics.py file as well? (or at split off these tests and handle the rest later?

woodwork/statistics_utils/frequency_inference/_generate_freq_candidates.py
```diff
@@ -0,0 +1,43 @@
...
+        a dictionary containing the following keys:
+            - alias (str): The Pandas Freq Alias
+            - count (int): The number of windows where this alias is valid
+            - min_dt (pd.TimeStamp): The earliest timestamp for this frequency.
+            - max_dt (pd.TimeStamp): The latest timestamp for this frequency.
+    """
+    
+    alias_dict = {}
+    for window in time_series.rolling(WINDOW_LENGTH):
+        if len(window) == WINDOW_LENGTH:
```
tamargrey, line 22: If the series has fewer rows than the window length, you don't get any frequency calculations other than the frequency of the whole series. Is this okay for what we want? If so, we should note it somewhere. If not, then we might as well make the minimum length the length of the window instead of 3

woodwork/statistics_utils/frequency_inference/_infer_frequency.py
```diff
@@ -0,0 +1,135 @@
...
+            actual_range_end=actual_range_end,
+            estimated_freq=most_likely_freq,
+            estimated_range_start=estimated_range_start,
+            estimated_range_end=estimated_range_end,
+            missing_values=missing_values,
+            duplicate_values=duplicate_values,
+            extra_values=extra_values,
+            nan_values=nan_values
+        ),
+        debug=debug
```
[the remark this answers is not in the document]
dvreed77: good point

woodwork/statistics_utils/frequency_inference/_infer_frequency.py
dvreed77: no, I think you are right, good catch

docs/source/guides/statistical_insights.ipynb
```diff
@@ -209,6 +209,209 @@
...
     "With the normally distributed set of data, the likelyhood of outliers is closer to what we'd expect, around .7%. "
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Inferring Frequency from Noisy Timeseries Data\n",
+    "\n",
+    "`df.ww.infer_temporal_frequencies` will infer the observation frequency (daily, biweekly, yearly, etc) of each temporal column, even on noisy data. If a temporal column is predominantly of a single frequency, but is noisy in any way (ie. contains duplicate timestamps, nans, gaps, or timestamps that do not align with overall frequency), this table accessor will provide the most likely frequency, and the rows of data that do not adhere to this frequency."
```
tamargrey, line 219: Suggestion for a slight wording change here from 

> this table accessor will provide the most likely frequency, and the rows of data that do not adhere to this frequency.

to

> this table accessor method will provide the most likely frequency as well as information about the rows of data that do not adhere to this frequency.

docs/source/guides/statistical_insights.ipynb
```diff
@@ -209,6 +209,209 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "```python\n",
+    "{\n",
+    "    'dt1': 'H', \n",
+    "    'dt2': 'B'\n",
+    "}\n",
```
tamargrey, line 253: The code cells will get executed when the docs are built, and the output here will be shown under the code cell. It will have slightly less nice formatting, but is still pretty readable to me. 

image

docs/source/guides/statistical_insights.ipynb
```diff
@@ -209,6 +209,209 @@
...
+   "metadata": {},
+   "source": [
+    "### Debug Object Description\n",
+    "\n",
+    "The debug object contains the following information:\n",
+    "\n",
+    "- **actual_range_start**: an ISO 8601 formatted string of the observed start time\n",
+    "- **actual_range_end**: an ISO 8601 formatted string of the observed end time\n",
+    "- **message**: a message to describe why the frequency cannot be inferred\n",
+    "- **estimated_freq**: a message to describe why the frequency cannot be inferred\n",
```
tamargrey, line 320: since `estimated_freq` is a large part of a user's goal in getting the debug object, it would be nice to add a code cell showing how you get that value from the output of `infer_temporal_frequencies`. You could stick it right before this, noting that the first element of the tuple being `None` indicates that there was not a perfect frequency. Then that leads into digging into what the rest of the elements of the debug dictionary mean.

docs/source/release_notes.rst
```diff
@@ -3,17 +3,19 @@
...
         * Updated ``__str__`` output for ``Ordinal`` logical types (:pr:`1340`)
     * Changes
     * Documentation Changes
     * Testing Changes
 
-.. Thanks to the following people for contributing to this release:
-    :user:`bchen1116`
+    Thanks to the following people for contributing to this release:
+    :user:`dvreed77`, :user:`bchen1116`
+>>>>>>> origin/main
```
tamargrey, line 18: Extra line left over from merge conflicts!

woodwork/statistics_utils/_infer_temporal_frequencies.py
```diff
@@ -12,7 +14,8 @@ def _infer_temporal_frequencies(dataframe, temporal_columns=None):
         temporal_columns (list[str], optional): Columns for which frequencies should be inferred. Must be columns
             that are present in the DataFrame and are temporal in nature. Defaults to None. If not
             specified, all temporal columns will have their frequencies inferred.
-
+        debug (boolean): A flag used to determine if more information should be returned for each temporal column if
+                no uniform frequency was found.
     Returns:
         (dict): A dictionary where each key is a temporal column from the DataFrame, and the
```
tamargrey, line 20: This should be updated to indicate that the return will be a tuple when debug=True!

woodwork/table_accessor.py
```diff
@@ -1082,6 +1082,8 @@ def infer_temporal_frequencies(self, temporal_columns=None):
             temporal_columns (list[str], optional): Columns for which frequencies should be inferred. Must be columns
                 that are present in the DataFrame and are temporal in nature. Defaults to None. If not
                 specified, all temporal columns will have their frequencies inferred.
+            debug (boolean): A flag used to determine if more information should be returned for each temporal column if
+                no uniform frequency was found.
 
         Returns:
             (dict): A dictionary where each key is a temporal column from the DataFrame, and the
```
tamargrey, line 1089: This also should be updated to indicate that the return will be a tuple when debug=True

woodwork/statistics_utils/frequency_inference/_infer_frequency.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+def inference_response(inferred_freq, debug_obj, debug):
+    if debug:
+        return (inferred_freq, dataclasses.asdict(debug_obj))
+    else:
+        return inferred_freq
+
+
+def infer_frequency(
```
tamargrey, line 24: If this and `inference_response` are intended as public functions, they should get entries in our `api_reference.rst`

tamargrey (commented):
A lot of the new files (`_determine_duplicate_values`, `_determine_missing_values`, and `_determine_nan_values` are all examples) seem to be almost entirely uncovered by codecov. If the tests that would cover these files are running, I suspect it's an issue with how these files are listed in an init file somewhere, since that's something we've seen in the past.

Submitting this now so I don't forget, but I still need to do another pass through the tests and `frequency_inference` files!

docs/source/guides/statistical_insights.ipynb
dvreed77: done

docs/source/guides/statistical_insights.ipynb
dvreed77: removed.

docs/source/guides/statistical_insights.ipynb
dvreed77: Done

woodwork/statistics_utils/_infer_temporal_frequencies.py
dvreed77: done

docs/source/release_notes.rst
dvreed77: done

woodwork/statistics_utils/frequency_inference/_determine_missing_values.py
```diff
@@ -0,0 +1,47 @@
...
+from ._constants import ESTIMATED_COLUMN_NAME, OBSERVED_COLUMN_NAME
+from ._get_ranges import _get_ranges
+from ._types import RangeObject
+
+
+def _determine_missing_values(estimated, observed):
+    """Calculate Missing Values in time_series:
+
+    Args:
+        observed_ts (Series): The observed time_series.
```
tamargrey, line 12: missing `estimated` from the args

woodwork/statistics_utils/frequency_inference/_determine_extra_values.py
```diff
@@ -0,0 +1,48 @@
...
+from ._constants import ESTIMATED_COLUMN_NAME, OBSERVED_COLUMN_NAME
+from ._get_ranges import _get_ranges
+from ._types import RangeObject
+
+
+def _determine_extra_values(estimated, observed):
+    """Calculate Extra Values in time_series:
+
+    Args:
+        observed_ts (Series): The observed time_series.
```
tamargrey, line 12: also missing estimated from the args

woodwork/statistics_utils/frequency_inference/_determine_most_likely_freq.py
```diff
@@ -0,0 +1,20 @@
+from ._constants import FREQ_INFERENCE_THRESHOLD, NON_INFERABLE_FREQ
+
+
+def _determine_most_likely_freq(alias_dict, threshold=FREQ_INFERENCE_THRESHOLD):
```
tamargrey, line 4: The dosctrings for the other helper utils are very useful for understanding what's going on. Could you include one here as well?

woodwork/statistics_utils/frequency_inference/_determine_most_likely_freq.py
```diff
@@ -0,0 +1,20 @@
...
+    most_likely_alias = sorted_aliases[0]["alias"]
+    most_likely_count = sorted_aliases[0]["count"]
+
+    if most_likely_count / n_total < threshold:
+        return None
+
+    if most_likely_alias is not NON_INFERABLE_FREQ:
+        return most_likely_alias
+    else:
+        return None
```
tamargrey, line 20: This is uncovered by tests, but I'm not sure how you'd cover it. Since for the `most_likely_alias` to be one of the `NON_INFERRABLE_FREQ`, it'd have to have been inferred in one of the rolling windows by `pd.infer_freq`, which isn't possible, right?

tamargrey (commented):
Still have to review the tests, but wanted to get this out to you now. Part three coming soon! 

The way you've put `_infer_frequency.py` together makes it very easy to read and understand what's happening. Considering how lengthy it's needed to be, it's wonderfully put together! 

If we ever are tasked with optimizing it, it may need to lose some of its clarity in order to batch some operations in the same iteration over the data (this is a half-baked thought, so it may be less possible than I'm thinking). But I'd stick that in a V2 issue of potential improvements.

woodwork/statistics_utils/frequency_inference/_determine_duplicate_values.py
```diff
@@ -0,0 +1,41 @@
...
+
+    observed_dupes = observed_df[observed_df[OBSERVED_COLUMN_NAME] == pd.Timedelta(0)]
+
+    if len(observed_dupes) == 0:
+        return []
+
+    ranges = _get_ranges(observed_dupes.index)
+    out = []
+
+    for start_idx, end_idx in ranges:
```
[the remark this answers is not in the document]
dvreed77: I like it!

woodwork/tests/accessor/test_statistics.py
```diff
@@ -1385,23 +1386,44 @@ def test_box_plot_optional_return_values(outliers_df):
...
+)
+@pytest.mark.parametrize(
+    "expected_call_args",
+    [{}, {"debug": True}, {"temporal_columns": ["2D_freq", "3M_freq"]}],
+)
+def test_infer_temporal_frequencies(
+    infer_frequency, expected_call_args, datetime_freqs_df_pandas
+):
+
+    # TODO: Add support for Dask and Koalas DataFrames
```
tamargrey, line 1401: Koalas -> Spark

woodwork/statistics_utils/frequency_inference/_determine_most_likely_freq.py
dvreed77: I should be able to create a test that reaches this

woodwork/statistics_utils/frequency_inference/_get_ranges.py
```diff
@@ -0,0 +1,36 @@
...
+    Returns:
+        (DatetimeIndex): The estimated datetime index
+    """
+    start_idx = idx_series[0]
+    last_idx = idx_series[0]
+    ranges = []
+    for idx in idx_series[1:]:
+        if (idx - last_idx) > 1:
+            ranges.append((start_idx, last_idx))
+            start_idx = idx
```
tamargrey, line 31: this block is also uncovered. If it's possible to adda test case, lets!

woodwork/statistics_utils/frequency_inference/_determine_most_likely_freq.py
```diff
@@ -0,0 +1,20 @@
...
+    alias_info = alias_dict.values()
+
+    n_total = sum([x["count"] for x in alias_info])
+    sorted_aliases = sorted(alias_info, key=lambda item: item["count"], reverse=True)
+
+    most_likely_alias = sorted_aliases[0]["alias"]
+    most_likely_count = sorted_aliases[0]["count"]
+
+    if most_likely_count / n_total < threshold:
+        return None
```
tamargrey, line 15: Also uncovered--let's add a test if we can

tamargrey (commented):
I want to try and get the edge cases covered since that'll help anyone changing the frequency inference logic in the future not miss those edge cases.

woodwork/tests/fixtures/datetime_freq.py
```diff
@@ -0,0 +1,528 @@
...
+    case1(),
+    case2(),
+    case3(),
+    case4(),
+    case5(),
+    case6(),
+    case7(),
+    case8(),
+    case9(),
+    case10(),
```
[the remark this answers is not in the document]
dvreed77: I named them this way cause they come in parameterized this way. The input variable is named case, so if case10 fails, I know which function to immediately go to.

woodwork/statistics_utils/frequency_inference/_determine_most_likely_freq.py
```diff
@@ -0,0 +1,20 @@
...
+
+def _determine_most_likely_freq(alias_dict, threshold=FREQ_INFERENCE_THRESHOLD):
+
+    alias_info = alias_dict.values()
+
+    n_total = sum([x["count"] for x in alias_info])
+    sorted_aliases = sorted(alias_info, key=lambda item: item["count"], reverse=True)
+
+    most_likely_alias = sorted_aliases[0]["alias"]
+    most_likely_count = sorted_aliases[0]["count"]
```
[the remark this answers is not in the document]
dvreed77: I don't think its possible for this to be empty. I can add a check here, but not sure how I would test it

ParthivNaresh (approved):
Looks great, thanks for working on this and creating such a well organized series of functions to deal with the intricacies of various datetime issues. As Tamar noted, a v2 of this focusing on areas of optimization will be made much easier thanks to the clarity of this work.

eccabay (approved):
This looks great, thanks for getting it done!

tamargrey (approved):
Looks good to me! Let's get the latest changes from main and once those pass, 🚀

PR #1285, opened 2022-02-04
Add validation to URL logical type

### LINKED ISSUE #1081: Add validation to URL Logical Type
- As a user, I wish I could use Woodwork with to set columns as URL Logical Type, and have Woodwork validate my data (make sure the column(s) are actually validate URLs).
- This validation would make sure that all non-nan values are actually URLs.

jeff-hernandez:
- Closes #1081

### DIFF
--- a/woodwork/logical_types.py
+++ b/woodwork/logical_types.py
@@ -581,6 +581,43 @@ class URL(LogicalType):
 
     primary_dtype = "string"
 
+    def validate(self, series, return_invalid_values=False):
+        """Validates URL values based on the regex in the config.
+
+        Args:
+            series (Series): Series of URL values
+            return_invalid_values (bool): Whether or not to return invalid URLs
+
+        Returns:
+            Series: If return_invalid_values is True, returns invalid URLs.
+        """
+        regex = config.get_option("url_inference_regex")
+
+        if _is_koalas_series(series):
+
+            def match(x):
+                if isinstance(x, str):
+                    return bool(re.match(regex, x))
+
+            invalid = ~series.apply(match).astype("boolean")
+
+        else:
+            invalid = ~series.str.match(regex).astype("boolean")
+
+        if return_invalid_values:
+            return series[invalid]
+
+        else:
+            any_invalid = invalid.any()
+            if dd and isinstance(any_invalid, dd.core.Scalar):
+                any_invalid = any_invalid.compute()
+
+            if any_invalid:
+                info = f"Series {series.name} contains invalid URL values. "
+                info += "The URL regex can be changed "
+                info += "in the config if needed."
+                raise TypeValidationError(info)
+
 
 class PostalCode(LogicalType):
     """Represents Logical Types that contain a series of postal codes for
--- a/woodwork/tests/accessor/test_table_accessor.py
+++ b/woodwork/tests/accessor/test_table_accessor.py
@@ -2980,32 +2980,49 @@ def test_nan_index_error(sample_df_pandas):
 
 
 def test_validate_logical_types(sample_df):
-    df = sample_df[["email", "age"]]
+    df = sample_df[["email", "url"]]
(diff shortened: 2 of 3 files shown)


docs/source/release_notes.rst
```diff
@@ -6,6 +6,7 @@ Release Notes
 Future Release
 ==============
     * Enhancements
+        * Add validation to URL logical type (:pr:`1285`)
```
thehomebrewnerd, line 9: Not related to this PR: I think we missed this last time, but the entry for validating email address should be in this Future Release section, and not the v0.12.0 release notes section, correct?

If so, can you please fix that?

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -2980,32 +2980,49 @@ def test_nan_index_error(sample_df_pandas):
 
 
 def test_validate_logical_types(sample_df):
-    df = sample_df[["email", "age"]]
+    df = sample_df[["email", "url", "age"]]
```
thehomebrewnerd, line 2983: I think going forward we may want to change our approach here slightly. As we add validation to more logical types, this test is going to get messy. I'm fine with leaving this test as it is for now since it tests that we get data back for multiple series as we expect.

In the future, instead of continuing to modify this test, we might want to add a simpler test just to verify that the proper `LogicalType.validate()` functions get called when we call `validate_logical_types` from the table accessor, without checking the actual values we get back. We can test that the validate functions have the proper behavior with the targeted tests for each specific logical type.

tamargrey:
The URL validate method has a similar enough structure to the EmailAddress one, that I wonder if we could pull out the logic into a helper util that can be used for any validate method that relies on a regex.

docs/source/release_notes.rst
jeff-hernandez: fixed

woodwork/tests/accessor/test_table_accessor.py
jeff-hernandez: sure I don't mind updating the test

jeff-hernandez: [0f5cb30](https://github.com/alteryx/woodwork/pull/1285/commits/0f5cb30d881e51fe8364b2995cbb020d43bf1b7d)

woodwork/logical_types.py
```diff
@@ -617,3 +604,36 @@ class PostalCode(LogicalType):
...
 }
+
+
+def _regex_validate(logical_type, series, return_invalid_values):
+    """Validates data values based on the logical type regex in the config."""
+    regex_key = {
+        "url": "url_inference_regex",
+        "email_address": "email_inference_regex",
+    }[logical_type.type_string]
+    regex = config.get_option(regex_key)
```
thehomebrewnerd, line 615: I like the use of this `_regex_validate` function to clean up the individual logical type methods. 

I'm thinking we may want to keep the logic for getting the proper regex from the config inside the logical type, rather than putting logical type specific logic inside this function. That way, all the logical type specific stuff stays with the logical type and isn't split between the logical type method and this function.

Thoughts?

thehomebrewnerd (commented):
Looking pretty good, just a couple additional minor suggestions.

docs/source/release_notes.rst
```diff
@@ -6,6 +6,8 @@ Release Notes
 Future Release
 ==============
     * Enhancements
+        * Validate email address logical types (:pr:`1247`)
```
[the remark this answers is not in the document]
jeff-hernandez: sure updated

woodwork/logical_types.py
jeff-hernandez: sure I don't mind splitting

thehomebrewnerd (approved):
LGTM!

PR #1287, opened 2022-02-07
Handle large range in extra stats

tamargrey:
- There are situations where the range between a column's min and max value is so large that `len(range(min, max))` produces an OverflowError, so we're changing the call in describe_dict's extra stats to just subtract `max + 1 - min`.


dvreed77 (approved):
LGTM

PR #1289, opened 2022-02-08
Add validation to Age, AgeFractional, and AgeNullable

### LINKED ISSUE #877: Add validation to Age and AgeNullable
- As a user, I wish I could use Woodwork with Age Logical Types, and validate my data.
- If my column contains negative Integer values, I wish that Woodwork throws an error and/or corrects the values for me (by changing the negative integers to np.nan).
- This would apply to Age and AgeNullable

jeff-hernandez:
Closes #877

### DIFF
--- a/woodwork/logical_types.py
+++ b/woodwork/logical_types.py
@@ -102,6 +102,15 @@ class Age(LogicalType):
     standard_tags = {"numeric"}
 
     def validate(self, series, return_invalid_values=True):
+        """Validates age values by checking for non-negative values.
+
+        Args:
+            series (Series): Series of age values
+            return_invalid_values (bool): Whether or not to return invalid age values
+
+        Returns:
+            Series: If return_invalid_values is True, returns invalid age values.
+        """
         return _validate_not_negative(series, return_invalid_values)
 
 
@@ -135,6 +144,15 @@ class AgeNullable(LogicalType):
     standard_tags = {"numeric"}
 
     def validate(self, series, return_invalid_values=True):
+        """Validates age values by checking for non-negative values.
+
+        Args:
+            series (Series): Series of age values
+            return_invalid_values (bool): Whether or not to return invalid age values
+
+        Returns:
+            Series: If return_invalid_values is True, returns invalid age values.
+        """
         return _validate_not_negative(series, return_invalid_values)
 
 
--- a/woodwork/logical_types.py
+++ b/woodwork/logical_types.py
@@ -128,6 +128,18 @@ class AgeFractional(LogicalType):
     primary_dtype = "float64"
     standard_tags = {"numeric"}
 
+    def validate(self, series, return_invalid_values=True):
+        """Validates age values by checking for non-negative values.
+
+        Args:
+            series (Series): Series of age values
+            return_invalid_values (bool): Whether or not to return invalid age values
+
+        Returns:
+            Series: If return_invalid_values is True, returns invalid age values.
+        """
+        return _validate_not_negative(series, return_invalid_values)
+
 
 class AgeNullable(LogicalType):
     """Represents Logical Types that contain whole numbers indicating a person's age.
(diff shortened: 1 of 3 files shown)


jeff-hernandez:
I think we can also include `AgeFractional` in this PR

woodwork/logical_types.py
```diff
@@ -638,3 +644,19 @@ def match(x):
...
+        return series[invalid]
+
+    else:
+        any_invalid = invalid.any()
+        if dd and isinstance(any_invalid, dd.core.Scalar):
+            any_invalid = any_invalid.compute()
+
+        if any_invalid:
+            info = f"Series {series.name} contains negative values."
+            raise TypeValidationError(info)
```
jeff-hernandez, line 662: If necessary, we can make the validation error message more specific to the logical type by making the message a parameter in the function

woodwork/tests/logical_types/test_logical_types.py
```diff
@@ -17,6 +19,7 @@
 from woodwork.tests.testing_utils.table_utils import to_pandas
 from woodwork.utils import import_or_none
 
+dd = import_or_none("dask.dataframe")
```
thehomebrewnerd, line 22: Do we need this? Doesn't seem like it is used in the other new code that was added?

thehomebrewnerd (commented):
Just one small comment.

Also, seems like we should also add validation for `AgeFractional` while we are at it, as you mentioned.

woodwork/tests/logical_types/test_logical_types.py
```diff
@@ -243,3 +246,25 @@ def test_url_validate(sample_df):
...
+
+
+@pytest.mark.parametrize(
+    argnames="logical_type",
+    ids=["age", "age_fractional", "age_nullable"],
+    argvalues=[Age(), AgeFractional(), AgeNullable()],
+)
+def test_age_validate(sample_df, logical_type):
+    invalid_row = pd.Series({4: -3}, name="age", dtype="Int64")
+    series = sample_df["age"]
```
thehomebrewnerd, line 258: I think for testing `AgeFractional` we should pass in a series with a `float64` dtype, since that is what a column idetified as `AgeFractional` will be stored as. Might not matter much, but I think that would be more representative of what would actually happen when validating a dataframe that has that type.

woodwork/tests/logical_types/test_logical_types.py
```diff
@@ -243,3 +246,25 @@ def test_url_validate(sample_df):
     actual = logical_type.validate(series, return_invalid_values=True)
     expected = pd.Series({4: "bad_url"}, name="url")
     assert to_pandas(actual).equals(expected)
+
+
+@pytest.mark.parametrize(
+    argnames="logical_type",
+    ids=["age", "age_fractional", "age_nullable"],
```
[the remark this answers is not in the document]
jeff-hernandez: yeah it helps to identify which paramtrized test is running during pytest

woodwork/tests/logical_types/test_logical_types.py
jeff-hernandez: sure, it won't affect validation behavior but I don't mind changing

woodwork/tests/logical_types/test_logical_types.py
```diff
@@ -243,3 +246,29 @@ def test_url_validate(sample_df):
...
+def test_age_validate(sample_df, logical_type):
+    invalid_row = pd.Series({4: -3}, name="age", dtype="Int64")
+    series = sample_df["age"]
+
+    if _is_koalas_series(series):
+        invalid_row = ks.from_pandas(invalid_row)
+    assert logical_type.validate(series, return_invalid_values=False) is None
+
+    series = series.append(invalid_row).astype("Int64")
+    if isinstance(logical_type, AgeFractional):
```
thehomebrewnerd, line 265: With the change in this spot we aren't testing the first assertion with a `float64` dtype.

jeff-hernandez: updated

thehomebrewnerd (approved):
I like the cleanup! LGTM!

PR #1299, opened 2022-02-11
Fixes issue with initializing Woodwork Series with LatLong values

### LINKED ISSUE #1298: Initialize Series with tuple of nans fails
Initializing a series with a tuple of nans fails.

#### Code Sample

```python
ww.init_series(pd.Series([(1.0, 2.0), (float('nan'), float('nan'))]), logical_type='LatLong')
```

Returns this error
```
ValueError: Cannot initialize Woodwork. Series does not contain properly formatted LatLong data. Try reformatting before initializing or use the woodwork.init_series function to initialize.
```

dvreed77:
- Fixes issue with initializing Woodwork Series with LatLong values
- Closes #1298


docs/source/release_notes.rst
```diff
@@ -14,6 +14,7 @@ Future Release
         * Add validation to Age, AgeFractional, and AgeNullable logical types (:pr:`1289`)
     * Fixes
         * Check range length in table stats without producing overflow error (:pr:`1287`)
+        * Fixes issue with initializing Woodwork Series with LatLong values (:pr:`1299`)
```
thehomebrewnerd, line 17: Add your username to the list of contributors below.

woodwork/tests/utils/test_accessor_utils.py
```diff
@@ -101,6 +101,14 @@ def test_init_series_with_datetime(sample_datetime_series):
     assert isinstance(series.ww.logical_type, Datetime)
 
 
+def test_init_series_with_latlong(latlong_df_pandas):
```
thehomebrewnerd, line 104: Is there a reason for testing this with only the pandas version of the fixture and not testing with Dask & Koalas?

woodwork/utils.py
```diff
@@ -216,6 +216,10 @@ def _is_url(string):
 
 def _reformat_to_latlong(latlong, use_list=False):
     """Reformats LatLong columns to be tuples of floats. Uses (np.nan, np.nan) for null values."""
```
thehomebrewnerd, line 218: The docstring for this needs to be updated now that we can also have a single `nan` value. Probably should also review all the other latlong related docstrings to make sure they are consistent with the new behavior as well.

woodwork/tests/utils/test_utils.py
```diff
@@ -282,7 +282,7 @@ def test_reformat_to_latlong():
     assert _reformat_to_latlong("[1, 2]") == simple_latlong
     assert _reformat_to_latlong("1, 2") == simple_latlong
 
-    assert _reformat_to_latlong(None) == (np.nan, np.nan)
+    assert np.isnan(_reformat_to_latlong(None))
```
thehomebrewnerd, line 285: Do we need to test any more input types here - like these for example:
```
_reformat_to_latlong(np.nan)
_reformat_to_latlong(pd.NA)
_reformat_to_latlong((np.nan, np.nan))
_reformat_to_latlong((pd.NA, pd.NA))
_reformat_to_latlong((None, None))
```

woodwork/tests/utils/test_utils.py
```diff
@@ -345,21 +345,20 @@ def test_is_nan():
...
+@pytest.mark.parametrize(
+    "test_input,expected",
+    [
+        ((1.0, 2.0), True),
+        (np.nan, True),
+        ([1.0, 2.0], False),
+        ((np.nan, np.nan), True),
+        (("a", 2.0), False),
+        ((1.0, 2.0, 3.0), False),
+        (None, False),
```
thehomebrewnerd, line 357: Since we are generally supporting nullable columns types, may be we should test with some `pd.NA` inputs as well and make sure those return False?

docs/source/release_notes.rst
gsheni: done

woodwork/tests/utils/test_utils.py
```diff
@@ -243,87 +247,107 @@ def test_is_s3():
...
+        ((None, None), (np.nan, np.nan)),
+        (None, np.nan),
+        (np.nan, np.nan),
+        (pd.NA, np.nan),
+        ("None", np.nan),
+        ("NaN", np.nan),
+        ("<NA>", np.nan),
+    ],
+)
+def test_reformat_to_latlong(test_input, expected):
```
rwedge, line 295: if the only difference between `test_reformat_to_latlong` and `test_reformat_to_latlong_koalas`, could parametrize `is_koalas` True/False and cast from list to tuple or vice versa if expected needs to be a list and is a tuple

woodwork/exceptions.py
```diff
@@ -78,3 +78,24 @@ class WoodworkNotInitWarning(UserWarning):
 
 class TypeValidationError(Exception):
     pass
+
+
+class LatLongIsNotTupleError(ValueError):
+    def __init__(self, value):
+        return super().__init__(
+            f"LatLongs must either be a tuple, a list, or a string representation of a tuple. {value} does not fit the criteria."
```
rwedge, line 86: This error message does not mention a NaN value is ok.

woodwork/utils.py
```diff
@@ -258,36 +274,44 @@ def _to_latlong_float(val):
...
 
 
-def _is_valid_latlong_value(val, bracket_type=tuple):
+def _is_valid_latlong_value(val, is_koalas=False):
     """Returns True if the value provided is a properly formatted LatLong value for a
     pandas, Dask or Koalas Series, otherwise returns False."""
-    if isinstance(val, bracket_type) and len(val) == 2:
+    if isinstance(val, (list, tuple)) and len(val) == 2:
+        if len(val) != 2:
+            return False
```
rwedge, line 301: can this be hit?

woodwork/exceptions.py
```diff
@@ -78,3 +78,29 @@ class WoodworkNotInitWarning(UserWarning):
 
 class TypeValidationError(Exception):
     pass
+
+
+class LatLongIsNotTupleError(ValueError):
```
thehomebrewnerd, line 83: Do we want/need separate error classes for each of these, or should we use a single class and customize the error based on the specific failure? Also, is this distinct from the existing `TypeValidationError` or could we use that instead? If we don't use `TypeValidationError` should these be a subclass of that?

I'm not sure what the right answer is to any of those questions, just raising it for discussion.

thehomebrewnerd (approved):
This looks good to me, but might not hurt to get a second opinion before you merge this in.

tamargrey (approved):
lgtm!

PR #1322, opened 2022-03-03
Calculate number of unique for Unknown columns

### LINKED ISSUE #1257: Calculate nunique for Unknown columns in `_get_describe_dict`
- As a user, I wish I could use Woodwork's describe method to inspect the uniqueness of Unknown columns.

This would entail adding a field to `agg_stats_to_calculate`

```python
 Unknown: [ "nunique"],
```

mingdavidqi:
- Calculate nunique for Unknown columns in `_get_describe_dict`
- Closes #1257


woodwork/column_schema.py
```diff
@@ -152,6 +152,11 @@ def is_boolean(self):
         ltype_class = type(self.logical_type)
         return ltype_class == Boolean or ltype_class == BooleanNullable
 
+    @property
+    def is_unknown(self):
+        """Whether the ColumnSchema is a Datetime column"""
```
tamargrey, line 157: docstring needs updating!

woodwork/statistics_utils/_get_describe_dict.py
```diff
@@ -62,6 +62,7 @@ def _get_describe_dict(
         "category": ["count", "nunique"],
         "numeric": ["count", "max", "min", "nunique", "mean", "std"],
         Datetime: ["count", "max", "min", "nunique", "mean"],
+        Unknown: ["nunique"],
```
tamargrey, line 65: Need to add `"count "` or Unknown columns will no longer get count calculated for them (see line 122 where all other columns get `"count"` calculated for them

woodwork/column_schema.py
mingdavidqi: updated

woodwork/statistics_utils/_get_describe_dict.py
mingdavidqi: added

tamargrey:
I don't think we currently test `describe` on any Unknown columns, which is why these changes haven't broken any tests. Can we add a test that explicitly sets a logical type as Unknown before running `describe` so we can confirm that both `count` and `nunique` are present?

tamargrey (approved):
lgtm!

PR #1325, opened 2022-03-04
Refactor serialization and deserialization

### LINKED ISSUE #1307: Refactor Serialization/Deserialization for easier updates and better user control of process
The serialization and deserialization process should be refactored to achieve better separation between the serialization approaches. Currently the serialization logic is spread over multiple functions and changing the serialization approach for one format is difficult without breaking serialization for another format.

We should create individual serializer classes (`CSVSerializer`, `ParquetSerializer`, etc) that can be updated independently of one another. Common logical can be stored on a parent `Serializer` class. This may also be accompanied by corresponding `Deserializer` classes. 

Finally, as part of the refactor, users should be given more control over the names of the files used to serialize both the data and the typing information.

This default behavior after the refactor should match the current serialization behavior and the current serialization tests should remain unmodified, although new tests may be needed.

thehomebrewnerd:
- Refactors serialization and deserialzation
- Closes #1307 

Implements `Serializer` and `Deserializer` classes to use a more modular approach to serializing and deserializing, allowing for easier customization for specific file formats in the future.


woodwork/serialize.py
```diff
@@ -1,225 +1,4 @@
...
-    from pyarrow import Table, orc
-
-    df = dataframe.copy()
-    for c in df:
-        if df[c].dtype.name == "category":
-            df[c] = df[c].astype("string")
-    pa_table = Table.from_pandas(df, preserve_index=False)
-    orc.write_table(pa_table, filepath)
+# This is included to prevent a breaking API change
+from woodwork.serializers.serializer_base import typing_info_to_dict
```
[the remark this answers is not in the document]
thehomebrewnerd: Not sure - perhaps. I left this in for now, as Featuretools uses this function, so moving it would be a breaking change for Featuretools. However, if we are going to be limiting the min Woodwork version in Featuretools as a result of the pyspark pandas API changes anyway, we might be able to go ahead and remove this now.

docs/source/guides/saving_and_loading_dataframes.ipynb
```diff
@@ -114,7 +119,9 @@
     "\n",
     "## Loading a Woodwork DataFrame\n",
     "\n",
-    "After saving a Woodwork DataFrame, you can load the DataFrame and typing information by using `woodwork.deserialize.read_woodwork_table`. This function will use the stored typing information in the specified directory to recreate the Woodwork DataFrame."
+    "After saving a Woodwork DataFrame, you can load the DataFrame and typing information by using `woodwork.deserialize.read_woodwork_table`. This function will use the stored typing information in the specified directory to recreate the Woodwork DataFrame. \n",
+    "\n",
+    "If you have modified any of the default values for the filename, data subdirectory or typing information file, you will need to specify those when calling `read_woodwork_table`. Since we did not change any of the defaults for this example, we do not need to speify them here."
```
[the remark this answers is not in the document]
thehomebrewnerd: Fixed

woodwork/serialize.py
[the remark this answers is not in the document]
jeff-hernandez: We'll likely increase the min Woodwork version in Featuretools, so we might as well remove this file.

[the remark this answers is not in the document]
thehomebrewnerd: Sounds good to me

woodwork/serializers/serializer_base.py
```diff
@@ -0,0 +1,203 @@
...
+import tarfile
+import tempfile
+
+from woodwork.accessor_utils import _is_dask_dataframe, _is_koalas_dataframe
+from woodwork.logical_types import LatLong
+from woodwork.s3_utils import get_transport_params, use_smartopen
+from woodwork.type_sys.utils import _get_ltype_class, _get_specified_ltype_params
+from woodwork.utils import _is_s3, _is_url
+
+SCHEMA_VERSION = "11.3.0"
```
jeff-hernandez, line 13: Do we want to increase the version as a result of adding custom names for the data subdirectory and typing info?

woodwork/serializers/serializer_base.py
```diff
@@ -0,0 +1,203 @@
...
+        self.typing_info_filename = typing_info_filename
+        self.dataframe = None
+        self.typing_info = None
+        self.location = None
+        self.kwargs = {}
+
+    def serialize(self, dataframe, profile_name, **kwargs):
+        """Serialize data and typing information to disk."""
+        self.dataframe = dataframe
+        self.profile_name = profile_name
```
jeff-hernandez, line 39: It seems like `profile_name` only applies inside `save_to_s3`. Is there a reason we need to keep this outside of `save_to_s3`?

jeff-hernandez (commented):
Overall looks fine, just had a couple of questions.

woodwork/serializers/serializer_base.py
thehomebrewnerd: Good question. I debated about that a little myself. Ultimately, I landed on leaving it the same since we aren't changing anything in the JSON data use to store the typing information. Also, older versions of WW will be able to deserialize data serialized with this version when the default values for filename, data subdirectory and typing info file are used.

Not exactly sure what's best here though.

jeff-hernandez: Yeah, I'm not entirely sure either. There's also the case of saving the data and typing info using custom names with this version and not being able to load using previous versions.

thehomebrewnerd: Yeah, we would lose the ability to deserialize with older versions in cases where the defaults are changed.

@rwedge @tamargrey Any thoughts on what is appropriate here? Should we keep the SCHEMA_VERSION the same or update with the changes in this PR?

tamargrey: The goal here is that this wouldn't be a breaking change, right? If previously serialized dataframes (with schema 11.3.0) can still be deserialized now, and older versions of Woodwork can deserialize from this version of Woodwork with the default file names, that feels like we should keep the schema the same.

thehomebrewnerd: Yeah, the idea here was that the same serialize and deserialize API calls that worked before would still work after this update and achieve the same result. I think that was accomplished at the level of `df.ww.to_disk` and `read_woodwork_table`. There were some breaking API changes in terms of the other utility functions used in the process that have now been removed from the public API.

The main wrinkle is that this PR implements additional capability that is not backward compatible.

woodwork/serializers/serializer_base.py
thehomebrewnerd: Was not needed. Updated in [7cbba99](https://github.com/alteryx/woodwork/pull/1325/commits/7cbba99525ca9f0701caa6e2fe5d025c9fb0ed2b)

jeff-hernandez: let's also do the same for deserializing `read_from_s3`

thehomebrewnerd: Good call. Done [6f5a6cc](https://github.com/alteryx/woodwork/pull/1325/commits/6f5a6ccef2f97b719e800e18141cc2ade51a88aa)

jeff-hernandez (approved):
LGTM

PR #1332, opened 2022-03-09
Upgrade moto requirement

### LINKED ISSUE #1328: Remove pinned minimum responses version when moto can handle responses 0.19.0
Currently there is an [error in moto](https://github.com/spulec/moto/issues/4913) with the latest version of responses (0.19.0). This caused our minimum dependency tests to break, so we've limited `responses>=0.9.0,<0.19.0` in the `test-requirements.txt`. 

Once there is a moto version that can handle responses 0.19.0, we should be able to remove the pinned minimum version and raise the min moto version in `test-requirements.txt`.

rwedge:
Fixes #1328

Only the newest `moto` release resolves the issues with the latest `responses` version so I bumped the minimum `moto` version to the newest one


thehomebrewnerd (approved):
LGTM

gsheni (approved):

PR #1340, opened 2022-03-15
Update __str__ output for Datetime and Ordinal logical types

### LINKED ISSUE #226: Add format to __str__ of Datetime Logical Type
- If a Datetime Logical type has a format, this should be printed out in the __str__ representation

### LINKED ISSUE #227: Add ranking to __str__ of Ordinal
- With the Ordinal Logical Type, we have/will have a ranking attribute that stores the order
- This should be printed in the __str__

bchen1116:
- Update the logical string types for Ordinal logical types
- Closes #227

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,18 +65,18 @@ df.ww
 ```
 
 ```
-                Physical Type                 Logical Type Semantic Tag(s)
+                   Physical Type     Logical Type Semantic Tag(s)
 Column
-order_id             category                  Categorical    ['category']
-product_id           category                  Categorical    ['category']
-description            string              NaturalLanguage              []
-quantity                int64                      Integer     ['numeric']
-order_date     datetime64[ns]  Datetime: %Y-%m-%d %H:%M:%S              []
-unit_price            float64                       Double     ['numeric']
-customer_name          string               PersonFullName              []
-country              category                  Categorical    ['category']
-total                 float64                       Double     ['numeric']
-cancelled                bool                      Boolean              []
+order_id                category      Categorical    ['category']
+product_id              category      Categorical    ['category']
+description               string  NaturalLanguage              []
+quantity                   int64          Integer     ['numeric']
+order_date        datetime64[ns]         Datetime              []
+unit_price               float64           Double     ['numeric']
+customer_name             string   PersonFullName              []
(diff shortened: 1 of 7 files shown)


gsheni:
We might want to update the README.md as well
- https://github.com/alteryx/woodwork#example

woodwork/logical_types.py
```diff
@@ -299,6 +299,9 @@ def transform(self, series):
                     )
         return super().transform(series)
 
+    def __str__(self):
+        return "{}: {}".format(str(self.__class__), self.datetime_format)
```
gsheni, line 303: @jeff-hernandez @thehomebrewnerd Do you think maybe we should not have the datetime format be a part of the __str__? It makes sense for the Ordinal one because it is actually a part of the Logical Type. 
- For Datetime however, we have already changed it to Datetime variables (so the format isn't really useful anymore).

bchen1116: I agree. The current implementation also doesn't do anything special when there is no format passed in (None), but I personally don't think it's necessary for datetime to print out the format

gsheni: Let's remove it for Datetime.

jeff-hernandez: yep not necessary

jeff-hernandez (approved):

README.md
```diff
@@ -65,18 +65,18 @@ df.ww
...
-unit_price               float64           Double     ['numeric']
-customer_name             string   PersonFullName              []
-country                 category      Categorical    ['category']
-total                    float64           Double     ['numeric']
-cancelled                   bool          Boolean              []
+order_id             category                  Categorical    ['category']
+product_id           category                  Categorical    ['category']
+description            string              NaturalLanguage              []
+quantity                int64                      Integer     ['numeric']
+order_date     datetime64[ns]  Datetime: %Y-%m-%d %H:%M:%S              []
```
gsheni, line 74:
```suggestion
order_date     datetime64[ns]              Datetime              []
```

gsheni, line 74:
```suggestion
quantity                int64                      Integer     ['numeric']
order_date     datetime64[ns]                     Datetime              []
```

gsheni (approved):
lgtm

PR #1351, opened 2022-03-21
Conditionally run scheduled workflows

### LINKED ISSUE #1349: Prevent workflows from running on forked repos
Our scheduled workflows such as Auto Approve Dependency PRs and Latest Dependency Checker should not run on forked repos. 
We should make it so the workflow only runs if the repo owner `(if: github.owner == alteryx)`.

thehomebrewnerd:
### Conditionally run scheduled workflows

Closes #1349 

Update scheduled workflows to run only if repository_owner is alteryx. This should prevent these scheduled workflows from running on repo forks and failing repeatedly.

### DIFF
--- a/.github/workflows/auto_approve_dependency_PRs.yml
+++ b/.github/workflows/auto_approve_dependency_PRs.yml
@@ -12,6 +12,7 @@ on:
       - completed
 jobs:
   build:
+    if: ${{ github.repository_owner == 'alteryx' }}
     runs-on: ubuntu-latest
     steps:
       - name: Find dependency PRs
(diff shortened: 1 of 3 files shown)


gsheni (approved):

.github/workflows/auto_approve_dependency_PRs.yml
```diff
@@ -12,6 +12,7 @@ on:
       - completed
 jobs:
   build:
+    if: ${{ github.repository_owner == 'alteryx' }}
```
jeff-hernandez, line 15: The outer wrapper `${{ }}` is not necessary

gsheni, line 15:
```suggestion
    if: github.repository_owner == 'alteryx'
```

gsheni: fixed

PR #1353, opened 2022-03-23
Upgrade minimum pandas to 1.4.0

gsheni:
- Fixes #1353 1294

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -13,12 +13,13 @@ Future Release
     * Changes
         * Updated lint check to only run on Python 3.10 (:pr:`1345`)
         * Transition to use pyproject.toml and setup.cfg (moving away from setup.py) (:pr:`1346`)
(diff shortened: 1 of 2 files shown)


ParthivNaresh (commented):
Does `spark = pandas < 1.4.0` have to be [updated](https://github.com/alteryx/woodwork/blob/main/setup.cfg#L78)?

jeff-hernandez:
@gsheni  pandas 1.4.0 introduced breaking changes to woodwork. Whenever we support 1.4.0 in woodwork, we can also try to upgrade for spark.

thehomebrewnerd:
> @gsheni pandas 1.4.0 introduced breaking changes to woodwork. Whenever we support 1.4.0 in woodwork, we can also try to upgrade for spark.

I don't think we are restricting the pandas version unless pyspark is installed, are we? I think the issues were Koalas related before, and those may have carried over into the pyspark pandas api implementation as well. I'm not sure we ever fully investigated to document the specific issues that caused us to add the restriction.

PR #1360, opened 2022-03-24
Change string to string[pyarrow]

### LINKED ISSUE #1318: Use string[arrow] dtype for all Logical Types that use string dtype
- pandas 1.3 has a new string[arrow] dtype that saves on memory and improves speed
  - https://pythonspeed.com/articles/pandas-string-dtype-memory/
- We should use it, and verify that all subsequent calls on this work (Featuretools, EvalML).

bchen1116:
fix #1318 

Since this feature is still experimental, I went with a less strict implementation. This means that `string` and `string[pyarrow]` will be treated as the same type when doing validation, but internally, we will use the `string[pyarrow]` type.

Confirmed locally that this is compatible with Featuretools and EvalML.


setup.cfg
```diff
@@ -48,7 +48,7 @@ dev =
     codecov == 2.1.12
     flake8 == 3.9.2
     isort == 5.9.3
-    black == 21.9b0
+    black == 22.3.0
```
gsheni, line 51: Do we need to require `pyarrow` for Woodwork with the new string[arrow] dtype?

woodwork/tests/conftest.py
```diff
@@ -50,24 +50,33 @@ def sample_df_pandas():
...
-                "team@featuretools.com",
-                "junk@example.com",
-            ],
-            "phone_number": [
-                "5555555555",
-                "555-555-5555",
-                "1-(555)-555-5555",
-                "555-555-5555",
-            ],
+            "full_name": pd.Series(
```
thehomebrewnerd, line 53: Just curious why you are specifying the dtype for these columns now. Did things break otherwise?

bchen1116: Great point! I changed this during testing earlier, but it works without the casting! Removing now.

setup.cfg
```diff
@@ -40,6 +40,7 @@ include_package_data = True
 install_requires =
     pandas >= 1.3.0
     scikit-learn >= 0.22
+    pyarrow >= 1.0.0
```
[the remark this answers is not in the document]
gsheni: We will need to update this as well
https://woodwork.alteryx.com/en/latest/install.html#optional-python-dependencies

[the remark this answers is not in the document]
bchen1116: `pyarrow` is already on the dependencies list!

gsheni:
Does this pass with the latest in Featuretools and EvalML?

setup.cfg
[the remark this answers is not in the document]
ParthivNaresh: Did this change from `oyarrow == 1.0.1` the docs here?
https://pandas.pydata.org/pandas-docs/version/1.4.1/whatsnew/v1.4.0.html#multi-threaded-csv-reading-with-a-new-csv-engine-based-on-pyarrow

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -3035,15 +3035,21 @@ def test_validate_logical_types(sample_df):
...
-            "url": "string",
-            "email": "string",
-            "email_2": "string",
+            "url": "string[pyarrow]",
+            "email": "string[pyarrow]",
+            "email_2": "string[pyarrow]",
         }
     )
-
+    if _is_spark_dataframe(df):
```
ParthivNaresh, line 3045: Because `pyspark` doesn't support this yet right?

woodwork/logical_types.py
```diff
@@ -63,7 +63,7 @@ def _get_valid_dtype(cls, series_type):
     def transform(self, series):
         """Converts the series dtype to match the logical type's if it is different."""
         new_dtype = self._get_valid_dtype(type(series))
-        if new_dtype != str(series.dtype):
+        if new_dtype != str(series.dtype) and new_dtype != series.dtype:
```
ParthivNaresh, line 66: What's the purpose of this distinction?

woodwork/logical_types.py
```diff
@@ -216,7 +218,7 @@ class Categorical(LogicalType):
     """
 
     primary_dtype = "category"
-    backup_dtype = "string"
+    backup_dtype = "string[pyarrow]"
```
ParthivNaresh, line 221: Does setting the `backup_dtype` to `string[pyarrow]` here and in other places cause an issue with `_get_valid_dtype` in the `LogicalType` parent class if a series is determined to be from pyspark? Since the `backup _type` would get returned? I'm sure I'm missing something in the logic of how it works

ParthivNaresh (commented):
Looks good, just left some comments to get a better understanding of the implementation!

setup.cfg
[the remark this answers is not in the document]
bchen1116: I think the `1.0.1` update allowed `read_csv` to use pyarrow internally, but we only need `1.0.0` to be able to use `string[pyarrow]` as the backend representation

woodwork/logical_types.py
bchen1116: Pyspark does support `string[pyarrow]` dtypes, so this is fine!

woodwork/tests/accessor/test_statistics.py
```diff
@@ -962,15 +963,16 @@ def test_value_counts(categorical_df):
                 {k: (str(v) if k == "value" else v) for k, v in items.items()}
             )
         expected_cat1 = updated_results
+        nan = pd.NA
```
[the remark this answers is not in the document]
bchen1116: See comment below!

woodwork/column_accessor.py
```diff
@@ -446,7 +450,9 @@ def _validate_schema(schema, series):
         raise TypeError("Provided schema must be a Woodwork.ColumnSchema object.")
 
     valid_dtype = schema.logical_type._get_valid_dtype(type(series))
-    if str(series.dtype) != valid_dtype:
+    if str(series.dtype) != valid_dtype and not (
```
Cmancuso, line 453: nit, but we might want to consider just making a convenience method for this.

woodwork/tests/accessor/test_table_accessor.py
Cmancuso: is there any concern that `validate_logical_types()` returns a different type?

thehomebrewnerd: @Cmancuso @bchen1116 I'm a little but confused/concerned with this. Was going to try to take a closer look to see if I could understand why this is the case, but haven't had a chance yet.

bchen1116: I can also try to look into it when I get back next week! I tried yesterday/today, but didn't end up having time to

woodwork/column_accessor.py
bchen1116: Will do!

woodwork/tests/accessor/test_table_accessor.py
thehomebrewnerd: Looked into this a little bit - the dtype change is happening during the concat operation here (coming from pyspark): https://github.com/alteryx/woodwork/blob/6332317c111993be4ad7f12686a077dae4341edf/woodwork/table_accessor.py#L1148

The series dtypes are all `string[pyarrow]` up until that point for pandas, Dask and Spark dataframes. For some reason the spark concat operation changes the dtype, while the pandas and Dask concat operations don't (maybe a spark bug/issue?)

However, looking at this test a bit more closely I don't think we care about the dtypes of `actual`. In this case `actual` is just a DataFrame that contains the invalid values. This DataFrame does not have Woodwork initialized and as such we shouldn't really have any expectation for getting certain dtypes back in that df as long as the values are correct.

I'm thinking we could update this test completely to remove the type casting on expected dataframe and compare `expected` to `actual` with `pd.testing.assert_frame_equal` using `check_dtype=False` since the dtype isn't so critical in this case.

@jeff-hernandez Do you have any strong opinions on this?

thehomebrewnerd (commented):
@bchen1116 A couple more comments on this PR:

- As part of this, you should also scan the documentation and look for references to `string` dtype that should be updated to `string[pyarrow]`
- Since the original intent of making this change was to improve memory usage and speed, it might be good to compare this against the current release to see what was actually achieved with this update. If the gains aren't significant, we might want to reconsider making this change and adding an experimental dtype that could change/break in the future without much warning depending on the direction pandas heads here.

bchen1116:
@thehomebrewnerd I've gone through and updated the documentation in all the parts that reference the regular string type. Additionally, here's a side by side comparison with the `string[python]` (original) and `string[pyarrow]` (new) type on both runtime and memory usage! The resulting changes are fairly substantial, which seems like a great sign to move forward with this!

fyi @gsheni @Cmancuso

Cmancuso:
> @thehomebrewnerd I've gone through and updated the documentation in all the parts that reference the regular string type. Additionally, here's a side by side comparison with the `string[python]` (original, on the left) and `string[pyarrow]` (new, on the right) type on both runtime and memory usage! The resulting changes are fairly substantial, which seems like a great sign to move forward with this!
> 
> image
> 
> fyi @gsheni @Cmancuso

That is a significant performance improvement. Very cool stuff.

Cmancuso (approved):
This looks good to me! Our existing tests should cover any issues with backward compatibility between string types, right? As long as that's true, ⛴

PR #1361, opened 2022-03-24
Allow attr version in setup.cfg

gsheni:


.github/workflows/install_test.yml
```diff
@@ -8,12 +8,13 @@ on:
 name: Install Test
 jobs:
   install_ww_complete:
-    name: ${{ matrix.python_version }} install woodwork complete
-    runs-on: ubuntu-latest
+    name: ${{ matrix.python_version }} - ${{ matrix.python_version }} install woodwork complete
```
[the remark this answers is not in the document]
gsheni: yep. fixed

Makefile
```diff
@@ -28,9 +28,7 @@ testcoverage:
 	pytest woodwork/ --cov=woodwork
 
 .PHONY: installdeps
-installdeps:
-	pip install --upgrade pip
-	pip install -e .
+installdeps: upgradepip
```
jeff-hernandez, line 31: Do we still need `installdeps-dev`?

Makefile
```diff
@@ -28,9 +28,7 @@ testcoverage:
 	pytest woodwork/ --cov=woodwork
 
 .PHONY: installdeps
-installdeps:
-	pip install --upgrade pip
-	pip install -e .
```
ParthivNaresh, line 33: Why was this removed? Shouldn't we expand this to cover
`pip install -e .`, `pip install -e ".[dev]"` and `pip install -e ".[test]"`

Makefile
gsheni: removed

Makefile
gsheni: Additionally, we don't use the test, and dev commands anywhere (CI wise). Devs can just directly install them via `pip`.

jeff-hernandez (approved):

PR #1363, opened 2022-03-28
Adding from_disk convenience function

### LINKED ISSUE #1308: [SPIKE] - Add a new from_disk convenience method, corresponding to existing to_disk method
We currently have a `to_disk` method for serializing a Woodwork table, but the function for deserializing the data is called `read_woodwork_table`. It might make more sense for the deserialization function to be `from_disk` instead. This could either replace the existing method (breaking change) or be a convenience function that just calls the existing `read_woodwork_table` function.

ParthivNaresh:
Closes #1308

### DIFF
--- a/docs/source/api_reference.rst
+++ b/docs/source/api_reference.rst
@@ -129,6 +129,7 @@ Deserialization
 .. autosummary::
     :toctree: generated/
 
+    from_disk
     read_woodwork_table
 
 Logical Types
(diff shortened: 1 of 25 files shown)


thehomebrewnerd (commented):
Generally I think this looks fine, but I'm thinking if we want to keep `read_woodwork_table` as the "primary" method for reading in a serialized table and `from_disk` is just a convenience method, we should leave the serialization tests unchanged so that they all still use `read_woodwork_table`. 

To test `from_disk` you could just create a simple test where you mock `read_woodwork_table` and confirm that `from_disk` calls `read_woodwork_table` with all the right parameters.

Also, we should add the new function to the API reference documentation, and consider updating the serialization guide to mention this function if we want users to be aware of its existence.

ParthivNaresh:
@thehomebrewnerd Will do!

PR #1365, opened 2022-03-29
Improve exception handling in col_is_datetime

### LINKED ISSUE #130: Improve exception handling in col_is_datetime utility function
In the `col_is_datetime` utility function in `utils.py` there is a try/except block that catches `Exception`. This should be reviewed to see if it can be improved by catching more specific exceptions instead.

Relevant code:
```
    try:
        pd.to_numeric(dropped_na, errors='raise')
    except (ValueError, TypeError):
        # finally, try to cast to datetime
        if col.dtype.name.find('str') > -1 or col.dtype.name.find('object') > -1:
            try:
                pd.to_datetime(dropped_na, errors='raise')
            except Exception:
                return False
            else:
                return True
```

For reference:
https://github.com/pandas-dev/pandas/issues/16757
https://github.com/pandas-dev/pandas/blob/v1.1.2/pandas/core/tools/datetimes.py#L605-L832

jeff-hernandez:
Closes #130

### DIFF
--- a/woodwork/tests/utils/test_utils.py
+++ b/woodwork/tests/utils/test_utils.py
@@ -456,6 +456,7 @@ def test_col_is_datetime():
         pd.Series(["a", "b", "c"]),
         pd.Series([pd.NA, "b", "c"]),
         pd.Series([pd.NA, pd.NA, pd.NA]),
+        pd.Series(["12/01/2022", pd.NA, pd.NA]),
     ]
 
     expected_values = [
@@ -470,6 +471,7 @@ def test_col_is_datetime():
         False,
         False,
         False,
+        True,
     ]
 
     for input, expected in list(zip(inputs, expected_values)):
--- a/woodwork/tests/utils/test_utils.py
+++ b/woodwork/tests/utils/test_utils.py
@@ -456,7 +456,6 @@ def test_col_is_datetime():
         pd.Series(["a", "b", "c"]),
         pd.Series([pd.NA, "b", "c"]),
         pd.Series([pd.NA, pd.NA, pd.NA]),
-        pd.Series(["12/01/2022", pd.NA, pd.NA]),
     ]
 
     expected_values = [
@@ -471,7 +470,6 @@ def test_col_is_datetime():
         False,
         False,
         False,
-        True,
     ]
 
     for input, expected in list(zip(inputs, expected_values)):
--- a/woodwork/type_sys/utils.py
+++ b/woodwork/type_sys/utils.py
@@ -1,4 +1,5 @@
 import pandas as pd
+from dateutil.parser import ParserError
 
 import woodwork as ww
 from woodwork.accessor_utils import _is_dask_series, _is_spark_series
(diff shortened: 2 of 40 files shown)


woodwork/type_sys/utils.py
```diff
@@ -16,30 +14,24 @@ def col_is_datetime(col, datetime_format=None):
     if _is_spark_series(col):
         col = col.to_pandas()
 
-    if col.dtype.name.find("datetime") > -1 or (
-        len(col) and isinstance(col.head(1), datetime)
-    ):
+    if pd.api.types.is_datetime64_any_dtype(col):
```
bchen1116, line 17: Nice! I like this change of letting pandas handle datetime detection!

woodwork/type_sys/utils.py
```diff
@@ -16,30 +14,24 @@ def col_is_datetime(col, datetime_format=None):
...
+        pd.to_datetime(
+            col.astype(str),
+            errors="raise",
+            format=datetime_format,
+            infer_datetime_format=True,
+        )
+        return True
 
-    return False
+    except Exception:
```
bchen1116, line 33: I see these are the two exceptions that can be raised: 
image

Are we planning on treating both of these equally and returning False for both scenarios? If so, I think it'd be great to add a small comment on that here for future visibility.

bchen1116 (approved):
LGTM! Left one question, but big fan of the cleaned logic of datetime detection!

woodwork/type_sys/utils.py
Cmancuso: I think in general it's better to not have generic exception handling `except Exception` and we should be specific about the exceptions that we are handling here

Cmancuso (commented):
Can we also add a test for this case?

woodwork/type_sys/utils.py
```diff
@@ -16,30 +14,24 @@ def col_is_datetime(col, datetime_format=None):
...
-                    col,
-                    errors="raise",
-                    format=datetime_format,
-                    infer_datetime_format=True,
-                )
-            except Exception:
-                return False
-            else:
-                return True
+        pd.to_datetime(
```
thehomebrewnerd, line 25: @jeff-hernandez Does this updated logic with the string conversion work as expected if the input series contains missing values like `None`, `np.nan` or `pd.NA`?

Would this series be recognized as a datetime series?
```
series = pd.Series(["12/01/2022", pd.NA])
```

By having the `.astype(str)` call inside the try/except block this could fail because the `pd.to_datetime` conversion fails or because the `.astype(str)` call fails.

woodwork/type_sys/utils.py
jeff-hernandez: Sounds good, updated!

jeff-hernandez: Other specific exceptions in woodwork's use case include overflow and type errors

woodwork/tests/utils/test_utils.py
```diff
@@ -456,6 +456,7 @@ def test_col_is_datetime():
         pd.Series(["a", "b", "c"]),
         pd.Series([pd.NA, "b", "c"]),
         pd.Series([pd.NA, pd.NA, pd.NA]),
+        pd.Series(["12/01/2022", pd.NA, pd.NA]),
```
thehomebrewnerd, line 459: Minor detail but this situation is already covered by line 455, so this can be removed. Didn't realize this was already being tested when I asked about handling of this type of series previously.

thehomebrewnerd (approved):
One really small thing, but otherwise looks good to me.

woodwork/tests/utils/test_utils.py
jeff-hernandez: removed

PR #1366, opened 2022-03-30
Update to use better error message

### LINKED ISSUE #724: Need better error if user assigns new column to dataframe outside of Woodwork

bchen1116:
Fixes #724 

Verified that it works locally. We use `repr` in the test in order to see the error and allow pytest to catch
image

### DIFF
--- a/woodwork/accessor_utils.py
+++ b/woodwork/accessor_utils.py
@@ -207,7 +207,7 @@ def _check_table_schema(method):
             )
             raise WoodworkNotInitError(msg)
         diff_cols = set(self._dataframe.columns).difference(
-            set(self._schema.logical_types.keys())
+            set(self._schema.columns.keys())
         )
         if diff_cols:
             raise ColumnNotPresentInSchemaError(sorted(list(diff_cols)))
--- a/woodwork/accessor_utils.py
+++ b/woodwork/accessor_utils.py
@@ -3,7 +3,7 @@ from functools import wraps
 import numpy as np
 import pandas as pd
 
-from woodwork.exceptions import WoodworkNotInitError
+from woodwork.exceptions import ColumnNotPresentInSchemaError, WoodworkNotInitError
 from woodwork.utils import _get_column_logical_type, import_or_none
 
 dd = import_or_none("dask.dataframe")
@@ -206,6 +206,11 @@ def _check_table_schema(method):
                 "calling DataFrame.ww.init"
             )
             raise WoodworkNotInitError(msg)
+        diff_cols = set(self._dataframe.columns).difference(
+            set(self._schema.logical_types.keys())
+        )
+        if diff_cols:
+            raise ColumnNotPresentInSchemaError(sorted(list(diff_cols)))
         return method(self, *args, **kwargs)
 
     return wrapper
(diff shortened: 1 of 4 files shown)


woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -2651,8 +2652,9 @@ def test_maintain_column_order_disordered_schema(sample_df):
     column_order = list(sample_df.columns)
 
     scramble_df = sample_df.ww.copy()
-    id_col = scramble_df.ww.columns.pop("id")
-    scramble_df.ww.columns["id"] = id_col
+    table_schema = scramble_df.ww.columns
+    id_col = table_schema.pop("id")
+    table_schema["id"] = id_col
```
bchen1116, line 2657: Worked with @thehomebrewnerd on this. Previously, calling `scramble_df.ww` after popping `id` would cause the `ColumnsNotPresentInSchemaError` to appear. We instead reference it and have verified the test continues to test what we expect here.

woodwork/accessor_utils.py
```diff
@@ -206,6 +206,11 @@ def wrapper(self, *args, **kwargs):
                 "calling DataFrame.ww.init"
             )
             raise WoodworkNotInitError(msg)
+        diff_cols = set(self._dataframe.columns).difference(
+            set(self._schema.logical_types.keys())
```
thehomebrewnerd, line 210: I think you can do `self._schema.columns.keys()` here which might make it a little more clear that we are comparing the dataframe columns to the columns in the schema.

thehomebrewnerd (commented):
Overall looks fine, just a couple minor suggestions.

thehomebrewnerd (approved):
LGTM

Cmancuso:
This looks good to me. Begs the question for me though, how are we handling these exceptions as they're thrown in tempo? And similarly, are we passing back to the client enough metadata that we would be able to properly translate the error messages on the UI side? How can we make that guarantee going forward. Food for thought. Looks good though!

Cmancuso (approved):

ParthivNaresh (approved):
Looks great!

bchen1116:
@Cmancuso definitely agreed. Although I think this error in particular is more geared towards OS users/developers internally, it's definitely important for us to keep in mind that we need to pass enough information downstream to make useful messages and allow users/developers to take action upon the errors raised.

PR #1373, opened 2022-04-04
Cap pandas version

bchen1116:
Pandas 1.4.2 has a bug where `replace` will change the dtypes of other columns not included in the replace call, which did not exist in 1.4.1.

Filed an issue with the pandas-dev team [here](https://github.com/pandas-dev/pandas/issues/46634)!

image

image

There are two options to move forward with this. We can either cap the pandas version (the approach that I chose), or alter the test to allow them to pass with this bug while pandas puts up a fix for it, and follow up on it afterwards when pandas rolls out a fix. 

Planning to use this PR to close out [this PR](https://github.com/alteryx/woodwork/pull/1370)

### DIFF
--- a/woodwork/tests/requirement_files/latest_core_dependencies.txt
+++ b/woodwork/tests/requirement_files/latest_core_dependencies.txt
@@ -1,4 +1,4 @@
 numpy==1.22.3
-pandas==1.4.2
+pandas==1.4.1
 pyarrow==7.0.0
 scikit-learn==1.0.2
--- a/woodwork/tests/requirement_files/latest_core_dependencies.txt
+++ b/woodwork/tests/requirement_files/latest_core_dependencies.txt
@@ -1,4 +1,4 @@
 numpy==1.22.3
-pandas==1.4.1
+pandas==1.4.2
 pyarrow==7.0.0
 scikit-learn==1.0.2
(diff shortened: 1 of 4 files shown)


woodwork/tests/requirement_files/latest_core_dependencies.txt
```diff
@@ -1,4 +1,4 @@
 numpy==1.22.3
-pandas==1.4.1
+pandas==1.4.2
```
gsheni, line 2:
```suggestion
pandas==1.4.1
```

gsheni: You don't want to increase this, until Woodwork supports that pandas version

bchen1116: Oh weird, I thought I had set this to 1.4.1. Will change!

Cmancuso (approved):
LGTM just curious though if we have a plan to open up versions again once this bug is resolved. Is that something we want to do or should we be generally having tighter control over the versions allowed?

bchen1116:
@Cmancuso I can file an issue to track uncapping the pandas version at the next update to include a fix for the mentioned bug!

PR #1375, opened 2022-04-05
Automated Minimum Dependency Updates

machineFL:
This is an auto-generated PR with **minimum** dependency updates. Please do not delete the `min-dep-update` branch because it's needed by the auto-dependency bot.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -40,7 +40,7 @@ include_package_data = True
 install_requires =
     pandas >= 1.3.0, <1.4.2
     scikit-learn >= 0.22
-    pyarrow >= 1.0.0
+    pyarrow >= 5.0.0
 python_requires = >=3.7, <4
 
 [options.extras_require]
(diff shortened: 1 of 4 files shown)


setup.cfg
```diff
@@ -40,7 +40,7 @@ include_package_data = True
 install_requires =
     pandas >= 1.3.0, <1.4.2
     scikit-learn >= 0.22
-    pyarrow >= 1.0.0
+    pyarrow >= 5.0.0
```
bchen1116, line 43: Pyarrow versions less than this don't include the `pyarrow._orc` module that is required for the way ww writes or reads files. We raise the limit of the pyarrow version to satisfy this.

bchen1116 (approved):
LGTM once tests pass

setup.cfg
gsheni: Nice

PR #1376, opened 2022-04-05
Stores timezone in Datetime logical type

### LINKED ISSUE #111: Verify support for Datetime with timezone
- We should verify that we can support inputted datetime columns with a timezone. 
- We should determine if we want to extract the timezone, and store it on the Datetime Logical Type. (Can some rows have different timezone than others or do all datetimes have the same timezone?)
- These will show up as `datetime64[ns, tz]` pandas dtype

jeff-hernandez:
The expected behavior as a result from this PR:

- A `datetime64[ns]` series with no timezone will remain as `datetime64[ns]`. The logical type will have no timezone.
- A `datetime64[ns, UTC]` series will change to  `datetime64[ns]`. The logical type will store the UTC timezone.
- A `datetime64[ns, US/Eastern]` series will change to `datetime64[ns]`. The logical type will store the ET timezone.
- A series that contains datetime strings with timezones will be converted to UTC and then changed to `datetime64[ns]`. The logical types will store the UTC timezone.
- A series that contains datetime strings with no timezones will change to `datetime64[ns]`. The logical types will have no timezone.

#### Limitations

I found a few inconsistencies in pandas, dask, and spark when handling timezones.

- Dask cannot create metadata for timezones

    ```
    >>> series = dd.to_datetime(pd.Series(['2022-01-01T00:00:00+00:00']))
    >>> series.dt.tz_localize('UTC').dt.tz
    TypeError: Don't know how to create metadata from UTC
    ```

- Dask mismatches datetime dtypes between delayed and computed series.

    ```
    >>> x = dd.to_datetime(pd.Series(['2022-01-01T00:00:00+00:00']))

    Dask Series Structure:
    npartitions=1
    0    datetime64[ns]
    0               ...
    dtype: datetime64[ns]
    Dask Name: to_datetime, 2 tasks

    >>> x.compute()

    0   2022-01-01 00:00:00+00:00
    dtype: datetime64[ns, UTC]
    ```

- Pandas and dask cannot infer timezones from strings except UTC. It will return timezone offsets instead.

    ```
    >>> dtype = 'datetime64[ns, US/Eastern]'
    >>> x = pd.Series(['2022-01-01', '2022-01-02', 'NaT'], dtype=dtype)
    >>> pd.to_datetime(x.astype(str))

        0   2022-01-01 00:00:00-05:00
    1   2022-01-02 00:00:00-05:00
    2                         NaT
    dtype: datetime64[ns, pytz.FixedOffset(-300)]

    >>> pd.to_datetime(x.astype(str), utc=True)

    0   2022-01-01 05:00:00+00:00
    1   2022-01-02 05:00:00+00:00
    2                         NaT
    dtype: datetime64[ns, UTC]
    ```

-  Spark doesn't support dateime series with timezones.

    ```
    >>> data = ['2022-01-01 00:00:00+00:00', '2022-01-02 00:00:00+00:00', 'NaT']
    >>> ps.from_pandas(pd.Series(data, dtype='datetime64[ns, UTC]'))
    TypeError: Type datetime64[ns, UTC] was not understood.

    >>> ps.from_pandas(pd.Series(data)).astype('datetime64[ns, UTC]')
    TypeError: Type datetime64[ns, UTC] was not understood.
    ```

Closes #111

### DIFF
--- a/woodwork/logical_types.py
+++ b/woodwork/logical_types.py
@@ -278,6 +278,7 @@ class Datetime(LogicalType):
         self.timezone = None
 
     def _remove_timezone(self, series):
+        """Removes timezone from series and stores in logical type."""
         if hasattr(series.dtype, 'tz') and series.dtype.tz:
             self.timezone = str(series.dtype.tz)
             series = series.dt.tz_localize(None)
@@ -322,7 +323,6 @@ class Datetime(LogicalType):
                         series, format=self.datetime_format, errors="coerce"
                     )
 
-        # series = self._remove_timezone(series)
         return super().transform(series)
 
 
--- a/woodwork/logical_types.py
+++ b/woodwork/logical_types.py
@@ -279,11 +279,10 @@ class Datetime(LogicalType):
 
     def _remove_timezone(self, series):
         """Removes timezone from series and stores in logical type."""
-        if hasattr(series.dtype, 'tz') and series.dtype.tz:
+        if hasattr(series.dtype, "tz") and series.dtype.tz:
             self.timezone = str(series.dtype.tz)
             series = series.dt.tz_localize(None)
         return series
-
 
     def transform(self, series):
         """Converts the series data to a formatted datetime. Datetime format will be inferred if datetime_format is None."""
@@ -295,10 +294,11 @@ class Datetime(LogicalType):
             self.datetime_format = self.datetime_format or _infer_datetime_format(
                 series
             )
+            utc = self.datetime_format.endswith("%z")
             if _is_dask_series(series):
                 name = series.name
                 series = dd.to_datetime(
-                    series, format=self.datetime_format, errors="coerce"
+                    series, format=self.datetime_format, errors="coerce", utc=utc
                 )
                 series.name = name
             elif _is_spark_series(series):
@@ -310,7 +310,9 @@ class Datetime(LogicalType):
                 )
             else:
                 try:
-                    series = pd.to_datetime(series, format=self.datetime_format)
+                    series = pd.to_datetime(
+                        series, format=self.datetime_format, utc=utc
+                    )
                 except (TypeError, ValueError):
                     warnings.warn(
                         f"Some rows in series '{series.name}' are incompatible with datetime format "
@@ -323,6 +325,7 @@ class Datetime(LogicalType):
                         series, format=self.datetime_format, errors="coerce"
                     )
 
+        series = self._remove_timezone(series)
         return super().transform(series)
 
 
--- a/woodwork/logical_types.py
+++ b/woodwork/logical_types.py
@@ -273,8 +273,9 @@ class Datetime(LogicalType):
     primary_dtype = "datetime64[ns]"
     datetime_format = None
 
-    def __init__(self, datetime_format=None):
+    def __init__(self, datetime_format=None, timezone=None):
         self.datetime_format = datetime_format
+        self.timezone = None
 
     def transform(self, series):
         """Converts the series data to a formatted datetime. Datetime format will be inferred if datetime_format is None."""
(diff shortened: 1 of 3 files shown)


woodwork/tests/conftest.py
```diff
@@ -914,3 +914,48 @@ def outliers_df_spark(outliers_df_pandas):
...
+                "2022-01-01T00:00:00+00:00",
+                "2022-01-01 00:00:00-05:00",
+            ],
+            [
+                "2022-01-02",
+                "2022-01-02 00:00:00+00:00",
+                "2022-01-02 00:00:00-05:00",
+                "2022-01-02",
+                "2022-01-02T00:00:00+00:00",
+                "2022-01-02 00:00:00-05:00",
```
bchen1116, line 945: The last column here is supposed to be `Eastern`, right? Since there is the `-05:00` time? The column name calls it `eastern_2`, but the test that uses this checks that it's `UTC`. Can you explain which one it should be?

bchen1116 (commented):
Tests look good, I just left a few comments on questions that I had. 

Great job digging into why dask, pandas, and spark each have their limitations! Made it much easier to understand the implementation choices here.

woodwork/tests/conftest.py
jeff-hernandez: Sure thing, so the original datetime string `00:00:00-05:00` contains an eastern timezone, but pandas cannot infer the timezone from the string and instead returns a fixed offset, so preserving the timezone is done by converting the datetime string to UTC `05:00:00+00:00`. Otherwise, the corresponding timezone gets lost after converting to `datetime64[ns]`. The test checks for the expected UTC conversion.

woodwork/logical_types.py
```diff
@@ -313,6 +325,8 @@ def transform(self, series):
                     series = pd.to_datetime(
                         series, format=self.datetime_format, errors="coerce"
```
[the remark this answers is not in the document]
jeff-hernandez: MIssed it, good catch!

woodwork/tests/conftest.py
bchen1116: Got it, thanks for explaining!

Cmancuso:
Do we currently lose the timezone when we write to disk? Would we need @thehomebrewnerd 's change to go in in order to actually maintain timezone information on the datatable across writes?

Cmancuso:
Can you also elaborate on this point? 

> A series that contains datetime strings with timezones will be converted to UTC and then changed to datetime64[ns]. The logical types will store the UTC timezone.

Does this mean series with timezones are stripped of their TZ and replaced with UTC?

bchen1116 (approved):
Looks great to me! I think the only thing I would potentially add are datetime tests with varying formats of datetime inputs, but I think this has good coverage for our expected pandas and dask support for datetime!

jeff-hernandez:
@Cmancuso 

> Do we currently lose the timezone when we write to disk? 

No, the timezone currently gets written to disk as a parameter of the datetime logical type in the json file. 


> Does this mean series with timezones are stripped of their TZ and replaced with UTC?

The current Woodwork behavior (on the main branch) already strips datetime strings from their timezone and coverts it without storing the new UTC timezone. This happens as a result of a dtype conversion from `string` to `datetime64[ns]`.

```python
>>> data = ['2022-01-02 00:00:00-05:00']  # ET
>>> pd.Series(data, dtype='string').astype('datetime64[ns]') # UTC
0   2022-01-02 05:00:00
dtype: datetime64[ns]
```


In this PR, the UTC conversion still happens because of a pandas limitation, but we are now storing the new UTC timezone in the logical type. Ideally, the timezone from a datetime string shouldn't change but pandas cannot infer timezones from strings.

Cmancuso (approved):
Thanks for the clarifications @jeff-hernandez, looks good!

PR #1377, opened 2022-04-07
Use single file for parquet serialization

### LINKED ISSUE #968: Update parquet serialization to use a single file for data and typing info
- As a user, I wish I Woodwork would store all data and typing information in a single parquet file when serializing, rather than writing a data file and a separate typing information JSON file.

The example below shows how we can use pyarrow to add custom typing information to the parquet file metadata prior to writing the file, along with showing how the typing information can be read from the metadata during deserialization for pandas. Dask allows users to write directly to the metadata during serialization by passing data in the `cusom_metadata` argument during the call to `df.to_parquet`. Koalas will likely need to follow a similar approach to pandas.

This update should not change the behavior when serializing to csv or pickle.

#### Code Example - Pandas

```python
>>> import pandas as pd
>>> import pyarrow as pa
>>> import pyarrow.parquet as pq
>>> import json
>>> 
>>> df = pd.DataFrame({
...     'id': [0, 1, 2],
...     'bools': [True, False, True],
...     'strings': ['a', 'b', 'c']
... })
>>> table = pa.Table.from_pandas(df)
>>> metadata = table.schema.metadata
>>> metadata
{b'pandas': b'{"index_columns": [{"kind": "range", "name": null, "start": 0, "stop": 3, "step": 1}], "column_indexes": [{"name": null, "field_name": null, "pandas_type": "unicode", "numpy_type": "object", "metadata": {"encoding": "UTF-8"}}], "columns": [{"name": "id", "field_name": "id", "pandas_type": "int64", "numpy_type": "int64", "metadata": null}, {"name": "bools", "field_name": "bools", "pandas_type": "bool", "numpy_type": "bool", "metadata": null}, {"name": "strings", "field_name": "strings", "pandas_type": "unicode", "numpy_type": "object", "metadata": null}], "creator": {"library": "pyarrow", "version": "4.0.1"}, "pandas_version": "1.2.4"}'}
>>> 
>>> custom_meta = {'custom_metadata': 'this is user defined table metadata',
...                'columns': {'id_metadata': 'this is metadata defined for the id column'}}
>>> combined_meta = {'ww_meta'.encode(): json.dumps(custom_meta).encode(), **metadata}
>>> table = table.replace_schema_metadata(combined_meta)
>>> pq.write_table(table, 'test_meta.parquet')
>>> 
>>> file_metadata = pa.parquet.read_metadata('test_meta.parquet')
>>> file_metadata.metadata
{b'ARROW:schema': b'/////zgEAAAQAAAAAAAKAA4ABgAFAAgACgAAAAABBAAQAAAAAAAKAAwAAAAEAAgACgAAAGADAAAEAAAAAgAAALACAAAEAAAAYP3//5ACAAAEAAAAggIAAHsiaW5kZXhfY29sdW1ucyI6IFt7ImtpbmQiOiAicmFuZ2UiLCAibmFtZSI6IG51bGwsICJzdGFydCI6IDAsICJzdG9wIjogMywgInN0ZXAiOiAxfV0sICJjb2x1bW5faW5kZXhlcyI6IFt7Im5hbWUiOiBudWxsLCAiZmllbGRfbmFtZSI6IG51bGwsICJwYW5kYXNfdHlwZSI6ICJ1bmljb2RlIiwgIm51bXB5X3R5cGUiOiAib2JqZWN0IiwgIm1ldGFkYXRhIjogeyJlbmNvZGluZyI6ICJVVEYtOCJ9fV0sICJjb2x1bW5zIjogW3sibmFtZSI6ICJpZCIsICJmaWVsZF9uYW1lIjogImlkIiwgInBhbmRhc190eXBlIjogImludDY0IiwgIm51bXB5X3R5cGUiOiAiaW50NjQiLCAibWV0YWRhdGEiOiBudWxsfSwgeyJuYW1lIjogImJvb2xzIiwgImZpZWxkX25hbWUiOiAiYm9vbHMiLCAicGFuZGFzX3R5cGUiOiAiYm9vbCIsICJudW1weV90eXBlIjogImJvb2wiLCAibWV0YWRhdGEiOiBudWxsfSwgeyJuYW1lIjogInN0cmluZ3MiLCAiZmllbGRfbmFtZSI6ICJzdHJpbmdzIiwgInBhbmRhc190eXBlIjogInVuaWNvZGUiLCAibnVtcHlfdHlwZSI6ICJvYmplY3QiLCAibWV0YWRhdGEiOiBudWxsfV0sICJjcmVhdG9yIjogeyJsaWJyYXJ5IjogInB5YXJyb3ciLCAidmVyc2lvbiI6ICI0LjAuMSJ9LCAicGFuZGFzX3ZlcnNpb24iOiAiMS4yLjQifQAABgAAAHBhbmRhcwAACAAMAAQACAAIAAAAlAAAAAQAAACEAAAAeyJjdXN0b21fbWV0YWRhdGEiOiAidGhpcyBpcyB1c2VyIGRlZmluZWQgdGFibGUgbWV0YWRhdGEiLCAiY29sdW1ucyI6IHsiaWRfbWV0YWRhdGEiOiAidGhpcyBpcyBtZXRhZGF0YSBkZWZpbmVkIGZvciB0aGUgaWQgY29sdW1uIn19AAAAAAcAAAB3d19tZXRhAAMAAABwAAAAMAAAAAQAAACs////AAABBRAAAAAYAAAABAAAAAAAAAAHAAAAc3RyaW5ncwDY////1P///wAAAQYQAAAAHAAAAAQAAAAAAAAABQAAAGJvb2xzAAAABAAEAAQAAAAQABQACAAGAAcADAAAABAAEAAAAAAAAQIQAAAAHAAAAAQAAAAAAAAAAgAAAGlkAAAIAAwACAAHAAgAAAAAAAABQAAAAAAAAAA=', b'pandas': b'{"index_columns": [{"kind": "range", "name": null, "start": 0, "stop": 3, "step": 1}], "column_indexes": [{"name": null, "field_name": null, "pandas_type": "unicode", "numpy_type": "object", "metadata": {"encoding": "UTF-8"}}], "columns": [{"name": "id", "field_name": "id", "pandas_type": "int64", "numpy_type": "int64", "metadata": null}, {"name": "bools", "field_name": "bools", "pandas_type": "bool", "numpy_type": "bool", "metadata": null}, {"name": "strings", "field_name": "strings", "pandas_type": "unicode", "numpy_type": "object", "metadata": null}], "creator": {"library": "pyarrow", "version": "4.0.1"}, "pandas_version": "1.2.4"}', b'ww_meta': b'{"custom_metadata": "this is user defined table metadata", "columns": {"id_metadata": "this is metadata defined for the id column"}}'}
>>> 
>>> new_df = pd.read_parquet('test_meta.parquet')
>>> new_df
   id  bools strings
0   0   True       a
1   1  False       b
2   2   True       c
>>> 
>>> pd.testing.assert_frame_equal(df, new_df)
>>> 
>>> ww_meta = json.loads(file_metadata.metadata[b'ww_meta'])
>>> ww_meta
{'custom_metadata': 'this is user defined table metadata', 'columns': {'id_metadata': 'this is metadata defined for the id column'}}

```

#### Code Example - Dask

```python
>>> import pandas as pd
>>> import pyarrow as pa
>>> import dask.dataframe as dd
>>> import json
>>> 
>>> df = pd.DataFrame({
...     'id': [0, 1, 2],
...     'bools': [True, False, True],
...     'strings': ['a', 'b', 'c']
... })
>>> df = dd.from_pandas(df, npartitions=2)
>>> custom_meta = {'custom_metadata': 'this is user defined table metadata',
...                'columns': {'id_metadata': 'this is metadata defined for the id column'}}
>>>                
>>> df.to_parquet('test_dask', custom_metadata={'ww_meta': json.dumps(custom_meta)})
>>> metadata = pa.parquet.read_metadata('test_dask/part.0.parquet')
>>> new_meta = json.loads(metadata.metadata[b'ww_meta'])
>>> new_meta
{'custom_metadata': 'this is user defined table metadata', 'columns': {'id_metadata': 'this is metadata defined for the id column'}}

```

thehomebrewnerd:
- Use single file for parquet serialization
- Closes #968 

Implements changes to store Woodwork typing information in parquet file metadata instead of in `woodwork_typing_info.json` file (or equivalent).


woodwork/deserializers/utils.py
```diff
@@ -26,16 +26,35 @@
...
-    path, filename, data_subdirectory, typing_info_filename, profile_name
+    path,
+    filename,
+    data_subdirectory,
+    typing_info_filename,
+    profile_name,
+    format,
 ):
     """Determine the proper Deserializer class to use based on the specified parameters.
     Initializes and returns the proper Deserializer object."""
```
bchen1116, line 37: Would also be nice to add `Args` and `Returns` info do docstring here

woodwork/serializers/parquet_serializer.py
```diff
@@ -26,6 +39,69 @@ def serialize(self, dataframe, profile_name, **kwargs):
         return super().serialize(dataframe, profile_name, **kwargs)
 
     def write_dataframe(self):
```
bchen1116, line 41: nit: Can we rename this function since it isn't actually writing the dataframe now? Perhaps `make_pyarrow` or something of the sort would be more in line with the method

woodwork/deserializers/parquet_deserializer.py
```diff
@@ -1,11 +1,79 @@
...
+from woodwork.s3_utils import get_transport_params, use_smartopen
+from woodwork.utils import _is_s3, _is_url
 
 
 class ParquetDeserializer(Deserializer):
     """Deserialize Woodwork table from serialized data in parquet format."""
 
     format = "parquet"
 
+    def deserialize(self, profile_name, validate):
```
bchen1116, line 19: Nit: Can we add docstrings for these methods?

woodwork/serializers/parquet_serializer.py
bchen1116: Also docstrings would be helpful here! I realize these methods were already missing docstrings before, but would be great to add them or file an issue to track adding them.

woodwork/serializers/parquet_serializer.py
```diff
@@ -26,6 +39,69 @@ def serialize(self, dataframe, profile_name, **kwargs):
...
+            "type": self.format,
+            "params": self.kwargs,
+        }
+        self.typing_info["loading_info"].update(loading_info)
+        # For Dask and Spark we only get the WW metadata because we haven't created
+        # the pyarrow table yet, but for pandas we combine the existing parquet
+        # metadata with the WW metadata.
+        if _is_dask_dataframe(self.dataframe) or _is_spark_dataframe(self.dataframe):
+            metadata = {
+                "ww_meta".encode(): json.dumps(self.typing_info).encode(),
```
bchen1116, line 59: the `.encode()` is really cool! I didn't realize this existed.

woodwork/serializers/parquet_serializer.py
```diff
@@ -26,6 +39,69 @@ def serialize(self, dataframe, profile_name, **kwargs):
...
+            }
+        else:
+            table_metadata = self.table.schema.metadata
+            metadata = {
+                "ww_meta".encode(): json.dumps(self.typing_info).encode(),
+                **table_metadata,
```
bchen1116, line 65: Question: Here, we're writing the `table_metadata` into the metadata dictionary, but in `deserialize.configure_deserializer`, we only look at the `ww_meta` value. Do we read/look for the `table_metadata` keys/values elsewhere?

woodwork/serializers/parquet_serializer.py
```diff
@@ -26,6 +39,69 @@ def serialize(self, dataframe, profile_name, **kwargs):
...
+            }
+        self._save_parquet_table_to_disk(metadata)
+
+    def _save_parquet_table_to_disk(self, metadata):
+        if _is_dask_dataframe(self.dataframe):
+            path, dataframe = self._setup_for_dask_and_spark()
+            dataframe.to_parquet(path, custom_metadata=metadata)
+        elif _is_spark_dataframe(self.dataframe):
```
bchen1116, line 73: This is cool! So with `Dask` and `Pandas`, the writing is fairly straightforward, but for `spark`, we need to write to parquet, open the first file, and then update the metadata of that file with the table schema meta. Interesting how the behaviors of these are so different with parquet support.

woodwork/tests/accessor/test_serialization.py
```diff
@@ -480,17 +504,18 @@ def test_to_disk_custom_data_filename(sample_df, tmpdir, file_format):
...
             to_pandas(sample_df, index=sample_df.ww.index, sort_index=True),
             to_pandas(deserialized_df, index=deserialized_df.ww.index, sort_index=True),
         )
         assert sample_df.ww.schema == deserialized_df.ww.schema
 
 
-@pytest.mark.parametrize(
-    "file_format", ["csv", "pickle", "parquet", "arrow", "feather", "orc"]
-)
+@pytest.mark.parametrize("file_format", ["csv", "pickle", "arrow", "feather", "orc"])
```
bchen1116, line 518: Since we took parquet out of this, do we have other test coverage to ensure that custom typing changes we've made to the dataframe are preserved when we write to and then read from parquet?

woodwork/tests/accessor/test_serialization.py
```diff
@@ -601,6 +647,54 @@ def test_to_disk_with_latlong(latlong_df, tmpdir, file_format):
...
+        filename=filename,
+        typing_info_filename=None,
+        format=format,
+    )
+    pd.testing.assert_frame_equal(
+        to_pandas(sample_df, index=sample_df.ww.index, sort_index=True),
+        to_pandas(deserialized_df, index=deserialized_df.ww.index, sort_index=True),
+    )
+
+
```
bchen1116, line 697: I think it'd be beneficial to do a test where we explicitly check that the `ww_meta` in the parquet file metadata matches what we expect given some dataframe that we've initialized and altered. This would give us a good indication on whether or not the serializer code works as we'd expect.

bchen1116 (commented):
Great work on this Nate! I left a few nits on adding documentation, some questions, and some testing suggestions. This is quite an intense PR, so great job on moving quickly with this.

woodwork/deserializers/utils.py
thehomebrewnerd: I can add those. Also, that made me think that this should probably be a private function. I would not ever expect a user to call this function and interact with the deserializer object that is returned.

woodwork/serializers/parquet_serializer.py
thehomebrewnerd: We don't do anything with the `table_metadata` inside Woodwork, but there is other information stored in there that is needed for deserialization into a pandas DataFrame. If we don't include it here, deserialization breaks.

woodwork/serializers/parquet_serializer.py
thehomebrewnerd: Yeah, unfortunately these behavior differences make this code flow a lot more complicated than I would prefer, but not sure how to get around it at this point exactly.

woodwork/tests/accessor/test_serialization.py
thehomebrewnerd: Ah, the intent of this test is to confirm that if a user specifies something for `typing_info_filename` in the `ww.to_disk` call that we can properly deserialize.

With this PR, parquet no longer writes a separate typing info JSON file so we don't need to test this behavior any more.

woodwork/serializers/parquet_serializer.py
bchen1116: Mm, that's interesting. Thanks for the explanation

woodwork/deserializers/parquet_deserializer.py
thehomebrewnerd: I had originally left docstrings off of all these methods in the serializer/deserializer classes since they are defined on the `Serializer` and `Deserializer` base classes and thought it might be redundant to include them on the subclasses (and something else that would need to be maintained if the base docstrings changed).

I'm not opposed to adding them here though - especially for methods that override the base class method and might behave slightly differently.

woodwork/serializers/parquet_serializer.py
thehomebrewnerd: Regarding renaming the function - it's a little complicated. If you look at the code flow we first call the `serialize` method on this class, then the `serialize` method on the base class, and that ends up calling `save_to_local_path` (also from the base class) which then calls this `write_dataframe` method which overrides the base class method.

So, if we renamed this we would have to do one of two things:
1. Override the `save_to_local_path` method on this subclass to call the renamed method instead of `write_dataframe`
2. Leave a `write_dataframe` method here that simply calls the renamed method.

If we simply renamed this, when `Serializer` base tries to call `write_dataframe` we would get a `NotImplementedError` because the base class doesn't implement this method and neither would the subclass in this case.

I could also add a note here to explain why this behavior is different for this serializer and leave things as is.

Thoughts on any of these options?

woodwork/deserializers/parquet_deserializer.py
bchen1116: Great point, I missed that! It could just be nice to add them when there are additional args or different behavior to clarify how it differs from the base class.

woodwork/tests/accessor/test_serialization.py
bchen1116: Got it.

woodwork/deserializers/utils.py
thehomebrewnerd: Updated [efc0cd8](https://github.com/alteryx/woodwork/pull/1377/commits/efc0cd89032b4253fb601fde6b5f4c5c315dfbc7)

woodwork/serializers/parquet_serializer.py
thehomebrewnerd: Ended up going with option 1 and renamed a couple other methods to make things more clear. Added docstrings for those unique methods as well.

woodwork/tests/accessor/test_serialization.py
thehomebrewnerd: Added a new test in  for this purpose.

thehomebrewnerd:
@bchen1116 I think I addressed all of your comments with the recent updates, but please let me know if I missed anything. Thanks!

woodwork/serializers/parquet_serializer.py
```diff
@@ -25,7 +38,89 @@ def serialize(self, dataframe, profile_name, **kwargs):
         self.kwargs["engine"] = "pyarrow"
         return super().serialize(dataframe, profile_name, **kwargs)
 
-    def write_dataframe(self):
-        file = self._get_filename()
-        dataframe = clean_latlong(self.dataframe)
-        dataframe.to_parquet(file, **self.kwargs)
+    def save_to_local_path(self):
+        """Serialize data and typing information to a local directory. Overrides method on base class
+        due to different serialiation flow"""
```
bchen1116, line 43: super nit: period

woodwork/tests/accessor/test_serialization.py
```diff
@@ -601,6 +649,93 @@ def test_to_disk_with_latlong(latlong_df, tmpdir, file_format):
...
+    assert cat_info["logical_type"]["type"] == "CountryCode"
+    assert "age" in age_info["semantic_tags"]
+
+    # location, type and params are added during serialization, so they are not present
+    # in the expected typing information created from the Woodwork dataframe.
+    del ww_meta["loading_info"]["location"]
+    del ww_meta["loading_info"]["type"]
+    del ww_meta["loading_info"]["params"]
+
+    assert ww_meta == expected_typing_info
```
bchen1116, line 736: big fan of this! Thanks for adding it.

woodwork/tests/accessor/test_serialization.py
```diff
@@ -601,6 +649,93 @@ def test_to_disk_with_latlong(latlong_df, tmpdir, file_format):
...
+        typing_info_filename=None,
+        format=format,
+    )
+    pd.testing.assert_frame_equal(
+        to_pandas(sample_df, index=sample_df.ww.index, sort_index=True),
+        to_pandas(deserialized_df, index=deserialized_df.ww.index, sort_index=True),
+    )
+
+
+def test_to_disk_parquet_saves_custom_metdata_as_expected(sample_df, tmpdir):
```
bchen1116, line 700: nit: typo: metadata vs metdata

woodwork/serializers/parquet_serializer.py
```diff
@@ -25,7 +38,89 @@ def serialize(self, dataframe, profile_name, **kwargs):
...
+        if self.data_subdirectory:
+            location = os.path.join(self.write_path, self.data_subdirectory)
+            os.makedirs(location, exist_ok=True)
+        else:
+            os.makedirs(self.write_path, exist_ok=True)
+        self._create_pyarrow_table()
+        self._generate_parquet_metadata()
+        self._save_parquet_table_to_disk()
+
+    def _create_pyarrow_table(self):
```
bchen1116, line 53: thanks for changing the name! I think this makes it clearer what the function does!

bchen1116 (approved):
Great work with this, Nate! I left a few small nits, but LGTM otherwise!

woodwork/tests/accessor/test_serialization.py
thehomebrewnerd: Fixed

woodwork/serializers/parquet_serializer.py
thehomebrewnerd: Added

woodwork/serializers/parquet_serializer.py
```diff
@@ -12,6 +21,10 @@ class ParquetSerializer(Serializer):
 
     format = "parquet"
 
+    def __init__(self, path, filename, data_subdirectory, typing_info_filename):
+        super().__init__(path, filename, data_subdirectory, typing_info_filename)
+        self.typing_info_filename = None
```
rwedge, line 26: should we warn the user that we're ignoring the`typing_info_filename` if they provided one?

thehomebrewnerd: Sure. Added [82256d9](https://github.com/alteryx/woodwork/pull/1377/commits/82256d90c00a47c4c2e53cc16d292d90e267ba19)

thehomebrewnerd: Updated to check here ([e1cec8a](https://github.com/alteryx/woodwork/pull/1377/commits/e1cec8ae45bbecca7534f88832e19bdb26d59d1d)) to account for the fact that the default parameter value will be `woodwork_typing_info.json` when the user doesn't pass a value. This means that if the user actually specifies `woodwork_typing_info.json` they won't get a warning, but if they specify any other value except `None` they will get a warning.

rwedge (approved):
LGTM

PR #1378, opened 2022-04-07
Add postal code inference

### LINKED ISSUE #1312: Add US Postal Code Inference
- We can use a regex to support inferring Postal Code in Woodwork
  - https://www.oreilly.com/library/view/regular-expressions-cookbook/9781449327453/ch04s14.html
- The logic would only look at the first 10K rows

bchen1116:
fix #1312 

Only adds inference for US postal codes. Other countries, like [UK](https://www.oreilly.com/library/view/regular-expressions-cookbook/9781449327453/ch04s16.html) and [Canadian](https://www.oreilly.com/library/view/regular-expressions-cookbook/9781449327453/ch04s15.html) postal codes, follow different formats.

### DIFF
--- a/woodwork/tests/logical_types/test_logical_types.py
+++ b/woodwork/tests/logical_types/test_logical_types.py
@@ -371,6 +371,25 @@ def test_string_dtype_validate(sample_df):
     assert lt.dtype == "string[pyarrow]"
 
 
+def test_postal_code_validate(sample_df_postal_code):
+    pc = PostalCode()
+    series = sample_df_postal_code["postal_code"]
+    invalid_types = pd.Series(
+        [
+            "hello",
+            "HELLO",
+            "51342-HEL0",
+        ]
+    )
+    series = series.append(invalid_types)
+    series.name = "postal_code"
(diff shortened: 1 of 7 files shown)


ParthivNaresh:
@bchen1116 Solid work, could you add Postal Code to `sample_df_pandas` as well so all relevant tests are included for this?

woodwork/tests/logical_types/test_logical_types.py
```diff
@@ -368,3 +369,27 @@ def test_string_dtype_validate(sample_df):
     assert logical_type.validate(series) is None
     lt = logical_type.transform(series)
     assert lt.dtype == "string[pyarrow]"
+
+
+def test_postal_code_validate_complex(sample_df_postal_code):
```
jeff-hernandez, line 374: It might be good to also have `test_postal_code_validate` to test the type validation error like the other logical types

jeff-hernandez (approved):
Overall looks good, just had a comment.

ParthivNaresh (approved):
Now that we've clarified that we don't need to constantly update `sample_df` this is definitely good to go!

Issue #760, opened 2021-04-01
Run testing for spark, dask, and pandas separate


gsheni:
- We can improve our time to run unit tests by running koalas, dask, and pandas separately, and then combining the coverage report (CodeCov automatically [supports](https://docs.codecov.io/docs/merging-reports) this).
- This will improve the time to run the unit tests. 
- We will need to use pytest markers for this

PR #1380, opened 2022-04-11
Parallelize tests

ParthivNaresh:
Fixes: #760 

The issue surrounding the current testing time is primarily associated with `accessor/test_serialization.py` and `accessor/test_table_accessor.py.` These two files contain several tests accounting for the vast majority of the testing time.

Within `test_serialization.py`, `test_to_disk_with_latlong` itself takes around 4 min to run.

Taking the past 10 Linux Test GitHub jobs, the average time until completion is currently 18 min and 4 seconds.

After separating out the serialization tests, over the past 10 jobs the new average time until completion was 12 min and 44 seconds, a 30% improvement.

After further separation (the proposed implementation), over the past 10 jobs the new average time until completion was 10 min and 35 seconds, a 41% improvement.

### DIFF
--- a/.github/workflows/unit_tests_with_latest_deps.yml
+++ b/.github/workflows/unit_tests_with_latest_deps.yml
@@ -61,7 +61,7 @@ jobs:
         env:
           PYARROW_IGNORE_TIMEZONE: 1
           ALTERYX_OPEN_SRC_UPDATE_CHECKER: False
-      - if: ${{ matrix.python_version == 3.8 && matrix.directories == 'Non-serialization' }}
+      - if: ${{ matrix.python_version == 3.8 && matrix.directories == 'Serialization' }}
         name: Run Serialization Unit Tests with code coverage
         run: |
           python -m pip install "$(cat setup.cfg | grep codecov)"
@@ -72,7 +72,7 @@ jobs:
         env:
           PYARROW_IGNORE_TIMEZONE: 1
           ALTERYX_OPEN_SRC_UPDATE_CHECKER: False
-      - if: ${{ matrix.python_version == 3.8 && matrix.directories == 'everything_else' }}
+      - if: ${{ matrix.python_version == 3.8 && matrix.directories == 'Non-serialization' }}
         name: Run Non-serialization Unit Tests with code coverage
         run: |
           python -m pip install "$(cat setup.cfg | grep codecov)"
--- a/.github/workflows/unit_tests_with_latest_deps.yml
+++ b/.github/workflows/unit_tests_with_latest_deps.yml
@@ -15,7 +15,7 @@ jobs:
       fail-fast: true
       matrix:
         python_version: ["3.7", "3.8", "3.9", "3.10"]
-        libraries: ["core", "dask", "spark"]
+        directories: ["everything_else", "accessor"]
     steps:
       - name: Set up python ${{ matrix.python_version }}
         uses: actions/setup-python@v2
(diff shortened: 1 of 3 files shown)


.github/workflows/unit_tests_with_latest_deps.yml
```diff
@@ -37,33 +38,48 @@ jobs:
...
           ALTERYX_OPEN_SRC_UPDATE_CHECKER: False
-      - if: ${{ matrix.python_version == 3.8 }}
-        name: Run unit tests with code coverage
+      - if: ${{ matrix.python_version != 3.8 && matrix.directories == 'Non-serialization' }}
+        name: Run Non-serialization Unit Tests (no code coverage)
+        run: |
+          cd unpacked_sdist
+          pytest woodwork/ -n 2 --ignore=woodwork/tests/accessor/test_serialization.py --durations 0
+        env:
+          PYARROW_IGNORE_TIMEZONE: 1
```
gsheni, line 62: You can set the ENV for the whole workflow:
- https://docs.github.com/en/enterprise-cloud@latest/actions/learn-github-actions/environment-variables#about-environment-variables

.github/workflows/unit_tests_with_latest_deps.yml
```diff
@@ -37,36 +42,65 @@ jobs:
...
+        run: |
+          cd unpacked_sdist
+          pytest woodwork/tests/accessor/test_table_accessor.py -n 2 --durations 0
+      - if: ${{ matrix.python_version != 3.8 && matrix.directories == 'All Other Tests' }}
+        name: Run all other Unit Tests (no code coverage)
+        run: |
+          cd unpacked_sdist
+          pytest woodwork/ -n 2 --ignore=woodwork/tests/accessor/test_serialization.py --ignore=woodwork/tests/accessor/test_table_accessor.py --durations 0
+      - if: ${{ matrix.python_version == 3.8 && matrix.directories == 'Testing to Disk with LatLong' }}
+        name: Run Serialization Unit Tests with code coverage
```
bchen1116, line 73: Wrong test name here?

Makefile
```diff
@@ -21,7 +21,11 @@ lint-fix:
 
 .PHONY: test
 test:
-	pytest woodwork/
+	pytest woodwork/ --durations 10
```
bchen1116, line 24: do we need the `--durations` flag here?

Makefile
```diff
@@ -21,7 +21,11 @@ lint-fix:
...
+
+.PHONY: test-logical-types
+test-logical-types:
```
bchen1116, line 27: Why did we add this?

docs/source/release_notes.rst
```diff
@@ -23,9 +23,10 @@ Future Release
         * Add M1 installation instructions to docs and contributing guide (:pr:`1367`)
         * Update README text to Alteryx (:pr:`1381`)
     * Testing Changes
+        * Separate testing matrix to speed up GitHub Actions Linux tests for latest dependencies :pr:`1380`
 
     Thanks to the following people for contributing to this release:
-    :user:`bchen1116`, :user:`gsheni`, :user:`jeff-hernandez`, :user:`ParthivNaresh`, :user:`rwedge`, :user:`thehomebrewnerd`
+    :user:`bchen1116`, :user:`gsheni`, :user:`jeff-hernandez`, :user:`ParthivNaresh`, :user:`rwedge`, :user:`thehomebrewnerd`, :user:`ParthivNaresh`
```
bchen1116, line 29: Your name is already in the list

.github/workflows/unit_tests_with_latest_deps.yml
```diff
@@ -37,36 +42,65 @@ jobs:
...
+        run: |
+          python -m pip install "$(cat setup.cfg | grep codecov)"
+          coverage erase
+          cd unpacked_sdist/
+          coverage erase
+          pytest woodwork/tests/accessor/test_serialization.py::test_to_disk_with_latlong -n 2 --durations 0 --cov=woodwork --cov-config=../.coveragerc
+      - if: ${{ matrix.python_version == 3.8 && matrix.directories == 'All other Serialization' }}
+        name: Run Serialization Unit Tests with code coverage
```
bchen1116, line 81: This should be 'run all other serialization' right?

.github/workflows/unit_tests_with_latest_deps.yml
```diff
@@ -37,36 +42,65 @@ jobs:
...
-        env:
-          PYARROW_IGNORE_TIMEZONE: 1
-          ALTERYX_OPEN_SRC_UPDATE_CHECKER: False
-      - if: ${{ matrix.python_version == 3.8 }}
-        name: Run unit tests with code coverage
+          pytest woodwork/tests/accessor/test_serialization.py::test_to_disk_with_latlong -n 2 --durations 0
+      - if: ${{ matrix.python_version != 3.8 && matrix.directories == 'All other Serialization' }}
+        name: Run all other Serialization Unit Tests (no code coverage)
+        run: |
+          cd unpacked_sdist
```
bchen1116, line 60: Can we run `cd unpacked_sdist` before we get to all of these `if` statements? Would lead to marginal code cleanup, but would be nice to not have to repeat this call in every `if` cell

.github/workflows/unit_tests_with_latest_deps.yml
```diff
@@ -37,36 +42,65 @@ jobs:
...
+          coverage erase
+          cd unpacked_sdist/
+          coverage erase
+          pytest woodwork/tests/accessor/test_serialization.py::test_to_disk_with_latlong -n 2 --durations 0 --cov=woodwork --cov-config=../.coveragerc
+      - if: ${{ matrix.python_version == 3.8 && matrix.directories == 'All other Serialization' }}
+        name: Run Serialization Unit Tests with code coverage
+        run: |
+          python -m pip install "$(cat setup.cfg | grep codecov)"
+          coverage erase
+          cd unpacked_sdist/
```
bchen1116, line 85: Similarly for these 4 calls (before `pytest woodwork...`, can we run these before the `if` statements or outside of that logic? Would allow us to clean this up a little bit and make the file a bit more readable.

bchen1116 (commented):
Great work on getting this speed up! Left some questions and suggestions on potential changes, lmk what you think!

Makefile
ParthivNaresh: I felt like this was a nice inclusion especially if we can transition in the future to using `make` commands in CI but definitely not needed for this PR, thanks!

Makefile
ParthivNaresh: Oops slipped through

.github/workflows/unit_tests_with_latest_deps.yml
ParthivNaresh: I think there's a limit to what we can put in certain steps because each step is n individual instance.

bchen1116: Sounds good.

bchen1116 (approved):
LGTM! Thanks for the changes and the great work!

PR #1400, opened 2022-04-25
Add workflow to kickoff Featuretools unit tests with Woodwork main

gsheni:


.github/workflows/kickoff_featuretools_unit_tests.yml
```diff
@@ -0,0 +1,18 @@
+name: Kickoff Featuretools Unit Tests
+
+on:
+  push:
+    branches:
+      - main
```
jeff-hernandez, line 6: It'd be helpful to confirm if this workflow works as expected. I think we could test in this PR by temporarily removing the main branch condition.

gsheni: It works as expected:

1. https://github.com/alteryx/woodwork/pull/1400/commits/4871b297c8c63579238867d6d878939cbebc9bf9
2. https://github.com/alteryx/woodwork/runs/6164715123?check_suite_focus=true
3. https://github.com/alteryx/featuretools/actions/runs/2222652591

jeff-hernandez (approved):
LGTM

PR #1409, opened 2022-04-29
Delete setup.py, MANIFEST.in and move configuration to pyproject.toml

gsheni:
- Delete setup.py and move some configuration to pyproject.toml

> With the release of pip v19.1.1 , editable installs are again allowed in the presence of pyproject.toml. 

- Additionally, deleted MANIFEST.in because of the following:
  - https://packaging.python.org/en/latest/guides/using-manifest-in/#how-files-are-included-in-an-sdist
- following files are included in a source distribution:
  > all files specified by the package_data and data_files setup() arguments 
  > the file specified by the license_file option in setup.cfg (setuptools 40.8.0+)
- deleted codecov because we don't actually use it


ParthivNaresh (commented):
Understood, so the reasoning for these changes is:

Delete setup.py and move some configuration to pyproject.toml

- With the release of pip v19.1.1 , editable installs are again allowed in the presence of pyproject.toml.

MANIFEST.in removed
A certain number of files are automatically included in a source distribution by default, such as:
1. the file specified by the license_file option in setup.cfg
2. README.md
3. all files specified by the package_data
    - We specify *.txt, LICENSE, README.md (should't LICENSE and README.md be covered by 1 and 2)

Delete codecov
Why is this causing a loss of coverage for the `frequency_inference` files?

gsheni:
That is not due to my PR and is a separate issue. Either codecov is being flaky or our unit tests are not properly being run or we don't have proper coverage.

ParthivNaresh:
The most recently merged PR has full coverage for it: https://app.codecov.io/gh/alteryx/woodwork/compare/1410/tree/woodwork/statistics_utils

I'm rerunning the tests to be sure, we don't want to lose coverage for an entire functionality

gsheni:
@ParthivNaresh This is a common issue with codecov, it's being flaky/inaccurate.

ParthivNaresh (approved):
Looks good, I assume codecov is flaking out!

PR #1413, opened 2022-05-03
Fix incorrect datetime inference

### LINKED ISSUE #1411: Woodwork inferring numeric columns with single value as datetime
I would expect the default inference to be "Double" as opposed to datetime in this case.

```python
import pandas as pd
import woodwork as ww

print(ww.__version__)

df = pd.DataFrame({"column": [1951.5] * 10})
df.ww.init()

assert df.ww.logical_types['column'].type_string == "datetime"
```

I believe this was introduced in v 0.15.0 (Datetime inference v2). Cannot reproduce with versions 0.13 or 0.14

image

bchen1116:
fix #1411

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -7,7 +7,7 @@ Future Release
 ==============
     * Enhancements
     * Fixes
-        * Fixed ``col_is_datetime`` inference function to not infer float as datetime (:pr:`1413`)
+        * Fixed ``col_is_datetime`` inference function to not infer numeric dtypes as datetime (:pr:`1413`)
     * Changes
         * Delete setup.py, MANIFEST.in and move configuration to pyproject.toml (:pr:`1409`)
     * Documentation Changes
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -7,6 +7,7 @@ Future Release
 ==============
     * Enhancements
     * Fixes
+        * Fixed datetime inference function (:pr:``)
     * Changes
         * Delete setup.py, MANIFEST.in and move configuration to pyproject.toml (:pr:`1409`)
     * Documentation Changes
@@ -15,7 +16,7 @@ Future Release
         * Add workflow to kickoff Featuretools unit tests with Woodwork main (:pr:`1400`)
 
     Thanks to the following people for contributing to this release:
-    :user:`gsheni`
+    :user:`gsheni`, :user:`bchen1116`
     
 v0.16.2 Apr 25, 2022
 ====================
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -7,7 +7,7 @@ Future Release
 ==============
     * Enhancements
     * Fixes
-        * Fixed datetime inference function (:pr:`1413`)
+        * Fixed ``col_is_datetime`` inference function to not infer float as datetime (:pr:`1413`)
     * Changes
         * Delete setup.py, MANIFEST.in and move configuration to pyproject.toml (:pr:`1409`)
     * Documentation Changes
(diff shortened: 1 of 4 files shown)


woodwork/type_sys/utils.py
```diff
@@ -11,10 +11,14 @@
 
 def col_is_datetime(col, datetime_format=None):
     """Determine if a dataframe column contains datetime values or not. Returns True if column
-    contains datetimes, False if not. Optionally specify the datetime format string for the column."""
+    contains datetimes, False if not. Optionally specify the datetime format string for the column.
+    Will not infer float/double data as datetime."""
```
thehomebrewnerd, line 15: Might want to mention integers here as well since this is being added. As I understand it, this will also not identify integer columns as valid datetime columns.

docs/source/release_notes.rst
```diff
@@ -7,6 +7,7 @@ Future Release
 ==============
     * Enhancements
     * Fixes
+        * Fixed datetime inference function (:pr:`1413`)
```
thehomebrewnerd, line 10: Minor nitpick, but this is actually fixing the `col_is_datetime` function which is used for inference, but also other places. Might be better to say `Fixed col_is_datetime utility function for correct handling of floats` or something similar here for clarity.

woodwork/type_sys/utils.py
bchen1116: Integers will be unchanged here! Just floats will be not converted to datetime.

thehomebrewnerd: Yes, but we are now specifically calling out that float/double will not be inferred as a datetime. This is also true for integers, but that isn't mentioned. For consistency if we mention the behavior for float/double we should also mention it for integers.

bchen1116: Currently, the check that I added was to return `false` if the column is a float dtype. I don't add anything here to do the same for integer, and if an integer can be interpreted as a datetime (through string conversion), we will return true. You can look at the new test I added!

bchen1116: After discussion with @thehomebrewnerd, we decided that all numeric dtypes should be left as numerics rather than inferred as datetimes.

woodwork/type_sys/utils.py
```diff
@@ -22,6 +23,12 @@ def col_is_datetime(col, datetime_format=None):
     if len(col) == 0:
         return False
 
+    try:
+        if pd.api.types.is_numeric_dtype(pd.Series(col.values.tolist())):
+            return False
+    except AttributeError:
+        pass
```
thehomebrewnerd, line 30: I was looking (unsuccessfully) for alternatives to this as it seems a little confusing why this is needed. Our future selves might benefit from a comment here explaining why this try/except block is necessary.

thehomebrewnerd (approved):

woodwork/type_sys/utils.py
```diff
@@ -22,6 +23,17 @@ def col_is_datetime(col, datetime_format=None):
     if len(col) == 0:
         return False
 
+    try:
+        if pd.api.types.is_numeric_dtype(pd.Series(col.values.tolist())):
```
gsheni, line 27: This will be fine computationally? Can't we just do:
```suggestion
        if pd.api.types.is_numeric_dtype(col):
```

bchen1116: Yep, this is in response to an object edge-case:
image

We still want to infer this as numeric, and it turns out that the only way we can remove this dtype casting is by accessing the list of values via `tolist()`
image

gsheni: alright

gsheni (approved):

PR #1414, opened 2022-05-03
Add a workflow for testing Woodwork without test dependencies

gsheni:
- This workflow adds a test to Woodwork to install `core` Woodwork, and use the library


ParthivNaresh (approved):
Looks good thanks for adding this!

PR #1437, opened 2022-05-24
Fixes concat of invalid values

### LINKED ISSUE #1434: Concatenating empty list of invalid values
The default logical type behavior of returning invalid values leads to pandas value error. The base class method ``validate``
 returns ``None`` by default. Logical types derived from the base class override the ``validate`` method if necessary and should return a series when ``return_invalid_values=True``. The table accessor concatenates the returned invalid values across logical types. As a result, this behavior leads to a pandas error because the table accessor tries to concatenate an empty list when all logical types use the inherited validate method and return ``None``.

#### Example

```python
import pandas as pd
import woodwork as ww

df = pd.DataFrame(
    {
        "postal_code": ["77449", "11368", "60629", "79936"],
    }
)

df.ww.init(logical_types={"postal_code": "Unknown"})
df.ww.validate_logical_types(return_invalid_values=True)
```

```
ValueError: No objects to concatenate
```

**Acceptance Criteria (AC)**
- Update ``pd.DataFrame.ww.validate_logical_types`` to only concatenate when the list of invalid values is not empty

jeff-hernandez:
Closes #1434

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -7,6 +7,7 @@ Future Release
 ==============
     * Enhancements
     * Fixes
+        * Fix concatenation of invalid logical type values (:pr:`1437`)
     * Changes
     * Documentation Changes
 	* Add slack icon to footer in docs (:pr:`1432`)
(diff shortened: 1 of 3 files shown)


bchen1116 (approved):
Nice! LGTM

PR #1439, opened 2022-05-24
Fixes validation for numeric postal codes

### LINKED ISSUE #1433: Integer postal code validation
Validating postal code data of integers raises an attribute error. The `PostalCode` logical type has a primary dtype of category, so the actual data can be integer or string. However, the logical type's validation method only supports string data.

#### Example

```python
import pandas as pd
import woodwork as ww

df = pd.DataFrame(
    {
        "postal_code": [77449, 11368, 60629, 79936],
    }
)

df.ww.init(logical_types={"postal_code": "PostalCode"})
df.ww.validate_logical_types()
```
```
AttributeError: Can only use .str accessor with string values!
```

##### Acceptance Criteria (AC)
- Updating `validate` method of `PostalCode` to handle integer types.

jeff-hernandez:
Closes #1433

### DIFF
--- a/woodwork/logical_types.py
+++ b/woodwork/logical_types.py
@@ -696,7 +696,11 @@ class PostalCode(LogicalType):
 
     def transform(self, series):
         if pd.api.types.is_numeric_dtype(series):
-            series = series.astype("Int64").astype("string")
+            try:
+                series = series.astype("Int64").astype("string")
+            except TypeError:
+                raise TypeConversionError(series, "string", type(self))
+
         return super().transform(series)
 
     def validate(self, series, return_invalid_values=False):
--- a/woodwork/logical_types.py
+++ b/woodwork/logical_types.py
@@ -694,6 +694,11 @@ class PostalCode(LogicalType):
     backup_dtype = "string"
     standard_tags = {"category"}
 
+    def transform(self, series):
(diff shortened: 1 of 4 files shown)


woodwork/logical_types.py
```diff
@@ -694,6 +694,11 @@ class PostalCode(LogicalType):
     backup_dtype = "string"
     standard_tags = {"category"}
 
+    def transform(self, series):
+        if pd.api.types.is_numeric_dtype(series):
```
bchen1116, line 698: Curious if we should leave this in a try-catch scenario. If a user accidentally passes in a numeric type float like `12345.5`,  pandas will throw a `TypeError`. Might be better for us to catch that and throw a useful message on our end (and include test for it if so)

bchen1116 (approved):
LGTM! I left one comment about catching non-equivalent floats (ie 12345.5), but the rest looks great!

woodwork/logical_types.py
jeff-hernandez: agreed, good catch!

jeff-hernandez: This would only apply to pandas since dask does not fail until compute and spark allows for float64 to Int64 conversion by dropping out decimal places.

PR #1440, opened 2022-05-26
Testing with Minimum reqs of minimum reqs

bchen1116:
Taking out the `python-dateutils` requirement results in test failures:
image

This shows that it installs `python-dateutils==2.7.3` rather than `2.8.1`, which suggests this works!


setup.cfg
```diff
@@ -44,6 +44,7 @@ install_requires =
     pandas >= 1.3.0, <1.4.2
     scikit-learn >= 0.22
     python-dateutil >= 2.8.1
+    scipy >= 1.4.0
```
bchen1116, line 47: `scipy < 1.4.0` install fails with `scikit-learn == 0.22` to `0.24`. Add this as a requirement to prevent install failures

docs/source/release_notes.rst
```diff
@@ -10,14 +10,15 @@ Future Release
         * Fix concatenation of invalid logical type values (:pr:`1437`)
     * Changes
     * Documentation Changes
-	* Add slack icon to footer in docs (:pr:`1432`)
+	    * Add slack icon to footer in docs (:pr:`1432`)
```
ParthivNaresh, line 13: Extra tab here?

woodwork/tests/requirement_files/minimum_test_requirements.txt
```diff
@@ -7,5 +7,7 @@ pytest-cov==2.10.1
 pytest-xdist==2.1.0
 pytest==7.0.1
 python-dateutil==2.8.1
+six==1.11.0
```
ParthivNaresh, line 10: How do we leverage `six`?

minimum_req.py
```diff
@@ -0,0 +1,159 @@
...
+        _package = pkg_resources.working_set.by_key[_package_name]
+        all_requirements.append(package)
+        reliance = [str(r) for r in _package.requires()]
+        all_requirements.extend(reliance)  # retrieve deps from setup.py
+
+    # for any requirements in our test-requirements that apply to the core requirement packages, add that in as well
+    for package in min_test_reqs:
+        pack = tuple(requirements.parse(package))[0]
+        _package_name = pack.name
+        if any([_package_name in s for s in all_requirements]):
```
ParthivNaresh, line 139: Why do we use `_package_name in s` as opposed to `_package_name == s`? Also just so I understand, why do we append `all_requirements` with a package it already has?

minimum_req.py
```diff
@@ -0,0 +1,159 @@
...
+    path = args.req_file_path
+    
+    # get the minimum requirements and install the min core requirements
+    min_reqs, min_core_reqs, min_test_reqs = get_min_test_and_core_requirements(path + "minimum_core_requirements.txt", path + "minimum_test_requirements.txt")
+    install_min_deps()
+
+    package_to_version_dict = {}
+    all_requirements = []
+
+    # find all packages that the core requirements rely on
```
ParthivNaresh, line 126: There's no relationship between `package` in line 111 which refers to `woodwork` and the package being iterated over here and below right? If that's the case we should probably change the variable name to avoid confusion.

minimum_req.py
```diff
@@ -0,0 +1,159 @@
+import argparse
+import json
+import re
+import subprocess
+import requirements
+from pip._vendor import pkg_resources
+from pkg_resources import packaging
+
+def get_all_package_versions(package_name: str) -> list:
```
ParthivNaresh, line 9: Loving the type hints

ParthivNaresh (commented):
This looks awesome! I love the way this is setup and the comments definitely helped me understand this. I had a few questions but everything looks great so far!

woodwork/tests/requirement_files/minimum_test_requirements.txt
bchen1116: `python-dateutil` relies on `six>1.5`, but a few of our testing packages have requirements of `six>1.9` or `six>=1.11`. In order for our tests to pass, we need to add this requirement

gsheni: You might want to add this as a min requirement then, in setup.cfg

bchen1116: Does `setup.cfg` track files that are required for the package to work, or for the tests to run as well?

bchen1116: Ah I see what you mean. Will add!

jeff-hernandez (commented):
Overall it's looking good!

minimum_req.py
```diff
@@ -0,0 +1,158 @@
...
+    try:
+        sub_output = subprocess.check_output(args).decode('utf-8')
+    except subprocess.CalledProcessError as e:
+        raise RuntimeError("command '{}' return with error (code {}): {}".format(e.cmd, e.returncode, e.output))
+    # this gets the string versions available from the `pip index` command
+    string_versions = [output_split[len(version_substring):].split(", ") 
+                        for output_split in sub_output.split("\n") if version_substring in output_split][0]
+    # we filter out versions that might have letters in it (ie 22.2.post1, etc) to only compare versions that we can convert to float
+    float_versions = list(filter(lambda version_number: not re.search(r'[A-Za-z]', version_number), 
+                          string_versions))
```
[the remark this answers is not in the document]
bchen1116: Nice, this would work!

minimum_req.py
```diff
@@ -0,0 +1,158 @@
...
+            f.write(return_string)
+        f.close()
+    return return_string
+
+def install_min_deps():
+    # handles the install of all min dependencies
+    print("Installing the minimum requirements generated")
+    process = ['pip', 'install']
+    process.extend(min_reqs.split(delim)[:-1])
+    # process = [x for x in process if ('wheel' not in x and 'pip==' not in x)]
```
Cmancuso, line 78: do we need this line?

minimum_req.py
```diff
@@ -0,0 +1,158 @@
...
+            all_requirements.append(package)
+    all_requirements = list(set(all_requirements))
+
+    # iterate through each dependency to determine the minimum version allowed
+    for package_value in all_requirements:
+        req = tuple(requirements.parse(package_value))
+        try:
+            package_name = req[0].name
+            version_logic = get_version_logic(req[0].specs)
+        except IndexError:
```
Cmancuso, line 147: what is this catching?

bchen1116: Goood call! This was left in when I was doing an earlier iteration. CAn remove

setup.cfg
```diff
@@ -78,6 +79,7 @@ test =
     moto[all] >= 3.0.7
     smart-open >= 5.0.0
     pyarrow >= 4.0.1
+    six==1.11.0
```
gsheni, line 82:
```suggestion
    six>=1.11.0
```

Cmancuso (approved):
Looks good!

PR #1449, opened 2022-06-21
Null invalid values for Double

jeff-hernandez:


setup.cfg
```diff
@@ -85,7 +85,7 @@ dask =
     dask[dataframe] >= 2021.10.0
 
 spark = 
-    pyspark >= 3.2.0
+    pyspark >= 3.2.0, <3.3.0
```
jeff-hernandez, line 88: the latest version of pyspark contains breaking changes

bchen1116: What breaks with the latest version? Should we also file an issue to unpin this in the future?

woodwork/table_accessor.py
```diff
@@ -521,6 +531,7 @@ def set_types(self, logical_types=None, semantic_tags=None, retain_index_tags=Tr
             retain_index_tags (bool, optional): If True, will retain any index or time_index
                 semantic tags set on the column. If False, will replace all semantic tags any time a column's
                 semantic tags or logical type changes. Defaults to True.
+            null_invalid_values (bool, optional): If True, replaces any invalid values with null.
```
bchen1116, line 534: Add 'Defaults to False.'

bchen1116 (approved):
Looks good to me! Left a few questions, but great work on this.

setup.cfg
jeff-hernandez: One unit test breaks:
- `test_series_methods_returning_frame_no_name[sample_series_spark]` raises a `KeyError`
I'll file an issue to investigate further. At an initial glance, it seems like a problem with having `Series.ww.name` set as None.

jeff-hernandez: Tracking issue in https://github.com/alteryx/woodwork/issues/1453

woodwork/logical_types.py
```diff
@@ -788,3 +793,15 @@ def _validate_not_negative(series, return_invalid_values):
         if any_invalid:
             info = f"Series {series.name} contains negative values."
             raise TypeValidationError(info)
+
+
+def _coerce_string(series):
```
ParthivNaresh, line 798: What's the purpose of casting to a `string` before coercing to `numeric`? Is it to support boolean transformations?

jeff-hernandez: `pandas.to_numeric` doesn't work properly and raises an error with object types

ParthivNaresh (approved):
Looks great!

PR #1450, opened 2022-06-23
Restrict pyspark below 3.3.0 

gsheni:


setup.cfg
```diff
@@ -42,6 +42,7 @@ packages = find:
 include_package_data = True
 install_requires =
     pandas >= 1.3.0, <1.4.2
+    numpy >= 1.21.0
```
thehomebrewnerd, line 45: I think we can remove this now

thehomebrewnerd (approved):

PR #1456, opened 2022-06-27
Null invalid integers

jeff-hernandez:


woodwork/logical_types.py
```diff
@@ -389,6 +389,11 @@ class IntegerNullable(LogicalType):
     primary_dtype = "Int64"
     standard_tags = {"numeric"}
 
+    def transform(self, series, null_invalid_values=False):
+        if null_invalid_values:
```
ParthivNaresh, line 393: Could you add a quick docstring to this describing what's going on here (coersion to string, numeric, and then nulling out floats)

woodwork/logical_types.py
```diff
@@ -800,6 +805,10 @@ def _validate_not_negative(series, return_invalid_values):
             raise TypeValidationError(info)
 
 
+def _get_index_invalid_integer(series):
+    return series.mod(1).ne(0)
```
ParthivNaresh, line 809: Nice!

ParthivNaresh (approved):
Looks great, just a quick docstring addition if you could please!

PR #1457, opened 2022-06-28
Null invalid emails

jeff-hernandez:


Cmancuso (approved):
LGTM! For all of these, transform just needs to be called first in tempo?

jeff-hernandez:
@Cmancuso 
> For all of these, transform just needs to be called first?

Yes, the transform gets called during woodwork initialization (`pd.DataFrame.ww.init()`).

Issue #1172, opened 2021-10-22
Improve Inference to support Boolean/IntegerNullable


chukarsten:
- As a [user/developer], I wish I could use Woodwork to properly infer the new Nullable types.

#### Code Example
A user might expect the following tests to pass:

```python
import numpy
import pandas

from woodwork.logical_types import BooleanNullable, IntegerNullable, Unknown

missing_types = [None, numpy.nan, pandas.NA]
col_lengths = [0,1,5,100]
booleans = [True for i in range(1, max(col_lengths))]
integers = [i for i in range(1, max(col_lengths))]

@pytest.mark.parametrize("missing_type", missing_types)
@pytest.mark.parametrize("col_length", col_lengths)
@pytest.mark.parametrize("values_and_expected_inferred_type", [(booleans, BooleanNullable), (integers, IntegerNullable)])
def test_ww_init_nullable(missing_type, col_length, values_and_expected_inferred_type):
    values, expected_inferred_type = values_and_expected_inferred_type
    # Build a DataFrame with a column that has a single missing type and the rest of the expected types
    test_df = pandas.DataFrame({"nullable_column":[missing_type] + values[:col_length-1]})
    test_df.ww.init()
    if col_length != 0:
        # Expect any column with a nullable and an integer or boolean to be the proper Pandas nullable type.
        assert isinstance(test_df.ww.logical_types["nullable_column"], expected_inferred_type)
    else:
        # Expect a column with only null types to be Unknown.  Probably redundant.
        assert isinstance(test_df.ww.logical_types["nullable_column"], Unknown)

```

I know that some inference is done not only be looking at the types of the columns, but also a statistical analysis of the presence of those types.  This test might or might not need to be extended to cover those cases, but I think the mere presence of a nullable type amidst integers or bool should be enough.

Issue #1038, opened 2021-07-07
Series full of `'nan'` strings inferred as `Datetime` logical type


tamargrey:
If a Series is passed to Woodwork that contains only `"nan"` strings, it gets inferred as `Datetime` logical type, filling the values with `NaT`.

```python
  nans = pd.DataFrame({
        "nans":pd.Series(['nan', 'nan', 'nan', 'nan', 'nan']),
        "nan_strings":pd.Series(['nan', 'nan', 'nan', 'nan', 'nan'], dtype='string'),
        "nan_objs":pd.Series(['nan', 'nan', 'nan', 'nan', 'nan'], dtype='object'),
       })
  nans.ww.init()

assert nans.ww.logical_types['nans'] == ww.logical_types.Datetime()
assert nans.ww.logical_types['nan_strings'] == ww.logical_types.Datetime()
assert nans.ww.logical_types['nan_objs'] == ww.logical_types.Datetime()

```

Here's the DataFrame and its dtypes before and after initialization:
![image](https://user-images.githubusercontent.com/64278226/124778382-f2255080-df0e-11eb-9541-39ee2266725a.png)

`Datetime` is definitely the incorrect logical type for these columns. What the correct logical type actually is may be up for discussion, but based off of the following behavior with `np.nan` and `pd.NA` inference, `Categorical` may be a good option that will allow for the most flexibility if values get added to the column in the future:
```python
nans = pd.DataFrame({
                            "np_nan":pd.Series([np.nan, np.nan, np.nan, np.nan]),
                            "pd_na":pd.Series([pd.NA, pd.NA, pd.NA, pd.NA]),
                            "np_nan_objs":pd.Series([np.nan, np.nan, np.nan, np.nan], dtype='object'),
                            "pd_na_objs":pd.Series([pd.NA, pd.NA, pd.NA, pd.NA], dtype='object'),
                           })

nans.ww.init()

nans.ww
```
![image](https://user-images.githubusercontent.com/64278226/124779326-afb04380-df0f-11eb-9948-84e7a9d7a976.png)

thehomebrewnerd:
@tamargrey It seems to me that the logical type of a column with all missing values is ambiguous. Do you think it would be better to return the logical type a `Unknown` in this case (when that PR is implemented), rather than `Categorical` or some other logical type?

tamargrey:
[...]

PR #1458, opened 2022-06-29
Better inference for Integer/Boolean Nullables

ParthivNaresh:
Fixes: #1172, #1038 

Performance tests are [here](https://alteryx.atlassian.net/wiki/spaces/PS/pages/1410727994/Performance+Tests+-+Expanding+nan+Inference+for+IntegerNullable+BooleanNullable)


woodwork/type_sys/inference_functions.py
```diff
@@ -52,6 +53,9 @@ def integer_nullable_func(series):
             return not _is_categorical_series(series, threshold)
         else:
             return True
+    elif pdtypes.is_float_dtype(series.dtype):
```
Cmancuso, line 56: How do we handle specificity. So if one column were to parse 80% successful as integer and 90% successful as float, which would we pick?

woodwork/column_accessor.py
```diff
@@ -103,7 +104,10 @@ def init(
                     logical_type.validate(self._series)
                 else:
                     valid_dtype = logical_type._get_valid_dtype(type(self._series))
-                    if valid_dtype != str(self._series.dtype):
+                    if valid_dtype != str(self._series.dtype) and not (
+                        pdtypes.is_integer_dtype(valid_dtype)
+                        and pdtypes.is_float_dtype(self._series.dtype)
+                    ):
```
ParthivNaresh, line 110: Needed because `nans` will currently render an otherwise `integer` dtype as a `float`. Conversion to `Int64` dtype then yields an error

woodwork/tests/type_system/test_ltype_inference.py
```diff
@@ -120,7 +120,9 @@ def test_categorical_inference(categories):
     dtypes = ["object", "string", "category"]
     if _is_spark_series(categories[0]):
         dtypes = get_spark_dtypes(dtypes)
-    for series in categories:
+    for ind, series in enumerate(categories):
+        if ind == len(categories) - 1:
+            dtypes = ["string", "category"]
```
ParthivNaresh, line 125: Casting an all integer series as an `object` and then attempting to have it be inferred as categorical will not work anymore, as the nan replacement will make it an `int` dtype. This will still work for `string` and `categorical` casting.

woodwork/tests/type_system/test_ltype_inference.py
```diff
@@ -233,10 +235,12 @@ def test_unknown_inference(strings):
 
 def test_unknown_inference_all_null(nulls):
     dtypes = ["object", "string", "category", "datetime64[ns]"]
-    if _is_spark_series(nulls[0]):
-        dtypes = get_spark_dtypes(dtypes)
 
-    for series in nulls:
+    for ind, series in enumerate(nulls):
+        if ind == len(nulls) - 1:
+            dtypes = ["object", "string", "category"]
```
ParthivNaresh, line 241: The result of the fix for #1038

woodwork/type_sys/inference_functions.py
ParthivNaresh: Yep good question, here's how the logic is determined for which type is picked in cases of multiple matches: https://woodwork.alteryx.com/en/latest/guides/custom_types_and_type_inference.html#Logical-Type-Relationships

woodwork/logical_types.py
```diff
@@ -820,6 +825,26 @@ def _regex_validate(regex_key, series, return_invalid_values):
...
+def _replace_nans(series):
+    """
+    Replaces empty string values, string representations of NaN values ("nan", "<NA>"), and NaN equivalents
+    with np.nan or pd.NA depending on column dtype.
+    """
+    original_dtype = series.dtype
+    if str(original_dtype) == "string":
+        series = series.replace(ww.config.get_option("nan_values"), pd.NA)
+        return series
+    series = series.apply(
```
gsheni, line 837: Can we run some performance tests with and without this change? I have some concerns that it might lead to slower init in Woodwork

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -2980,7 +2966,7 @@ def test_infer_missing_logical_types_force_infer(sample_df):
     parsed_logical_types = _infer_missing_logical_types(
         sample_df, force_logical_types, existing_logical_types
     )
-    assert parsed_logical_types["age"] == Double()
+    assert parsed_logical_types["age"] == IntegerNullable()
```
thehomebrewnerd, line 2969: Should we be asserting with `isinstance` here instead of checking object equality?

woodwork/tests/type_system/test_custom_types.py
```diff
@@ -57,7 +57,7 @@ def new_double_func(series):
     )
     ww.type_system.update_inference_function("Integer", inference_function=None)
     sample_df.ww.init()
-    assert isinstance(sample_df.ww["age"].ww.logical_type, Double)
-    assert sample_df["age"].dtype == "float64"
+    assert isinstance(sample_df.ww["age"].ww.logical_type, IntegerNullable)
```
thehomebrewnerd, line 60: I think this test is broken. I believe the original intent of this test was to confirm that we can replace the default inference function and that it gets used as expected. Before we were replacing the Double inference function and checking that the logical type was identified as Double. Now we are replacing the Double function, but asserting that the type is IntegerNullable.

There might be a better way to go about this, but maybe think through this and make sure the test is still testing what we want.

woodwork/tests/accessor/test_statistics.py
```diff
@@ -1290,37 +1282,17 @@ def test_value_counts(categorical_df):
             )
         expected_cat1 = updated_results
 
-        updated_results = []
```
thehomebrewnerd, line 1293: Nice cleanup here!

woodwork/tests/logical_types/test_logical_types.py
```diff
@@ -659,3 +664,42 @@ def test_null_invalid_postal_code_numeric():
     expected = pd.DataFrame({"data": data})
     df.ww.init(logical_types=types, null_invalid_values=True)
     pd.testing.assert_frame_equal(df, expected)
+
+
+@pytest.mark.parametrize("null_type", [None, pd.NaT, np.nan, "null", "N/A", "mix", 5])
```
chukarsten, line 669: Does `pd.NaT` cover `pd.NA`?

woodwork/tests/logical_types/test_logical_types.py
```diff
@@ -659,3 +664,42 @@ def test_null_invalid_postal_code_numeric():
...
+@pytest.mark.parametrize("data_type", [int, float])
+def test_integer_nullable(data_type, null_type):
+    nullable_nums = pd.DataFrame(
+        np.random.choice([data_type(i) for i in range(10)], 100), columns=["num_nulls"]
```
chukarsten, line 673: Always iffy about introducing randomness into unit tests....I think you're probably better off just building a dataframe with everything you want right off the bat.

woodwork/tests/logical_types/test_logical_types.py
```diff
@@ -659,3 +664,42 @@ def test_null_invalid_postal_code_numeric():
...
+    nullable_nums.ww.init()
+
+    if null_type != 5:
+        assert isinstance(nullable_nums.ww.logical_types["num_nulls"], IntegerNullable)
+        assert all(nullable_nums["num_nulls"][-5:].isna())
+    elif data_type is int:
+        assert isinstance(nullable_nums.ww.logical_types["num_nulls"], Integer)
+    else:
+        assert isinstance(nullable_nums.ww.logical_types["num_nulls"], Double)
+
```
chukarsten, line 687: What happens when we mix integers and booleans together?  What would we expect to happen?

chukarsten (commented):
I think this is a solid step forward.  I left mostly commentary on the testing as that's the thing I'm primarily concerned with moving forward with nullable types in EvalML.  I think this addresses the base concern, however, of columns that have either 1.) some boolean and nulls of any type or 2.) some integers and nulls of any type and woodwork resolving both of these cases into the proper nullable type.  Big time helpful, @ParthivNaresh , thank you.  I'll try and come back and address the impl later...

woodwork/tests/type_system/test_custom_types.py
ParthivNaresh: Great point, I updated the test to override the default `IntegerNullable` function instead and check that the logical type is identified as `Double`.

woodwork/tests/logical_types/test_logical_types.py
```diff
@@ -659,3 +664,42 @@ def test_null_invalid_postal_code_numeric():
...
+def test_integer_nullable(data_type, null_type):
+    nullable_nums = pd.DataFrame(
+        np.random.choice([data_type(i) for i in range(10)], 100), columns=["num_nulls"]
+    )
+    nullable_nums["num_nulls"].iloc[-5:] = (
+        [None, pd.NA, np.nan, "NA", "none"] if null_type == "mix" else [null_type] * 5
+    )
+    nullable_nums.ww.init()
+
+    if null_type != 5:
```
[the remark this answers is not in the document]
ParthivNaresh: It was a random integer test to make sure that `pdtypes.is_float_dtype(series.dtype)` which was added to the `IntegerNullable` inference function doesn't attempt to infer all integers which have been cast to `float` as an `Integer` or `IntegerNullable`

woodwork/tests/logical_types/test_logical_types.py
ParthivNaresh: If it's just integers and booleans, we would expect the same behaviour which is an inference of `Categorical` based on the `dtype: object`. If `nan` values are added it would be the same.

woodwork/logical_types.py
ParthivNaresh: I've run performance tests [here](https://alteryx.atlassian.net/wiki/spaces/PS/pages/1410727994/Performance+Tests+-+Expanding+nan+Inference+for+IntegerNullable+BooleanNullable). There's a significant increase in inference runtime when this line is added and I'm fine with removing it

gsheni: Yah I agree, let's remove casting, and let pandas handle it internally

woodwork/table_accessor.py
```diff
@@ -1615,7 +1615,12 @@ def _infer_missing_logical_types(
             null_invalid_values=null_invalid_values,
         )
         if updated_series is not series:
-            dataframe[name] = updated_series
+            # NotImplementedError thrown by dask when attempting to re-initialize
+            # data after being assigned a numeric column name
+            try:
+                dataframe[name] = updated_series
+            except NotImplementedError:
+                pass
```
bchen1116, line 1623: can we get around this by upgrading dask or something? Seems weird to just `pass` here

bchen1116 (commented):
Left some quick comments, will finish reviewing tomorrow

woodwork/tests/logical_types/test_logical_types.py
Cmancuso: Same thing with only 1's and 0's? That's a case we'll want to be inferred as Boolean at some point.

gsheni:
@ParthivNaresh what was the change in inference after removing the apply?

woodwork/table_accessor.py
ParthivNaresh: I couldn't find a version of dask that satisfies this issue unfortunately. We can't assign a numeric column name in this specific case for dask, but I'm open to replacing it with something else though. It doesn't seem like something that we should raise an error for imo

ParthivNaresh:
> @ParthivNaresh what was the change in inference after removing the apply?

It looks like `pandas` is able to handle `pd.NA` without a problem, all tests are still passing and expected logical types are still showing up

ParthivNaresh:
> @ParthivNaresh what was the change in inference speed/time after removing the apply?

The change should be seen [here](https://alteryx.atlassian.net/wiki/spaces/PS/pages/1410727994/Performance+Tests+-+Expanding+nan+Inference+for+IntegerNullable+BooleanNullable)

woodwork/type_sys/inference_functions.py
```diff
@@ -41,7 +41,7 @@ def categorical_func(series):
 
 def integer_func(series):
     if integer_nullable_func(series) and not series.isnull().any():
-        return True
+        return all(series.mod(1).eq(0))
```
bchen1116, line 44: nice!

woodwork/type_sys/inference_functions.py
```diff
@@ -78,6 +83,20 @@ def boolean_nullable_func(series):
...
+            series_no_null_unq = set(series_no_null)
+            if series_no_null_unq in [
+                {False, True},
+                {True},
+                {False},
+            ]:
+                return True
+        except TypeError as te:  # Necessary to check for non-hashable values because of object dtype consideration
+            te_msg = str(te)
+            if "unhashable type" in te_msg:
```
bchen1116, line 98: Do we technically not need this? It seems like either way, if we hit TypeError, we return False

woodwork/logical_types.py
```diff
@@ -830,6 +835,23 @@ def _regex_validate(regex_key, series, return_invalid_values):
             raise TypeValidationError(info)
 
 
+def _replace_nans(series):
```
bchen1116, line 838: would be nice to add some type-hinting here jjust to start getting it more incorporated into WW

bchen1116 (approved):
This looks pretty good to me! Left a few comments on improvements with understanding/documentation, but good to go after that. Great work here!

woodwork/table_accessor.py
```diff
@@ -1599,12 +1599,12 @@ def _infer_missing_logical_types(
     existing_logical_types = existing_logical_types or {}
     parsed_logical_types = {}
     for name in dataframe.columns:
-        series = dataframe[name]
         logical_type = (
             force_logical_types.get(name)
             if name in force_logical_types
             else existing_logical_types.get(name)
         )
+        series = dataframe[name]
```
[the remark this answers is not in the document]
ParthivNaresh: Just readability!

woodwork/logical_types.py
```diff
@@ -830,6 +835,23 @@ def _regex_validate(regex_key, series, return_invalid_values):
...
+    original_dtype = series.dtype
+    if str(original_dtype) == "string":
+        series = series.replace(ww.config.get_option("nan_values"), pd.NA)
+        return series
+    if not _is_spark_series(series):
+        series = series.replace(ww.config.get_option("nan_values"), np.nan)
+    if str(original_dtype) == "boolean":
+        series = series.astype(original_dtype)
+
+    return series
```
jeff-hernandez, line 852: I think there's already a function that does this. Is [_replace_nan_strings](https://github.com/alteryx/woodwork/blob/5e9b742ec8cc060807323d4c0b5f6daacf6d9659/woodwork/utils.py#L606) useful in this case?

ParthivNaresh: That function was removed in favor of `_replace_nans`

ParthivNaresh: It relied more on regex and string replacement as opposed to overall nan representation replacement

PR #1459, opened 2022-06-29
Null invalid URL

jeff-hernandez:


bchen1116 (approved):
Looks good to me! Should we not raise an error when someone tries to specify a column as URL but it can't be cast as that properly? We do so for numeric types, so could be good to do for parity

jeff-hernandez:
@bchen1116 
> Should we not raise an error when someone tries to specify a column as URL but it can't be cast as that properly? We do so for numeric types, so could be good to do for parity

I think string-based logical types can typecast any dtype to string, whereas numeric cannot typecast string to numeric in pandas. We currently don't raise errors if a column contains invalid strings based on the regex during ww init. Users can call `pd.DataFrame.ww.validate_logical_types` to check for invalid values and raise an error if necessary.

bchen1116:
@jeff-hernandez that makes sense to me. LGTM

PR #1465, opened 2022-07-05
Null invalid LatLong

jeff-hernandez:


woodwork/utils.py
```diff
@@ -620,3 +622,15 @@ def _replace_nan_strings(df: pd.DataFrame) -> pd.DataFrame:
...
+def _parse_latlong(latlong):
+    nan_values = [x for x in ww.config.get_option("nan_values") if len(x)]
+    nan_values = "|".join(nan_values)
+
+    latlong = re.sub(nan_values, "None", latlong)
+
+    try:
+        return ast.literal_eval(latlong)
+    except ValueError:
+        pass
```
Cmancuso, line 636: should we be handling this ?

woodwork/logical_types.py
```diff
@@ -852,6 +856,10 @@ def _get_index_invalid_age(series):
     return series.lt(0)
 
 
+def _get_index_invalid_latlong(series):
+    return ~series.apply(_is_valid_latlong_value)
```
Cmancuso, line 860: what is the bitwise operator doing here?

woodwork/utils.py
jeff-hernandez: This code was extracted from `_reformat_to_latlong`. I think the intent is to try and parse a string to return latlong if successful otherwise return None.

woodwork/logical_types.py
```diff
@@ -15,6 +15,7 @@
 from woodwork.utils import (
     _infer_datetime_format,
     _is_valid_latlong_series,
+    _is_valid_latlong_value,
```
Cmancuso, line 18: NAB , is this an anti-pattern? Importing a "private" method?

jeff-hernandez: Yes, I believe so. It happens a lot throughout the codebase.

PR #1467, opened 2022-07-06
Null invalid PostalCode

jeff-hernandez:


docs/source/release_notes.rst
```diff
@@ -13,6 +13,7 @@ Future Release
         * Added ability to null invalid values for ``URL`` logical type (:pr:`1459`)
         * Added ability to null invalid values for ``PhoneNumber`` logical type (:pr:`1460`)
         * Added ability to null invalid values for ``AgeFractional`` and ``AgeNullable`` logical types (:pr:`1462`)
+        * Added ability to null invalid values for ``PostalCode`` logical type (:pr:`1467`)
```
gsheni, line 16: We might want to be clear this is only for validation for US Postal Codes

woodwork/logical_types.py
```diff
@@ -860,7 +863,7 @@ def _coerce_string(series, regex=None):
         invalid = _get_index_invalid_string(series, regex)
 
         if invalid.any():
-            series[invalid] = None
+            series[invalid] = pd.NA
```
ParthivNaresh, line 866: Why `pd.NA` here?

thehomebrewnerd: @jeff-hernandez That change in behavior is a little surprising to me. Does that mean a pandas 1.4.3 series can contain a mix of None and pd.NA values while still using one of the new nullable dtypes like `string`?

jeff-hernandez: @thehomebrewnerd yes but the series needs to be an object dtype in order to contain both `None` and `pd.NA`. Assigning `None` to a string series will still convert to `pd.NA` in 1.4.3.

jeff-hernandez: @thehomebrewnerd it looks like a series with a string dtype can also contain a mix of `None` and `pd.NA`!

```python
>>> import pandas as pd
>>> x = pd.Series(['90210', '60018-0123', '10010', 'text', '6.7', "<class 'object'>", '123456'], dtype="string")
>>> invalid = pd.Series([False, False, False, True, True, True, True])
>>> x[invalid] = None
>>> x[6] = pd.NA
>>> x
0         90210
1    60018-0123
2         10010
3          None
4          None
5          None
6          <NA>
dtype: string
```

jeff-hernandez: @thehomebrewnerd I agree this inconsistency seems like a pandas bug. I can file an issue.

jeff-hernandez: file pandas issue [47628](https://github.com/pandas-dev/pandas/issues/47628)

ParthivNaresh (approved):
Looks sold to me! I'm a little concerned about Panda's inconsistency in null type handling and it feels like we're constantly making small changes to support, handle, assign, compare, and convert `pd.NA`. Its ambiguity is something we have to keep an eye on.

PR #1470, opened 2022-07-12
Add pre-commit hooks for linting

gsheni:
- This makes it so that linting is automatically run right before you commit code:
- We use this https://pre-commit.com/
- Example of this:
Screen Shot 2022-07-11 at 5 30 10 PM


woodwork/accessor_utils.py
```diff
@@ -209,7 +209,7 @@ def wrapper(self, *args, **kwargs):
             )
             raise WoodworkNotInitError(msg)
         diff_cols = set(self._dataframe.columns).difference(
-            set(self._schema.columns.keys())
+            set(self._schema.columns.keys()),
```
bchen1116, line 212: Is this comma notation part of py3.10?

bchen1116 (approved):
LGTM! Left a question about the comma behaviors but not blocking

woodwork/accessor_utils.py
gsheni: Nope but I added this for this reason:
https://github.com/asottile/add-trailing-comma#multi-line-method-invocation-style----why

PR #1476, opened 2022-07-18
Allow for manual kickoff for minimum dependency checker

gsheni:


.github/workflows/minimum_dependency_checker.yaml
```diff
@@ -5,6 +5,7 @@ on:
       - main
     paths:
       - 'setup.cfg'
+  workflow_dispatch:
```
bchen1116, line 8: where does this show up? How do we kick it off?

gsheni: It will show up in the actions tab. here is an example of an existing:
https://github.com/alteryx/woodwork/actions/workflows/latest_dependency_checker.yaml

gsheni: and here is Minimum Dependency Checker in Featuretools Action
https://github.com/alteryx/featuretools/actions/workflows/minimum_dependency_checker.yaml

PR #1485, opened 2022-08-04
concat_columns handles DataFrames with differing rows

chukarsten:
- When `woodwork.concat_columns` is called on two pandas DataFrames with different numbers of rows or mismatched indices, this causes the underlying library, be it pandas, dask or spark to insert null values into the data where there is no feature support.  This can cause the schema to be invalidated, requiring re-inference.

### DIFF
--- a/woodwork/tests/utils/test_concat.py
+++ b/woodwork/tests/utils/test_concat.py
@@ -4,7 +4,7 @@ import pandas as pd
 import pytest
 
 import woodwork as ww
-import woodwork.exceptions
+import woodwork.logical_types
 from woodwork.accessor_utils import _is_spark_dataframe
 from woodwork.logical_types import Categorical, Double, Integer
 from woodwork.tests.testing_utils import to_pandas
@@ -532,7 +532,17 @@ def test_concat_with_none(sample_df):
         concat_columns([df1, df2, None])
 
 
-def test_concat_shorter_null_int(sample_df):
+@pytest.mark.parametrize(
+    "nullable_type",
+    [
+        "integer",
+        pytest.param(
+            "boolean",
+            marks=pytest.mark.xfail(reason="Bug with BooleanNullable inference."),
+        ),
+    ],
+)
+def test_concat_shorter_null_int(sample_df, nullable_type):
     """Tests that concatenating dataframes with different numbers of rows works."""
     sample_df.ww.init()
     df1 = sample_df.ww[
@@ -547,7 +557,7 @@ def test_concat_shorter_null_int(sample_df):
     try:
         df2 = sample_df.ww[
             [
-                "integer",
+                nullable_type,
                 "categorical",
             ]
         ].iloc[1:, :]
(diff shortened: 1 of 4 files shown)


woodwork/utils.py
```diff
@@ -515,6 +515,20 @@ def concat_columns(objs, validate_schema=True):
...
+    # or mismatched indices.
+    null_cols = combined_df.isnull().any()[combined_df.isnull().any()].index
+    if "dask" not in lib.__name__:
+        null_cols = null_cols.to_numpy()
+    else:
+        null_cols = list(null_cols)
+    for null_col in null_cols:
+        if null_col in logical_types and isinstance(
+            logical_types[null_col],
+            ww.logical_types.Integer,
```
chukarsten, line 528: I expect this needs to be extended to Boolean/BooleanNullable

Cmancuso: We don't want to include that here?

chukarsten: We can't yet because of https://github.com/alteryx/woodwork/issues/1486

woodwork/tests/utils/test_concat.py
```diff
@@ -540,3 +530,51 @@ def test_concat_with_none(sample_df):
...
+    result = concat_columns([df1, df2])
+    expected_type = (
+        woodwork.logical_types.IntegerNullable
+        if nullable_type == "integer"
+        else woodwork.logical_types.BooleanNullable
+    )
+    assert isinstance(result.ww.logical_types[nullable_type], expected_type)
+    assert (
+        result.shape[0] == df1.shape[0]
+    ), "The output of concat_columns has the incorrect number of rows."
```
Cmancuso, line 577: these assertions are the that results are incorrect?

Cmancuso:
This LGTM, I'd have @jeff-hernandez or @ParthivNaresh give it a once over since they might be more familiar with these changes, but should be good to go. 👍

woodwork/utils.py
ParthivNaresh: Working on that now

ParthivNaresh (approved):
Looks solid to me, thanks for adding this!

woodwork/utils.py
```diff
@@ -515,6 +515,20 @@ def concat_columns(objs, validate_schema=True):
 
     combined_df = lib.concat(objs, axis=1, join="outer")
 
+    # The lib.concat breaks the woodwork schema for dataframes with different shapes
+    # or mismatched indices.
+    null_cols = combined_df.isnull().any()[combined_df.isnull().any()].index
```
thehomebrewnerd, line 520: Can you do something like `combined_df.columns[combined_df.isnull().any()]` here instead of using duplicate `.isnull().any()` calls?

thehomebrewnerd: Or save the result of the calls into a variable so you don't have to repeat the operation.

```
mask = combined_df.isnull().any()
null_cols = mask[mask].index
```

chukarsten: As you command

thehomebrewnerd (approved):
LGTM!

PR #1487, opened 2022-08-04
Fixed treatment of Boolean coercion for boolean column with nulls

### LINKED ISSUE #1486: Woodwork Incorrectly Infers Boolean
I would expect the following test to pass.  We're seeing within `concat_columns` that when a DataFrame with a column with mixed null/integers is passed the `Integer` logical type during inference, the `init` fails.  This is expected and an MR was put up to make `concat_columns` resilient to this.  When we extended the test to cover Boolean/BooleanNullable, it was discovered that the `init` will impute the missing boolean value rather than error out that there was an attempted coercion to a non-nullable type.

I would expect that the following test would pass and also be extendable to Integer/IntegerNullable (and float64/Float64 when they're a thing).

```python
import pytest
import numpy as np
@pytest.mark.parametrize("none_type", [None, np.nan, pd.NA])
@pytest.mark.parametrize("pass_logical_types", [True, False])
def test_boolean_inference(none_type, pass_logical_types):
    df = pd.DataFrame({"boolean": [none_type, True, False, True]})
    if pass_logical_types:
        with pytest.raises(Exception):
            # Would expect init to fail as you're trying to coerce a boolean to bool.
            df.ww.init(logical_types = {"boolean": Boolean})
    else:
        df.ww.init()
        assert isinstance(df.ww.logical_types["boolean"], BooleanNullable)

```

ParthivNaresh:
Fixes #1486


ParthivNaresh:
@chukarsten Is this along the lines of the change expected?

woodwork/logical_types.py
```diff
@@ -201,6 +201,15 @@ class Boolean(LogicalType):
...
     primary_dtype = "bool"
 
+    def transform(self, series, null_invalid_values=False):
+        if str(series.dtype) == "object":
+            series = _coerce_boolean(series)
+            if not _is_dask_series(
+                series
+            ):  # Dask cannot handle NA as an ambiguous type for booleans
+                series = series.replace(ww.config.get_option("nan_values"), pd.NA)
+        return super().transform(series)
```
jeff-hernandez, line 211: We can possibly refactor this to use the pandas method `convert_dtype`

```python
        if str(series.dtype) == "object" and isinstance(series, pd.Series):
            series = series.convert_dtypes()
        return super().transform(series)
```

jeff-hernandez (commented):
Overall looking good!

woodwork/logical_types.py
ParthivNaresh: I'll look into this thanks man!

bchen1116: Is this change going in @ParthivNaresh ?

PR #1502, opened 2022-08-19
Check dtype change before using `_replace_nans`

ParthivNaresh:
Addresses performance regression with datasets containing many `Double`, `BooleanNullable`, and `IntegerNullable` columns that end up calling `_replace_nans` many times.

Perf tests [here](https://alteryx.atlassian.net/wiki/spaces/PS/pages/1471611921/Replacing+nans+in+Woodwork+optimization+Performance+Tests).

While this doesn't address the entirety of the performance regressions, it's a good first step at eliminating redundant calls and brings the total run time down by a tremendous amount.


woodwork/logical_types.py
```diff
@@ -216,7 +217,7 @@ class BooleanNullable(LogicalType):
     primary_dtype = "boolean"
 
     def transform(self, series, null_invalid_values=False):
-        series = _replace_nans(series)
+        series = _replace_nans(series, BooleanNullable.primary_dtype)
```
thehomebrewnerd, line 220: Is there a reason for using `BooleanNullable.primary_dtype` instead of `self.primary_dtype` here and in the other methods?

thehomebrewnerd (approved):
Looks good!

woodwork/logical_types.py
```diff
@@ -835,12 +836,14 @@ def _regex_validate(regex_key, series, return_invalid_values):
...
 
 
-def _replace_nans(series: pd.Series) -> pd.Series:
+def _replace_nans(series: pd.Series, primary_dtype: Optional[str] = None) -> pd.Series:
     """
     Replaces empty string values, string representations of NaN values ("nan", "<NA>"), and NaN equivalents
     with np.nan or pd.NA depending on column dtype.
     """
     original_dtype = series.dtype
+    if primary_dtype == str(original_dtype):
```
bchen1116, line 845: can we add tests that check this scenario?

woodwork/logical_types.py
ParthivNaresh: It was just preference but I like `self.primary_type` better, thanks!

PR #1504, opened 2022-08-24
Allow selective transformation

ParthivNaresh:
`ignore_columns` has been added as an argument when initializing a dataframe with woodwork. Any column name(s) that are passed in to `df.ww.init(ignore_columns=[col_1, col_2, etc])` will be ignored during initialization, as long as they are included as part of an existing schema and aren't being forced to a logical type via the `logical_types` argument. This will cause all columns included in `ignore_columns` to not be transformed.


jeff-hernandez (approved):
LGTM, great work!

bchen1116 (approved):
Left one question but looks great to me! Thanks for getting this fix up!

PR #1507, opened 2022-08-26
Fix LatLong.transform on empty dask data

rwedge:
The `LatLong.transform` method had an issue when trying to transform an empty dask series.

The `meta` parameter for the dask `Series.apply` can be a tuple but if so it should be `(name, dtype)`.  The code was using `(series, dtype)`.

I believe this also removes the need to set `series.name` after the apply

### DIFF
--- a/woodwork/tests/logical_types/test_logical_types.py
+++ b/woodwork/tests/logical_types/test_logical_types.py
@@ -155,6 +155,25 @@ def test_latlong_transform(latlong_df):
         pd.testing.assert_series_equal(actual, expected)
 
 
+def test_latlong_transform_empty_series(empty_latlong_df):
+    df_type = str(type(empty_latlong_df))
+    dask = "dask" in df_type
+    spark = "spark" in df_type
+
+    latlong = LatLong()
+    series = empty_latlong_df['latlong']
+    actual = latlong.transform(series)
+
+    if dask:
+        actual = actual.compute()
+    elif spark:
+        actual = actual.to_pandas()
+
+    assert actual.empty
+    assert actual.name == "latlong"
+    assert actual.dtype == latlong.primary_dtype
+
+
 def test_latlong_validate(latlong_df):
     error_message = re.escape(
         "Cannot initialize Woodwork. Series does not contain properly formatted "
--- a/woodwork/tests/logical_types/test_logical_types.py
+++ b/woodwork/tests/logical_types/test_logical_types.py
@@ -4,7 +4,7 @@ import numpy as np
 import pandas as pd
 import pytest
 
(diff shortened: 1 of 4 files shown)


woodwork/tests/logical_types/test_logical_types.py
```diff
@@ -155,6 +155,25 @@ def test_latlong_transform(latlong_df):
         pd.testing.assert_series_equal(actual, expected)
 
 
+def test_latlong_transform_empty_series(empty_latlong_df):
+    df_type = str(type(empty_latlong_df))
+    dask = "dask" in df_type
+    spark = "spark" in df_type
```
gsheni, line 161:
```suggestion
    is_dask = "dask" in df_type
    is_spark = "spark" in df_type
```

woodwork/tests/logical_types/test_logical_types.py
```diff
@@ -155,6 +155,25 @@ def test_latlong_transform(latlong_df):
...
+
+    latlong = LatLong()
+    series = empty_latlong_df['latlong']
+    actual = latlong.transform(series)
+
+    if dask:
+        actual = actual.compute()
+    elif spark:
```
gsheni, line 169:
```suggestion
    if is_dask:
        actual = actual.compute()
    elif is_spark:
```

gsheni (changes requested):
Overall, looks good

woodwork/tests/logical_types/test_logical_types.py
```diff
@@ -155,6 +155,25 @@ def test_latlong_transform(latlong_df):
         pd.testing.assert_series_equal(actual, expected)
 
 
+def test_latlong_transform_empty_series(empty_latlong_df):
+    df_type = str(type(empty_latlong_df))
+    is_dask = "dask" in df_type
+    is_spark = "spark" in df_type
```
jeff-hernandez, line 161: Could we use `_is_dask_dataframe` and `_is_spark_dataframe`?

rwedge: what if I change the `if / elif` to `if "dask" in df_type / elif "spark" in df_type`? and remove these variables?

jeff-hernandez: Sure, that works too, but are there any drawbacks for using `if _is_dask_dataframe(empty_latlong_df) ... elif _is_spark_dataframe(empty_latlong_df)`?

rwedge: it would be fewer lines, but `_is_x_dataframe` will probably be the clearest in meaning

jeff-hernandez (approved):
LGTM

gsheni (approved):

PR #1508, opened 2022-08-30
Fix minimum dependency checker

gsheni:
- Fixes the following error:
```
The workflow is not valid. .github/workflows/minimum_dependency_checker.yaml (Line: 33, Col: 13): The identifier 'min_dep_gen_core' may not be used more than once within the same scope.
```
- Passing Run:
  - https://github.com/alteryx/woodwork/actions/runs/2957508891


bchen1116 (approved):
Nice!

PR #1512, opened 2022-08-31
Change pivot point of 2-digit year in datetime inference

bchen1116:
Choose a different pivot point for 2-digit year rather than the default that pandas datetime uses.

Chose `current year + 10` as the pivot date, but can easily be changed to another


woodwork/tests/logical_types/test_logical_types.py
```diff
@@ -744,3 +744,115 @@ def test_replace_nans_same_types():
...
+    else:
+        dd = pytest.importorskip(
+            "dask.dataframe",
+            reason="Dask not installed, skipping",
+        )
+        df = dd.from_pandas(df, npartitions=2)
+    df.ww.init(logical_types={"dates": Datetime})
+    df_expected = pd.DataFrame({"dates": expected_values}, dtype="datetime64[ns]")
+    df = to_pandas(df)
+    df.sort_index(inplace=True)
```
bchen1116, line 857: pyspark ends up shuffling the rows during the transformations.

woodwork/logical_types.py
```diff
@@ -306,10 +306,39 @@ def _remove_timezone(self, series):
...
+            if date.year > 2030:
+                date = date.replace(year=date.year - 100)
+            return date
+
+        apply_filter = False
         new_dtype = self._get_valid_dtype(type(series))
         series = self._remove_timezone(series)
         series_dtype = str(series.dtype)
 
+        if series_dtype == "object":
```
ParthivNaresh, line 321: Should we also be checking `string`?

woodwork/logical_types.py
```diff
@@ -306,10 +306,39 @@ def _remove_timezone(self, series):
 
     def transform(self, series, null_invalid_values=False):
         """Converts the series data to a formatted datetime. Datetime format will be inferred if datetime_format is None."""
+
+        def _year_filter(date):
+            """Applies a filter to the years to ensure that the pivot point isn't too far forward."""
+            if date.year > 2030:
+                date = date.replace(year=date.year - 100)
```
ParthivNaresh, line 313: I'm not sure if we want to remove 100 years from just the years past 2030. We should remove 100 years from all dates in that series. Otherwise we'll run into a lot of uneven datetime intervals where a user might go from `2026, 2027, 2028, 2029, 2030, 1931, 1932, etc`. Thoughts?

jeff-hernandez: I came across this comment on [stack overflow](https://stackoverflow.com/a/37767723/5386567):
> ISO C standards:  values 69–99 are mapped to 1969–1999, and values 0–68 are mapped to 2000–2068.

Do we want to match thos cutoff points?

bchen1116: @ParthivNaresh So pandas currently does the same thing, just with a later pivot date (I believe '69')
image

I figured we wanted the same overall behavior, just at a different pivot date, which is why I chose 2030

ParthivNaresh (approved):
Nice work! I was initially hesitant about the `transform` time increase, since based on profiling it looks like `Datetime` now takes 100% longer to transform a series of longer lengths (50K < ) rows, but this is only an issue during a first time initialization of a `string` or `object` type. I think that's acceptable

woodwork/logical_types.py
```diff
@@ -306,10 +306,39 @@ def _remove_timezone(self, series):
...
+            return date
+
+        apply_filter = False
```
Cmancuso, line 316: is this supposed to be hard coded?

Cmancuso: ah nevermind, I see it just gets changed later.

woodwork/logical_types.py
```diff
@@ -306,10 +306,39 @@ def _remove_timezone(self, series):
...
+                    vals = re.findall(r"\d+", series_val)
+                    if len(vals):
+                        numbers_in_date.append(vals)
+                except TypeError:
+                    # for values like NaT, None, etc that aren't string/byte types
+                    continue
+            list_lengths = list(
+                map(lambda x: max([len(g) for g in x]), numbers_in_date),
+            )
+            if len(list_lengths) and max(list_lengths) == 2:
```
Cmancuso, line 339: this is definitely reliable right? There's no other > 2 digit numbers we have to worry about?

bchen1116: Yep, this is just checking that the longest value in our provided datetime is 2 digits, and the filter only applies to year values. This is reliable for changing the year value if it appears since it guarantees that the passed in year value has only 2 digits

Cmancuso: Yeah - I'm just wondering if there's not a date that looks like 10/23/22 1230PM or something like that.

woodwork/logical_types.py
```diff
@@ -356,6 +364,11 @@ def transform(self, series, null_invalid_values=False):
                     )
 
         series = self._remove_timezone(series)
+        if self.datetime_format is not None and "%y" in self.datetime_format:
```
Cmancuso, line 367: Love how much simpler this is. I think we should keep an eye and just make sure we're getting date formats the right way once Jeff's stuff goes in, but hopefully this will integrate nicely.

Cmancuso (approved):
Looks good to me!

PR #1523, opened 2022-09-15
Add Spearman Correlation to WW

bchen1116:
Adding spearman to dependence_dict and table accessors

### DIFF
--- a/woodwork/statistics_utils/_calculate_max_dependence_for_pair.py
+++ b/woodwork/statistics_utils/_calculate_max_dependence_for_pair.py
@@ -22,10 +22,14 @@ def _calculate_max_dependence_for_pair(result, min_shared, extra_stats):
     """
     # if pearson was not measured, mutual info must be max, since columns valid
     # for pearson are a subset of columns valid for mutual info
-    if "pearson" in result:
+    if "pearson" in result or "spearman" in result:
         score = pd.Series(
-            [result["mutual_info"], abs(result["pearson"])],
-            index=["mutual_info", "pearson"],
+            [
+                result["mutual_info"],
+                abs(result.get("pearson", np.nan)),
+                abs(result.get("spearman", np.nan)),
+            ],
+            index=["mutual_info", "pearson", "spearman"],
         )
         # to keep sign, get name of max score and use original value
         measure_used = score.idxmax()
--- a/woodwork/statistics_utils/_convert_ordinal_to_numeric.py
+++ b/woodwork/statistics_utils/_convert_ordinal_to_numeric.py
@@ -9,7 +9,7 @@ def _convert_ordinal_to_numeric(schema, data):
     Returns:
         None
     """
-    columns = [k for k, v in schema.logical_types.items() if str(v) == "Ordinal"]
+    columns = [k for k, v in schema.logical_types.items() if "Ordinal" in str(v)]
     for col_name in columns:
         order = list(schema.columns[col_name].logical_type.order)
         data[col_name] = data[col_name].apply(lambda x: order.index(x))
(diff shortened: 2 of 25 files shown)


woodwork/statistics_utils/_calculate_max_dependence_for_pair.py
```diff
@@ -22,10 +22,14 @@ def _calculate_max_dependence_for_pair(result, min_shared, extra_stats):
     """
     # if pearson was not measured, mutual info must be max, since columns valid
     # for pearson are a subset of columns valid for mutual info
-    if "pearson" in result:
+    if "pearson" in result or "spearman" in result:
```
Cmancuso, line 25: does this mean you get spearman + pearson even if you only request one of the two?

bchen1116: No, this checks that as long as we have another metric that isn't mutual info (ie pearson, spearman), we try to grab all the values to calculate the max. The `dictionary.get` method here will return `nan` if they don't exist, but otherwise will grab the values we need.

Cmancuso:
Can you add the lookingglass results when you have those ready? It would also be nice to run EvalML and see what the change is on the data check side (if that's affected by this)

woodwork/statistics_utils/_convert_ordinal_to_numeric.py
```diff
@@ -0,0 +1,16 @@
...
+
+    Args:
+        schema (woodwork.TableSchema): Woodwork typing info for the data.
+        data (dict[pd.Series]): Dictionary of Pandas series to use for
+            calculating dependence.
+
+    Returns:
+        None
+    """
+    for col_name in data.keys():
```
Cmancuso, line 12: there's no way to know this without looping through all the columns?

woodwork/utils.py
```diff
@@ -389,6 +389,28 @@ def get_valid_pearson_types():
     return valid_types
 
 
+def get_valid_spearman_types():
+    """
+    Generate a list of LogicalTypes that are valid for calculating Spearman correlation. Note that
+    index columns are not valid for calculating dependence, but their types may be
```
Cmancuso, line 395: I was under the impression we should be ignoring index columns for all correlation calculations?

woodwork/statistics_utils/_convert_ordinal_to_numeric.py
bchen1116: I think this would be the most clean and understandable way to do this. We could use `filter` logic, but I believe we would still need to loop and it could make the code less readable.

woodwork/utils.py
bchen1116: Yep, they are invalid. This is just warning that their types might still be returned, but we won't use it for calculations

woodwork/table_accessor.py
```diff
@@ -1122,6 +1126,111 @@ def pearson_correlation(
         )
         return pd.DataFrame(pearson_dict)
 
+    @_check_table_schema
+    def spearman_correlation_dict(
```
Cmancuso, line 1130: Is there any way to consolidate these at all to reduce total lines? could we just take in a correlation_type as an input?

Cmancuso (approved):
Looking good, just one other comment!

woodwork/table_accessor.py
bchen1116: @Cmancuso According to @rwedge, we previously had `mutual_info` and `mutual_info_dict` as accessor methods before `dependence` and `dependence_dict` were added to WW. When he added Pearsons, he extended these accessor methods for continuity, and I decided to extend Spearman also for continuity. These are all supported purely by `dependence` and `dependence_dict` methods, but consolidating them could require a breaking change.

PR #1531, opened 2022-09-21
Allow calculation of correlation metrics with a target column

bchen1116:
Add `target_col` argument to `dependence/dependence_dict` methods to calculate correlation metrics between features and the specified `target_col`


woodwork/utils.py
```diff
@@ -411,6 +411,20 @@ def get_valid_spearman_types():
     return valid_types
 
 
+def get_valid_correlation_metrics():
```
Cmancuso, line 414: is this only used in testing / error handling? Also wondering if we should have this tied to a list defined in woodwork instead of hard-coding it.

Cmancuso: Like should we have a constant list somewhere that defines all of them and pull from that list. So that we don't have to define what these are in multiple places.

ParthivNaresh (approved):
This looks great nice work! Do we have any future plans on adding another parameter that takes `column_names` to only calculate the dependence dict between specified columns?

woodwork/utils.py
bchen1116: Moved the list to `config_defaults`!

bchen1116:
> This looks great nice work! Do we have any future plans on adding another parameter that takes `column_names` to only calculate the dependence dict between specified columns?

@ParthivNaresh currently no plan to do so. If there is a request for a feature like this, it can easily be done, but functionality-wise, it would be the same as taking a subset of the dataset and running `dependence/dependence_dict` on that set, ie
```python
df_subset = df[[column_names]]
df_subset.ww.dependence_dict()
```

Cmancuso (approved):

PR #1533, opened 2022-09-27
Replace uses of dataframe or series `append` with `concat` where appropriate

### LINKED ISSUE #1493: Stop using df.append in codebase to resolve warning
- In building the Featuretools docs I saw this warning
```python
/Users/gaurav.sheni/.pyenv/versions/3.9.13/lib/python3.9/site-packages/woodwork/type_sys/utils.py:132: FutureWarning: The frame.append method is deprecated and will be removed from pandas in a future version. Use pandas.concat instead.
```
- We should remove `df.append` and use `df.concat`
- Completion: No more `.append` calls in Woodwork

sbadithe:
- Fixes #1493 
- Replaces uses of deprecated `append` method for dataframes and series with `concat` 
- Fixes bug in `test_numeric_histogram` where results of call to `append` were not being used
- Other uses of `append` are not altered


woodwork/utils.py
```diff
@@ -54,6 +54,34 @@ def import_or_none(library):
         return None
 
 
+def ww_concat(base, to_add):
+    """Selects and calls the appropriate concat method based on the type on the base and to_add series
+
+    Args:
+        base: base series or dataframe
+        to_add: series or dataframe to be concatenated
+    Returns: the series or dataframe after concatenation
```
gsheni, line 63:
```suggestion
    Args:
        base: base Series/DataFrame
        to_add: Series/DataFrame to be concatenated
        
    Returns: 
        Series/DataFrame: the resulting after concatenation
```

woodwork/utils.py
```diff
@@ -54,6 +54,36 @@ def import_or_none(library):
         return None
 
 
+def ww_concat(base, to_add):
+    """Selects and calls the appropriate concat method based on the type on the base and to_add series
```
ParthivNaresh, line 58: Nit: based on the type **_of_** the base

woodwork/utils.py
```diff
@@ -54,6 +54,36 @@ def import_or_none(library):
...
+        base,
+    ) or ww.accessor_utils._is_dask_series(base):
+        concatenated_obj = dd.concat([base, to_add])
+    elif ww.accessor_utils._is_spark_dataframe(
+        base,
+    ) or ww.accessor_utils._is_spark_series(base):
+        concatenated_obj = ps.concat([base, to_add])
+    else:
+        raise TypeError(
+            f"When attempting concatenation, encountered unknown type of base: {base}",
```
ParthivNaresh, line 82: Why not include `to_add` in the error message? The unknown type could belong to that argument as well right?

ParthivNaresh (approved):
Looks good to me, thanks!

woodwork/tests/accessor/test_statistics.py
```diff
@@ -1523,7 +1523,7 @@ def test_datetime_get_recent_value_counts():
 
 def test_numeric_histogram():
     column = pd.Series(np.random.randn(1000))
-    column.append(pd.Series([np.nan, " ", "test"]))
+    column = pd.concat([column, pd.Series([np.nan])])
```
thehomebrewnerd, line 1526: Looks like maybe this was broken previously as we just threw away the result of the previous append call. Should we update this to include `" "` and `"test"` like we intended to do before, or does it no make sense to have non-numeric values included for this test?

woodwork/utils.py
```diff
@@ -54,6 +54,36 @@ def import_or_none(library):
...
+        concatenated_obj = pd.concat([base, to_add])
+    elif ww.accessor_utils._is_dask_dataframe(
+        base,
+    ) or ww.accessor_utils._is_dask_series(base):
+        concatenated_obj = dd.concat([base, to_add])
+    elif ww.accessor_utils._is_spark_dataframe(
+        base,
+    ) or ww.accessor_utils._is_spark_series(base):
+        concatenated_obj = ps.concat([base, to_add])
+    else:
```
thehomebrewnerd, line 80: If this is just an internal testing utility, I'm not sure we even need to include the error handling?

woodwork/utils.py
```diff
@@ -54,6 +54,36 @@ def import_or_none(library):
         return None
 
 
+def ww_concat(base, to_add):
```
thehomebrewnerd, line 57: Should this be a private testing utility instead? Seems like we are not using this anywhere outside of the tests and I'm thinking this is not a function we would ever expect a Woodwork user to use.

Also, the name `ww_concat` makes me thing this is a function for concatenating Woodwork objects...maybe we should consider a different name?

bchen1116, line 57: can we add specific tests here that are ensuring this function does as we'd expect

woodwork/tests/accessor/test_statistics.py
sbadithe: This is something I wondered too, I think I made a Slack post about this earlier. It is worth noting that the inclusion of those strings leads the test to fail, so we would have to alter the expected result if we wanted to keep the strings.

thehomebrewnerd: Not real sure why the were there to begin with, so seems fine to leave them out now.

sbadithe: sounds good!

woodwork/utils.py
sbadithe: will do, adding to `test_utils.py`

sbadithe: Regarding the name, how does `_concat_dataframe_or_series` sound? It is a little verbose but more descriptive.

thehomebrewnerd: I'm fine with that name

sbadithe: Interestingly, it was not my intent to alter test methods only. It just so happened that all uses of `append` (at least the ones that are apparent to me to be deprecated) were in testing functions. Would making this a test utility entail moving this function to another part of the codebase?

Would this be something useful in other parts of Woodwork?

For example, I noticed a prospective use here: 
https://github.com/alteryx/woodwork/blob/e5b1c66bbadf1cfd07de4ae6bef25f4226addac4/woodwork/utils.py#L528

@thehomebrewnerd @ParthivNaresh @bchen1116

bchen1116: That sounds reasonable to me. I think having it integrated into other parts of woodwork where we might use concat/append would be helpful, although that seems out of scope for this. Maybe it'd be useful to leave this as a testing util and file an issue to look into potential uses for it in woodwork general at a later point?

thehomebrewnerd: Yeah, I agree with @bchen1116. In order to use your new function in the `concat_columns` function, you would need to make some more updates to the function to allow for changing the concat axis. If the only place it would be used is in that one spot, I'm not sure that is worthwhile.

woodwork/tests/utils/test_utils.py
```diff
@@ -617,3 +618,20 @@ def test_callback_caller_no_callback():
     caller.update(1)
 
     assert caller.current_progress == 0
+
+
+def test_concat_dataframe_or_series():
```
bchen1116, line 623: Nit: can we add a test for dataframes as well to ensure that works. Also, can we check the ordering of the data and/or index of the result is what we expect?

bchen1116: As well as a test case that would raise the error that we throw!

woodwork/utils.py
```diff
@@ -54,6 +54,36 @@ def import_or_none(library):
         return None
 
 
+def concat_dataframe_or_series(base, to_add):
+    """Selects and calls the appropriate concat method based on the type of the base and to_add series
```
bchen1116, line 58: nit-wording: `the base and to_add objects.`?

woodwork/utils.py
sbadithe: Sounds good! To make sure I am understanding this correctly, we would want to to move this to `woodwork/tests/testing_utils/concat.py` and then create a follow-up issue to track potentially adding more functionality to this method (or adding a new method).

thehomebrewnerd: I think we can just add it to the existing `table_utils.py` file rather than creating a new one just for this utility function. Yes, you can then create a new issue to look into further updates.

woodwork/utils.py
sbadithe: Agreed -- removed it. Furthermore, the inbuilt error messages are adequately informative.

PR #1547, opened 2022-10-24
Added medcouple as a method for outlier detection

ParthivNaresh:


pyproject.toml
```diff
@@ -40,6 +40,7 @@ dependencies = [
     "scikit-learn >= 0.22",
     "python-dateutil >= 2.8.1",
     "scipy >= 1.4.0",
+    "statsmodels >= 0.13.2",
```
ParthivNaresh, line 43: Needed to support the medcouple implementation

woodwork/column_accessor.py
```diff
@@ -388,20 +389,52 @@ def set_semantic_tags(self, semantic_tags):
         """
         self._schema._set_semantic_tags(semantic_tags)
 
+    @_check_column_schema
+    def get_outliers(self, method="best"):
```
ParthivNaresh, line 393: Now that we have two ways for calculating outliers, and under the possibility of adding more ways in the future, I wanted to create a general function that could retrieve the outliers information based on the method requested.

woodwork/column_accessor.py
```diff
@@ -388,20 +389,69 @@ def set_semantic_tags(self, semantic_tags):
         """
         self._schema._set_semantic_tags(semantic_tags)
 
+    @_check_column_schema
+    def get_outliers(
```
ParthivNaresh, line 393: Now that we have two ways for calculating outliers, and under the possibility of adding more ways in the future, I wanted to create a general function that could retrieve the outliers information based on the method requested.

This would ideally act as the primary means by which to access outlier information.

woodwork/column_accessor.py
```diff
@@ -388,20 +389,69 @@ def set_semantic_tags(self, semantic_tags):
...
+        return _get_box_plot_info_for_column(
+            self._series,
+            method=method,
+            quantiles=quantiles,
+            include_indices_and_values=include_indices_and_values,
+            ignore_zeros=ignore_zeros,
+        )
+
     @_check_column_schema
     def box_plot_dict(
```
ParthivNaresh, line 442: This function will be maintained as it is currently used.

woodwork/column_accessor.py
```diff
@@ -434,6 +484,59 @@ def box_plot_dict(
...
         return _get_box_plot_info_for_column(
             self._series,
+            method="box_plot",
+            quantiles=quantiles,
+            include_indices_and_values=include_indices_and_values,
+            ignore_zeros=ignore_zeros,
+        )
+
+    @_check_column_schema
+    def medcouple_dict(
```
ParthivNaresh, line 494: This is added as a convenience function to match `box_plot_dict`. Ideally `box_plot_dict` would have been called `iqr_dict` to avoid confusion about the outlier calculation method (IQR aka Box Plot) vs the way we display the outlier information to the user (Box and Whiskers Plot), since both `box_plot_dict` and `medcouple_dict` generate information for a Box and Whiskers Plot. We will maintain the function name of `box_plot_dict` to avoid a sizable breaking change.

woodwork/statistics_utils/_determine_best_outlier_method.py
```diff
@@ -0,0 +1,43 @@
...
+    Note:
+        The calculation of the medcouple statistic has a large memory requirement of O(N**2), therefore larger series will
+        have a random subset selected in order to determine skewness and the best outlier method.
+
+    Returns:
+        MedcoupleHeuristicResult - Named tuple with 2 fields
+            method (str): Name of the outlier method to use.
+            mc (float): The medcouple statistic (if the method chosen is medcouple, otherwise None).
+    """
+    np.random.seed(42)
```
ParthivNaresh, line 31: A constant seed will make sure that no matter how many times the random sampling is run on a series, the same observations will be selected every time.

woodwork/statistics_utils/_determine_best_outlier_method.py
```diff
@@ -0,0 +1,63 @@
...
+        series (Series): Data on which sampling will occur.
+
+    Note:
+        The calculation of the medcouple statistic has a large memory requirement of O(N**2), therefore a series over 10,000
+        observations will be randomly sampled up to 10,000.
+
+    Returns:
+        float: Medcouple statistic
+    """
+    np.random.seed(42)
```
ParthivNaresh, line 23: A constant seed will make sure that no matter how many times the random sampling is run on a series, the same observations will be selected every time.

pyproject.toml
gsheni, line 43: I disagree with this and think this is a bad idea. We create our open source libraries to have as minimal dependencies as possible. The more core dependencies you rely on, the problematic it can be for someone installing your library, and more likely dependency conflicts are in the future. 

If you need this library for `medcouple`, we should make it an optional library for Woodwork. Users can do `pip install woodwork[stats]` and get this functionality. Additionally, not everyone that uses Woodwork, will want to use `scipy`, `scikit-learn` or `statsmodels`.

gsheni: Also `statsmodels` is a large package.

ParthivNaresh: I see, I can make `statsmodels` an optional library for sure. Regarding your second point, should `scipy` and `scikit-learn` be part of an optional installation as well?

ParthivNaresh: The medcouple implementation in `statsmodels` is actually fairly basic, we can make our own implementation to avoid importing an entire package just for this. Also I filed this issue [here](https://github.com/alteryx/woodwork/issues/1548) to track the rest

gsheni: nice find!

woodwork/statistics_utils/_determine_best_outlier_method.py
gsheni: Do you want to make it an optional argument on the function?

gsheni: Also, another way to guarantee the same results is to use the first 10K rather than a random sample of 10K. Up to you.

ParthivNaresh: Depending on how the data has been sorted, the first X observations might not capture the distribution in any meaningful way, a random sample would be more robust. But I like the idea of adding an optional argument!

woodwork/statistics_utils/_get_box_plot_info_for_column.py
```diff
@@ -1,24 +1,65 @@
...
 
 
+def _determine_coefficients(mc):
+    if mc >= 0:
+        lower_coef = -3.5
+        higher_coef = max(1, 4 * (1 - mc))
+    else:
+        lower_coef = min(-1, -4 * (1 + mc))
+        higher_coef = 3.5
+    return lower_coef, higher_coef
```
ParthivNaresh, line 21: One of the things that I noticed when going through the datasets that had skewed columns was that when a series had a large medcouple statistic, the corresponding upper or lower bound (depending on if the skew was right or left) was much larger than it should have been. This is due to the part of the formula where `e is raised to the power of (coefficient x medcouple stat)`.

This becomes an exponentially larger quantity which when multiplied by 1.5 and the IQR, can lead to bounds far outside of what might be helpful. The logic here works with that in mind and limits some of the larger quantities that result from a medcouple statistic closer to -1 and 1. An important reason for this is that a medcouple magnitude near 1 doesn't necessarily mean that the outliers are incredibly far away from the bulk of the observations, it also relates to the body of the distribution and the dispersion of the points.

woodwork/statistics_utils/_get_medcouple_statistic.py
```diff
@@ -0,0 +1,47 @@
...
+    y = np.squeeze(np.asarray(y))
+    y = np.sort(y)
+
+    n = y.shape[0]
+    if n % 2 == 0:
+        mf = (y[n // 2 - 1] + y[n // 2]) / 2
+    else:
+        mf = y[(n - 1) // 2]
+
+    z = y - mf
```
bchen1116, line 24: nit, but would be really helpful if these variables had more helpful names. Would make it easier to understand in the future

bchen1116 (commented):
Left a couple of nits, but looking great so far!

woodwork/statistics_utils/_determine_best_outlier_method.py
```diff
@@ -0,0 +1,66 @@
+from collections import namedtuple
+
+import numpy as np
+import pandas as pd
+
+import woodwork as ww
+from woodwork.statistics_utils._get_medcouple_statistic import _get_medcouple
+
+
+def _sample_for_medcouple(series, seed=42):
```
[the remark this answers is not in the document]
ParthivNaresh: I felt like the answer to the ultimate question felt like a good standard seed to work with, but I'm open to a different default

woodwork/statistics_utils/_determine_best_outlier_method.py
```diff
@@ -0,0 +1,66 @@
...
+
+    Returns:
+        float: Medcouple statistic from the sampled data
+    """
+    np.random.seed(seed)
+    series_size = len(series)
+    if series_size < ww.config.get_option("medcouple_sample_size"):
+        series_sample = series.copy()
+    else:
+        series_sample = np.random.choice(
```
Cmancuso, line 29: does this include the seed?

woodwork/statistics_utils/_get_box_plot_info_for_column.py
Cmancuso: Interesting - so this will actually mean we see more outliers in medcouple by reducing the size of the whiskers? The constant coefficient is just a part of this statistic?

woodwork/statistics_utils/_get_box_plot_info_for_column.py
```diff
@@ -125,6 +175,9 @@ def _get_box_plot_info_for_column(
             "high_indices": high_series.index.tolist(),
         }
 
+    if method == "medcouple":
+        outliers_dict["medcouple"] = mc
```
Cmancuso, line 179: are we ever returning both medcouple and box and whisker? If not, maybe we can just add a type = [medcouple | box] or something like that?

woodwork/tests/accessor/test_statistics.py
```diff
@@ -1901,6 +1908,220 @@ def test_box_plot_optional_return_values(outliers_df):
...
+def test_get_outliers_for_column_with_nans_medcouple(skewed_outliers_df):
+    contains_nans_series = skewed_outliers_df["has_outliers_with_nans"]
+    contains_nans_series.ww.init()
+
+    medcouple_dict = contains_nans_series.ww.medcouple_dict()
+
+    expected_skewed_dict = {
+        "low_bound": 1.9490141192882326,
+        "high_bound": 13.459435908219328,
+        "quantiles": {
```
Cmancuso, line 2027: do we do any testing with custom quantiles?

Cmancuso (commented):
Looking good, a few small comments and just need to double check performance stuff but good stuff!

woodwork/statistics_utils/_get_box_plot_info_for_column.py
ParthivNaresh: Yes reducing the size of the whiskers means we'll see more outliers. In the end it depends on the mindset we want to have and this can even become a configurable parameter.

Do we want to err on the side of caution and make it smaller to avoid accidentally removing values that would have been outliers? Or do we want to more freely set outlier boundaries and if observations that would previously have been outliers end up flying under the radar and ignored, then so be it.

woodwork/statistics_utils/_get_box_plot_info_for_column.py
```diff
@@ -1,24 +1,65 @@
...
+    if method == "medcouple":
+        if mc is None:
+            raise ValueError(
+                "If the method selected is medcouple, then mc cannot be None."
+            )
+        # Medcouple bounds calculation - coefficients change based on the skew direction
+        lower_bound_coef, higher_bound_coef = _determine_coefficients(mc)
+        low_bound = q1 - 1.5 * np.exp(lower_bound_coef * mc) * iqr
+        high_bound = q3 + 1.5 * np.exp(higher_bound_coef * mc) * iqr
+    elif method == "box_plot":
```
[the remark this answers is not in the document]
ParthivNaresh: I decided to make it a little cleaner and changed it, thanks!

woodwork/statistics_utils/_determine_best_outlier_method.py
ParthivNaresh: The seed is initialized a few lines above `np.random.seed(seed)`

woodwork/statistics_utils/_get_box_plot_info_for_column.py
ParthivNaresh: It's only one or the other, good idea!

woodwork/tests/accessor/test_statistics.py
ParthivNaresh: We do! The logic for it comes before the bounds or medcouple calculation so its method independent, but I went ahead and changes a test to account for the medcouple approach.

ParthivNaresh:
Here are the perf [tests](https://alteryx.atlassian.net/wiki/spaces/PS/pages/1576730738/Box+Plot+vs+Medcouple)

woodwork/column_accessor.py
```diff
@@ -388,20 +389,69 @@ def set_semantic_tags(self, semantic_tags):
...
+                outlier values and their indices will be included in the returned dictionary.
+                Defaults to True.
+            ignore_zeros (bool): Whether to ignore 0 values (not NaN values) when calculating the box plot and outliers.
+                Defaults to False.
+
+        Returns:
+            (dict[str -> float,list[number]]): Returns a dictionary containing box plot information for the Series.
+                The following elements will be found in the dictionary:
+
+                - low_bound (float): the lowest data point in the dataset excluding any outliers - to be used as a whisker
```
Cmancuso, line 419: should we include the method used as part of the response here?

woodwork/column_accessor.py
```diff
@@ -434,6 +484,59 @@ def box_plot_dict(
...
+        using the Medcouple statistic method.
+
+        Args:
+            quantiles (dict[float -> float], optional): A dictionary containing the quantiles for the data
+                where the key indicates the quantile, and the value is the quantile's value for the data. If
+                no quantiles are provided, they will be computed from the data.
+            include_indices_and_values (bool, optional): Whether or not the lists containing individual
+                outlier values and their indices will be included in the returned dictionary.
+                Defaults to True.
+            ignore_zeros (bool): Whether to ignore 0 values (not NaN values) when calculating the box plot and outliers.
```
Cmancuso, line 510: this will probably eventually be a semantic tag on the column - just curious if we'd still want that to be an input or if we have access to that via the series. NAB

woodwork/statistics_utils/_get_medcouple_statistic.py
```diff
@@ -0,0 +1,42 @@
...
+
+    Returns:
+        float: Medcouple statistic from the sampled data
+    """
+    np.random.seed(seed)
+    series_size = len(series)
+    if series_size < ww.config.get_option("medcouple_sample_size"):
+        series_sample = series.copy()
+    else:
+        series_sample = np.random.choice(
```
Cmancuso, line 37: will this generate the same results each time it's run?

Cmancuso:
Some small comments, but looking good to me!

woodwork/column_accessor.py
ParthivNaresh: Once it becomes a semantic tag on the column we should access it via the series, for now I'd prefer to keep it as an argument

woodwork/statistics_utils/_get_medcouple_statistic.py
ParthivNaresh: Yep! Setting `np.random.seed(seed)` in line 32 sets the seed that numpy uses in "random" selections, making it deterministic

Cmancuso (approved):

PR #1549, opened 2022-10-25
Add Boolean Inference

bchen1116:


woodwork/tests/utils/test_read_file.py
```diff
@@ -82,7 +82,7 @@ def test_read_file_validation_control(
             "sample.csv",
             ("to_csv", {"index": False}),
             {
-                "nrows": 2,
+                "nrows": 3,
```
bchen1116, line 85: changing to 3 rows because otherwise, `id` has values `'0', '1'`, which gets inferred as boolean

woodwork/type_sys/inference_functions.py
```diff
@@ -73,7 +75,10 @@ def double_func(series):
 
 
 def boolean_func(series):
-    if boolean_nullable_func(series) and not series.isnull().any():
+    is_null = series.isnull().any()
+    if _is_dask_series(series):
+        is_null = is_null.compute()
```
bchen1116, line 80: need to call `compute()` to accurately determine if column has null for dask

woodwork/type_sys/inference_functions.py
```diff
@@ -93,8 +94,18 @@ def boolean_nullable_func(series):
                 {False},
             ]:
                 return True
+            series_lower = set(str(s).lower() for s in set(series))
```
Cmancuso, line 97: should we have this be optional based on an input param? Not 100% convinced that's the best approach but curious on your thoughts

woodwork/type_sys/inference_functions.py
```diff
@@ -93,8 +94,18 @@ def boolean_nullable_func(series):
...
+            if series_lower in [
+                set(boolean_list)
+                for boolean_list in config.get_option("boolean_inference_strings")
+            ]:
```
ParthivNaresh, line 101: So currently if only `True` OR `False` (instead of both) is in a series, then it can be identified as `Boolean` (lines 91 to 94 above). This logic here says that both the positive and negative class (both "y" and "n", or "0" and "1") have to be present in the series for it to be classified as a Boolean match. Is this something we're ok with?

docs/source/release_notes.rst
```diff
@@ -7,6 +7,7 @@ Future Release
 ==============
     * Enhancements
         * Replace use of deprecated ``append`` method for dataframes and series with ``concat`` method (:pr:`1533`)
+        * Improved ``Boolean`` inference to detect common string representations of boolean values (:pr:`1549`)
```
ParthivNaresh, line 10: This doesn't cover `BooleanNullable` as well?

ParthivNaresh (commented):
Nice work! Just left some comments to better understand the behaviour

woodwork/type_sys/inference_functions.py
ParthivNaresh: Also if there is a `nan` value here, it will be stringified and then lowercased. In that event, it won't match any of the pairs in `config.get_option("boolean_inference_strings")`. Is this intentional?

woodwork/type_sys/inference_functions.py
Cmancuso: Yeah I think it would need to contain exactly 1 or both of the items in the set. So a full column of "true" should still be boolean. Particularly because we're still selecting from the head.

woodwork/logical_types.py
```diff
@@ -203,6 +207,34 @@ class Boolean(LogicalType):
 
     primary_dtype = "bool"
 
+    def __init__(self, cast_nulls_as=None):
+        if cast_nulls_as and not isinstance(cast_nulls_as, bool):
+            raise ValueError(
+                f"Parameter `cast_nulls_as` must be either True or False, recieved {cast_nulls_as}",
+            )
+        self.cast_nulls_as = cast_nulls_as
```
jeff-hernandez, line 215: I think it'd be helpful to store a dictionary for mapping the values back to the original values. That way we don't lose any information and it may help with data representation and serialization in tempo. Probably not necessary right now in this PR, but future work possibly.

jeff-hernandez (commented):
Overall looking good!

woodwork/type_sys/inference_functions.py
Cmancuso: @bchen1116 I think the problem here is that we're sampling from the head and it's entirely possible that the first 100k rows are all true while the last 100k are all false or something similar. If we were doing a fully random sample, Id say you're probably right, but while we're looking at the HEAD only it might be a bit risky. Ultimately, probably not a dealbreaker since Parthiv will be working on that soon anyway, but something to think about. And currently typelib works in that if only 1 appeared it would still parse as boolean.

bchen1116:
No big change in runtime. Perf test report is attached
[3109.html.zip](https://github.com/alteryx/woodwork/files/9881846/3109.html.zip)

woodwork/type_sys/inference_functions.py
gsheni: @ParthivNaresh Regarding your example of only `True` or only `False` Python objects, if that is the case, the user has most likely done their own casting (pre-processing). Therefore, it makes sense for us to assume that it's a Boolean Logical Type.

With the string representation of only `yes` or only `no`, it seems presumptuous to assume that it's a Boolean Logical Type. What if there was an another category of `maybe`? 

I originally created the issue for this implementation (https://github.com/alteryx/woodwork/issues/153) because I had seen a column full of `yes`, `no` in a dataset. I have never seen a column with only `yes` or only `no`. 
Given Parthiv's upcoming work, maybe we punt on this? Additionally, if we get an actual dataset where this is problematic, we can revisit?

Cmancuso: Yeah I think this is fine. Sampling should be in probably within 2 sprints, so I think I'm okay with punting on this edge case for now. Though I'd say we probably should be consistent across all string representations. In which case, with a randomly sampled dataset, this logic should probably be fine.

Cmancuso (approved):

bchen1116:
Leaving this unmerged for the 0.20.0 release because it introduces a lot of breaking changes. Will aim to get in for 0.21.0 release.

ParthivNaresh:
> Leaving this unmerged for the 0.20.0 release because it introduces a lot of breaking changes. Will aim to get in for 0.21.0 release.

Can we also add all documented breaking changes to the release notes?

ParthivNaresh (approved):
Awesome stuff!

woodwork/type_sys/type_system.py
```diff
@@ -298,7 +298,7 @@ def infer_logical_type(self, series):
             series = series.head(INFERENCE_SAMPLE_SIZE)
         else:
             if _is_dask_series(series):
-                series = series.head(INFERENCE_SAMPLE_SIZE)
+                series = series.compute().head(INFERENCE_SAMPLE_SIZE)
```
thehomebrewnerd, line 301: I'm sure you know this already, but this `.compute()` call could be quite costly in terms of performance. Not sure how this would compare to using `series.head(INFERENCE_SAMPLE_SIZE, npartitions=-1)` in terms of performance, but I think you would get the same result in both cases.

bchen1116:
Changing to `npartitions=-1`, as per @thehomebrewnerd's suggestion, leads to slightly better times on the perf tests.
[3160.html.zip](https://github.com/alteryx/woodwork/files/9922790/3160.html.zip)

Will keep this change over `.compute()`

PR #1565, opened 2022-11-01
Fixes error message in `validate` method in `logical_types.py`

sbadithe:
The error message refers to an incompatible `dtype` when it should be `LogicalType`.

### DIFF
--- a/woodwork/logical_types.py
+++ b/woodwork/logical_types.py
@@ -83,7 +83,7 @@ class LogicalType(object, metaclass=LogicalTypeMetaClass):
         valid_dtype = self._get_valid_dtype(type(series))
         if valid_dtype != str(series.dtype):
             raise TypeValidationError(
-                f"Series dtype '{series.dtype}' is incompatible with {self.type_string} LogicalType, try converting to {valid_dtype}",
+                f"Series dtype '{series.dtype}' is incompatible with {self.type_string} LogicalType, try converting to {valid_dtype} dtype",
             )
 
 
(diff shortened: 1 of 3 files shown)


gsheni (dismissed):

gsheni:
Looks good but you'll have to update the string in some tests (as well as update the release notes)

woodwork/logical_types.py
```diff
@@ -83,7 +83,7 @@ def validate(self, series, *args, **kwargs):
         valid_dtype = self._get_valid_dtype(type(series))
         if valid_dtype != str(series.dtype):
             raise TypeValidationError(
-                f"Series dtype '{series.dtype}' is incompatible with {self.type_string} dtype.",
+                f"Series dtype '{series.dtype}' is incompatible with {self.type_string} LogicalType.",
```
sbadithe, line 86: Should we include `should be {valid_dtype}` at the end of the message?

gsheni: Yep, we can

gsheni (commented):
Will approve once valid_dtype is added

woodwork/logical_types.py
sbadithe: Changed it to  `try converting to {valid_dtype}` to try to be consistent w/ the error msg in `column_accessor.py`. Not sure which is better

woodwork/tests/logical_types/test_logical_types.py
```diff
@@ -271,7 +271,7 @@ def test_ordinal_validate(sample_series):
 
     new_type = "string"
     error_message = re.escape(
-        f"Series dtype '{new_type}' is incompatible with ordinal dtype.",
+        f"Series dtype '{new_type}' is incompatible with ordinal LogicalType.",
```
gsheni, line 274:
```suggestion
        f"Series dtype '{new_type}' is incompatible with ordinal LogicalType, try converting to {valid_dtype}",
```

gsheni (approved):

PR #1567, opened 2022-11-03
Add GitHub action to run looking glass perf tests on merge to main

simha104:
- Looking glass performance tests now run on merge


docs/source/release_notes.rst
```diff
@@ -52,6 +53,7 @@ v0.19.0 September 27, 2022
         * Remove ``dask[dataframe]`` version restriction (:pr:`1527`)
     * Testing Changes
         * Add kickoff for create conda forge pull request from release (:pr:`1515`)
+        * Perform performance tests on merge (:pr:`1567`)
```
ParthivNaresh, line 56: Why is this down here?

.github/workflows/woodwork_main_performance_tests.yaml
```diff
@@ -0,0 +1,143 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
+    env:
+      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_SECRET_ACCESS_KEY_ID }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+      AWS_DEFAULT_REGION: 'us-east-1'
+      AWS_SESSION_TOKEN: ""
+    steps:
+      - name: Checkout woodwork
```
ParthivNaresh, line 20: This default checks out `main` right? So `git rev-parse --short HEAD~1` gets the previous commit on `main`?

.github/workflows/woodwork_main_performance_tests.yaml
```diff
@@ -0,0 +1,143 @@
...
+          token: ${{ secrets.LOOKING_GLASS_TOKEN }}
+      - name: Installing Dependencies
+        run: |
+          make installdeps
+          make installdeps-test
+      - name: Run performance test (current)
+        id: current
+        run: |
+          response=$( looking-glass run-woodwork --username machineFL --scenarios-yaml woodwork_scenarios.yaml --woodwork-branch ${{ steps.get_hashes.outputs.current_hash }} --job-desc ${{ steps.get_hashes.outputs.current_hash }} )
+          echo "::set-output name=job_id::$(echo $response | sed 's/.*Job ID: \([^ ]*\).*/\1/')"
```
ParthivNaresh, line 51: What's going on here in the `echo`?

ParthivNaresh (commented):
Looks good, have you had the chance to run this locally first?

.github/workflows/woodwork_main_performance_tests.yaml
simha104: This echo moves the results of the job into an environment variable named after the job_id

ParthivNaresh (approved):
I think this is good to go! If there's a numbers reporting issue we can verify the numbers through our own separate perf tests, and we can turn off the job if there's any other issues until they can be fixed

PR #1571, opened 2022-11-09
Fix division by zero error in `type_system.py` 

sbadithe:
- Fixes division by zero error by checking length of series first
- Adds test to `test_ltype_inference.py` to test case with empty series
- This should resolve integration test failures with Featuretools main

### DIFF
--- a/woodwork/tests/type_system/conftest.py
+++ b/woodwork/tests/type_system/conftest.py
@@ -334,21 +334,21 @@ def pdnas(request):
 
 # Empty Inference Fixtures
 @pytest.fixture
-def pandas_empty():
+def pandas_empty_series():
     return [pd.Series([], dtype="object")]
 
 
 @pytest.fixture
-def dask_empty(pandas_empty):
-    return [pd_to_dask(series) for series in pandas_empty]
+def dask_empty_series(pandas_empty_series):
+    return [pd_to_dask(series) for series in pandas_empty_series]
 
 
 @pytest.fixture
-def pyspark_empty(pandas_empty):
-    return [pd_to_spark(series) for series in pandas_empty]
+def pyspark_empty_series(pandas_empty_series):
+    return [pd_to_spark(series) for series in pandas_empty_series]
 
 
-@pytest.fixture(params=["pandas_empty", "dask_empty", "pyspark_empty"])
+@pytest.fixture(params=["pandas_empty_series", "dask_empty_series", "pyspark_empty_series"])
 def empty_series(request):
     return request.getfixturevalue(request.param)
 
--- a/woodwork/tests/type_system/conftest.py
+++ b/woodwork/tests/type_system/conftest.py
@@ -332,25 +332,7 @@ def pdnas(request):
     return request.getfixturevalue(request.param)
 
 
-# Empty Inference Fixtures
-@pytest.fixture
-def pandas_empty():
-    return pd.Series([], dtype="object")
 
-
-@pytest.fixture
-def dask_empty(pandas_empty):
-    return pd_to_dask(pandas_empty)
-
-
-@pytest.fixture
-def pyspark_empty(pandas_empty):
-    return pd_to_spark(pandas_empty)
-
-
-@pytest.fixture(params=["pandas_empty", "dask_empty", "pyspark_empty"])
-def empty_series(request):
-    return request.getfixturevalue(request.param)
 
 
 # Null Inference Fixtures
--- a/woodwork/tests/type_system/conftest.py
+++ b/woodwork/tests/type_system/conftest.py
@@ -332,9 +332,6 @@ def pdnas(request):
     return request.getfixturevalue(request.param)
 
 
-
-
-
 # Null Inference Fixtures
 @pytest.fixture
 def pandas_nulls():
--- a/woodwork/tests/type_system/conftest.py
+++ b/woodwork/tests/type_system/conftest.py
@@ -331,18 +331,22 @@ def dask_pdnas(pandas_pdnas):
 def pdnas(request):
     return request.getfixturevalue(request.param)
 
+
 # Empty Inference Fixtures
 @pytest.fixture
 def pandas_empty():
-    return pd.Series([], dtype='object')
+    return pd.Series([], dtype="object")
+
 
 @pytest.fixture
 def dask_empty(pandas_empty):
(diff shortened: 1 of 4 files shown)


sbadithe:
Working on getting test cases in for this

woodwork/type_sys/type_system.py
```diff
@@ -310,6 +310,10 @@ def get_random_sample(series_, **kwargs):
             series = get_random_sample(series, **kw_args_sampling)
         else:
             # Dask and Spark don't accept the n argument
+
+            # prevent division by zero error
+            if not len(series):
```
sbadithe, line 315: I considered moving this before the `if` block on L305. I am open to moving it there, but since we don't divide by `len` in L305-L310, I think it should be fine to keep it here too.

woodwork/tests/type_system/conftest.py
```diff
@@ -332,6 +332,27 @@ def pdnas(request):
     return request.getfixturevalue(request.param)
 
 
+# Empty Series Inference Fixtures
+@pytest.fixture
+def pandas_empty():
+    return [pd.Series([], dtype="object")]
```
thehomebrewnerd, line 338: You can just return a Series and not a list (here and in the other fixtures), then you can remove the `for` loop in the new test you added.

woodwork/tests/type_system/conftest.py
```diff
@@ -332,6 +332,27 @@ def pdnas(request):
     return request.getfixturevalue(request.param)
 
 
+# Empty Series Inference Fixtures
+@pytest.fixture
+def pandas_empty():
```
thehomebrewnerd, line 337: Minor nitpick, but I would maybe call this `pandas_empty_series` to make it clear this is a series and not a dataframe like most of the other fixtures.

woodwork/tests/type_system/test_ltype_inference.py
```diff
@@ -252,6 +252,13 @@ def test_unknown_inference_all_null(nulls):
             assert isinstance(inferred_type, Unknown)
 
 
+def test_unknown_inference_empty_series(empty_series):
+    for series in empty_series:
+        inferred_type = ww.type_system.infer_logical_type(series.astype("string"))
+        inferred_type.transform(series)
```
thehomebrewnerd, line 258: I don't think this line is doing anything - you are just throwing away the return value from the `.transform()` call

woodwork/type_sys/type_system.py
thehomebrewnerd: I would leave it here - you don't need it for the pandas flow, so we shouldn't do the work of calculating the len until we need it.

Also, I would store the len in a variable and then use that variable on L315 and L317 so you don't end up calculating the length twice for cases where there are values present.

sbadithe: I could be wrong, but I remember reading that `len` is a `O(1)` attribute lookup in Python.

sbadithe: I agree keeping it on L315 makes sense though, seeing as we don't use the `frac` argument for Pandas.

woodwork/tests/type_system/test_ltype_inference.py
sbadithe: Will do! I wasn't sure if there were any important side effects to test here. I followed the pattern in `test_unknown_inference_all_null`, which called this function.

woodwork/tests/type_system/conftest.py
sbadithe: Just to clarify -- wrap L338 in a `pd.Series` call?

woodwork/type_sys/type_system.py
thehomebrewnerd: > I could be wrong, but I remember reading that `len` is a `O(1)` attribute lookup in Python.

I'm not exactly sure on that, but in this code block we are dealing with Dask and Spark dataframes which can be quite large. My concern is that we will end up having to read in those dataframes into memory twice to get the len. I'm not sure how those libraries implement len, but I think it is potentially costly.

woodwork/tests/type_system/conftest.py
thehomebrewnerd: No, just make the fixture look like this:
```
@pytest.fixture
def pandas_empty():
    return pd.Series([], dtype="object")
```

There is no reason to return a list that contains only a single element as far as I can tell.

woodwork/tests/type_system/test_ltype_inference.py
thehomebrewnerd: I guess this *could* be testing that it is possible to convert the original series from the "object" type to the "string" type without an error.

thehomebrewnerd: If that is the intent, I would maybe suggest we add a comment to that effect since it isn't immediately clear why what call would exist otherwise.

woodwork/type_sys/type_system.py
sbadithe: Ah, got it. That makes sense! I looked at the code again, and there's no way we hit L316 and L317 at the same time, but we can cache and make L317 and L320 more efficient!

sbadithe: That makes sense! I cached on L320 too to make it more efficient as well :)

woodwork/type_sys/type_system.py
```diff
@@ -310,10 +310,15 @@ def get_random_sample(series_, **kwargs):
...
+
+            # prevent division by zero error
+            series_len = len(series)
+            if not series_len:
+                return Unknown()
+            kw_args_sampling["frac"] = INFERENCE_SAMPLE_SIZE / series_len
             if _is_dask_series(series):
                 series = get_random_sample(
-                    series.head(len(series), npartitions=-1), **kw_args_sampling
+                    series.head(series_len, npartitions=-1), **kw_args_sampling
```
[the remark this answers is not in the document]
sbadithe: I looked at PR#1556 and I think this was to get the entire Dask series without using compute. 

@ParthivNaresh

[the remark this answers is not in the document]
thehomebrewnerd: I'm thinking if we need the entire series in memory we should just call `.compute()` here.

thehomebrewnerd (approved):

PR #1572, opened 2022-11-09
Fix IntegerNullable inference

### LINKED ISSUE #1550: Woodwork initialization fails with nullable int in dataframe
If you try to initialize woodwork on a dataframe with nullable integers it correctly infers the type but throws an error when it uses the pandas type conversion on the underlying data (because pandas turns anything with nans into floats)

The error is:

```
        """Converts the series dtype to match the logical type's if it is different."""
        if new_dtype != str(series.dtype):
            # Update the underlying series
            try:
                series = series.astype(new_dtype)
            except (TypeError, ValueError):
>               raise TypeConversionError(series, new_dtype, type(self))
E               woodwork.exceptions.TypeConversionError: Error converting datatype for b from type float64 to type Int64. Please confirm the underlying data is consistent with logical type IntegerNullable.

/usr/local/lib/python3.9/site-packages/woodwork/logical_types.py:76: TypeConversionError
```

```python
        data = DataFrame.from_dict(
            {
                "a": [numpy.inf, numpy.nan, numpy.NAN, -numpy.Inf, None],
                "b": [numpy.finfo("d").max, numpy.finfo("d").min, 3, 1, None],
            }
        )
        data = data.ww.init()
```

bchen1116:
fix #1550

### DIFF
--- a/woodwork/tests/type_system/test_ltype_inference.py
+++ b/woodwork/tests/type_system/test_ltype_inference.py
@@ -204,6 +204,8 @@ def test_categorical_integers_inference(integers):
             dtypes = get_spark_dtypes(dtypes)
         for series in integers:
             for dtype in dtypes:
+                if len(integers) < 12 and "64" not in dtypes:
+                    continue
                 inferred_type = ww.type_system.infer_logical_type(series.astype(dtype))
                 assert isinstance(inferred_type, Categorical)
 
@@ -215,6 +217,8 @@ def test_categorical_double_inference(doubles):
             dtypes = get_spark_dtypes(dtypes)
         for series in doubles:
             for dtype in dtypes:
+                if len(doubles) < 16 and dtypes != "float64":
+                    continue
                 inferred_type = ww.type_system.infer_logical_type(series.astype(dtype))
                 assert isinstance(inferred_type, Categorical)
 
--- a/woodwork/tests/type_system/test_ltype_inference.py
+++ b/woodwork/tests/type_system/test_ltype_inference.py
@@ -204,8 +204,6 @@ def test_categorical_integers_inference(integers):
             dtypes = get_spark_dtypes(dtypes)
         for series in integers:
             for dtype in dtypes:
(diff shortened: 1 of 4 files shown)


bchen1116:
Results of initialization with too-large values:
image

woodwork/tests/type_system/test_ltype_inference.py
```diff
@@ -204,6 +204,8 @@ def test_categorical_integers_inference(integers):
             dtypes = get_spark_dtypes(dtypes)
         for series in integers:
             for dtype in dtypes:
+                if len(integers) < 12 and "64" not in dtypes:
```
bchen1116, line 207: values too large for other dtypes

woodwork/tests/type_system/test_ltype_inference.py
```diff
@@ -215,7 +215,10 @@ def test_categorical_double_inference(doubles):
             dtypes = get_spark_dtypes(dtypes)
         for series in doubles:
             for dtype in dtypes:
+                # if len(doubles) < 16 and dtypes == "float_":
+                #     continue
```
ParthivNaresh, line 219: No longer necessary prints?

ParthivNaresh (approved):
Looks good! Quick question, we don't need to change the inference function for `Integer` or `Double` right? This isn't something they're affected by?

bchen1116:
> Looks good! Quick question, we don't need to change the inference function for `Integer` or `Double` right? This isn't something they're affected by?

The `Integer` inference function uses the `IntegerNullable` inference, so we should be good!

ParthivNaresh:
> > Looks good! Quick question, we don't need to change the inference function for `Integer` or `Double` right? This isn't something they're affected by?
> 
> The `Integer` inference function uses the `IntegerNullable` inference, so we should be good!

Ah and an otherwise integer series with an `np.inf` becomes `float64`, so that works out to the same so it won't belong to the integer function either way, nice!

PR #1574, opened 2022-11-14
Fix categorical dtype inference for postal code

### LINKED ISSUE #1569: Improve Woodwork error message for invalid init calls
If we do
```python
df = pd.Series(["90210", "60018", "10010"] * 100)
df.ww.init()
```
we get the following error:
```python
raise TypeValidationError(
woodwork.exceptions.TypeValidationError: Cannot initialize Woodwork. Series dtype 'object' is incompatible with PostalCode LogicalType. Try converting series dtype to 'category' before initializing or use the woodwork.init_series function to initialize.
```

If we were to follow the first recommendation, we get the following:
```python
df = pd.Series(["90210", "60018", "10010"] * 100).astype('category')
df.ww.init()
df.ww.schema
->
<ColumnSchema (Logical Type = Categorical) (Semantic Tags = ['category'])>
```
The second recommended action is correct, resulting in us getting a postal code logical type.

This issue tracks fixing this message so that users can get accurate inference on their datatables. In the case of series data, we should tell them to use `init_series` rather than `ww.init()`.

sbadithe:
- closes #1569


woodwork/tests/type_system/conftest.py
```diff
@@ -282,6 +282,32 @@ def natural_language(request):
     return request.getfixturevalue(request.param)
 
 
+# Postal Inference Fixtures
+@pytest.fixture
+def pandas_postal_codes():
+    return [
+        pd.Series(10 * ["77002", "55106"]),
```
bchen1116, line 289: Let's include one with a mix of 5 digit and 9 digit postal codes

docs/source/release_notes.rst
```diff
@@ -14,6 +14,7 @@ Future Release
         * Fixed ``IntegerNullable`` inference by checking values are within valid ``Int64`` bounds (:pr:`1572`)
         * Update demo dataset links to point to new endpoint (:pr:`1570`)
         * Fix DivisionByZero error in ``type_system.py`` (:pr:`1571`)
+        * Fix Categorical dtype inference for PostalCode logical type (:pr:`1574`)
```
bchen1116, line 17: nit, can we add double tic marks around `PostalCode`

woodwork/tests/type_system/test_ltype_inference.py
```diff
@@ -125,13 +126,23 @@ def test_categorical_inference(categories):
     if _is_spark_series(categories[0]):
         dtypes = get_spark_dtypes(dtypes)
     for ind, series in enumerate(categories):
-        if ind == len(categories) - 1:
+        if ind >= len(categories) - 2:
```
bchen1116, line 129: why do we have this change?

bchen1116 (commented):
Left some suggestions/questions

woodwork/tests/type_system/test_ltype_inference.py
sbadithe: I was curious but a little unsure, should we add a postal series to this fixture? 

https://github.com/alteryx/woodwork/blob/807ea4fb6fa4ec01db415d483ae0f51ca3e8f23b/woodwork/tests/type_system/conftest.py#L185

bchen1116: I don't think we need to add postal to the fixture. You added the test that would test that scenario

sbadithe: sounds good!

bchen1116 (approved):
LGTM!

PR #1584, opened 2022-11-22
Fix importlib DeprecationWarning; Removes expected test message to account for changes in Python3.11

### LINKED ISSUE #1582: DeprecatingWarning from using open_text in Natural Language inference function
- Using Python 3.11 with Woodwork results in the following DeprecationWarning
```python
../.pyenv/versions/3.11.0/lib/python3.11/site-packages/woodwork/type_sys/inference_functions.py:16
  /Users/gaurav/.pyenv/versions/3.11.0/lib/python3.11/site-packages/woodwork/type_sys/inference_functions.py:16: DeprecationWarning: open_text is deprecated. Use files() instead. Refer to https://importlib-resources.readthedocs.io/en/latest/using.html#migrating-from-legacy for migration advice.
    line.strip().lower() for line in pkg_resources.open_text(data, "1-1000.txt")
```

sbadithe:
- closes #1582
- also fixes a test that's broken with python 3.11. After this change, on my local machine, all tests on an install of `woodwork[dask]` pass with python 3.11 
- also resolves a few more FutureWarnings


pyproject.toml
```diff
@@ -40,6 +40,7 @@ dependencies = [
     "scikit-learn >= 0.22",
     "python-dateutil >= 2.8.1",
     "scipy >= 1.4.0",
+    "jsonschema == 3.0.2"
```
gsheni, line 43: Is there a way to avoid this? We don't use this import anywhere

sbadithe:
I think the codecov issues should resolve as soon as we start testing with 3.11.

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -2456,7 +2459,11 @@ def test_accessor_schema_properties(sample_df):
         assert prop_from_accessor == prop_from_schema
 
         # Assumes we don't have setters for any of these attributes
-        error = "can't set attribute"
+        if Version(python_version().strip(punctuation)) < Version("3.11.0"):
```
sbadithe, line 2462: We need to strip because the `python_version` call returns `3.11.0+`
I preferred this over `sys.version` which returns a larger, more detailed string that needs to be parsed.

woodwork/tests/accessor/test_table_accessor.py
```diff
@@ -2456,7 +2459,11 @@ def test_accessor_schema_properties(sample_df):
...
+            error = "can't set attribute"
+        else:
+            # Error messages were changed as part of Python 3.11
+            error = "has no setter"
         with pytest.raises(AttributeError, match=error):
```
ozzieD, line 2467: is there a reason for the match parameter here? If it throws the same built-in python exception, it doesn't really matter what the message is. Id rather delete this param and all the custom logic above it. No reason to look for it imo

pyproject.toml
```diff
@@ -40,6 +40,7 @@ dependencies = [
     "scikit-learn >= 0.22",
     "python-dateutil >= 2.8.1",
     "scipy >= 1.4.0",
+    "importlib-resources >= 5.10.0"
```
thehomebrewnerd, line 43: If I am following correctly, we are adding this because we can't use the new API with Python 3.8. If that is the case, we should probably create an issue to drop this dependency after we have dropped support for Python3.8 as this will no longer be necessary and we can just use `importlib.resources` directly.

woodwork/tests/accessor/test_table_accessor.py
thehomebrewnerd: Yeah, I agree with @ozzieD here. If we were checking for a custom error we were raising, I think we would want to use `match`, but since we are just checking that a built-in exception gets raised, I'm not sure there is much value in adding complexity to the test and checking the actual error message since that is beyond our control.

sbadithe: I'm not too sure, maybe to disambiguate between different types of AttributeError?

woodwork/tests/fixtures/datetime_freq.py
```diff
@@ -50,8 +50,8 @@
 def pad_datetime_series(dates, freq, pad_start=0, pad_end=100):
     dates = [pd.Timestamp(d) for d in dates]
 
-    head = pd.Series([])
-    tail = pd.Series([])
+    head = pd.Series([], dtype="float64")
+    tail = pd.Series([], dtype="float64")
```
thehomebrewnerd, line 54: Should this be `datetime64[ns]` here instead of `float64`?

woodwork/tests/accessor/test_statistics.py
```diff
@@ -2087,7 +2088,10 @@ def test_get_low_high_bound_warnings():
 
 def test_get_medcouple(outliers_df_pandas, skewed_outliers_df_pandas):
     has_outliers_series = outliers_df_pandas["has_outliers"]
-    has_outliers_series = has_outliers_series.append(pd.Series([39]), ignore_index=True)
+    has_outliers_series = pd.concat(
```
thehomebrewnerd, line 2091: Is there a reason we don't concat with `concat_dataframe_or_series` here as well?

sbadithe: I used `pd.concat` because it looks like the fixtures are always pandas. Would it be better to generalize this into `concat_dataframe_or_series`?

thehomebrewnerd: Probably doesn't matter too much, but might be nice to be consistent in how we handle this unless there is a reason to do it differently.

pyproject.toml
gsheni: Agreed, I'm a big fan of removing dependencies

woodwork/tests/accessor/test_table_accessor.py
tamargrey: I think just checking for the attribute error should be fine. We already know that the attribute exists on the schema and accessor objects, so we shouldn't have to worry that an attribute error would pop up because the attribute doesn't exist, which might have been what I was trying to differentiate from here? Probably just an oversight on my part--I'm totally fine with removing the matching.

woodwork/tests/fixtures/datetime_freq.py
sbadithe: Was unsure here. The current default dtype of an empty series in Pandas is `float64`, so this made explicit what was being tested currently. I was unsure if setting it to `datetime64[ns]` would somehow lead to us not testing the scenario where the series dtype was `float64`, if that was important.

thehomebrewnerd: Probably doesn't matter too much, but it looks like everything ends up getting converted to `datetime64[ns]` somewhere downstream of this anyway, so maybe it makes sense to start with that to be clear we expect these series to contain datetime values.

woodwork/tests/accessor/test_statistics.py
sbadithe: EDIT: Actually, this was because this concat call makes use of the `ignore_index` parameter, which we don't support in `concat_dataframe_or_series`

pyproject.toml
sbadithe: Created: https://github.com/alteryx/woodwork/issues/1591

docs/source/release_notes.rst
```diff
@@ -3,15 +3,18 @@
...
-  ==============
+Future Release
+==============
     * Enhancements
     * Fixes
+        * Fix importlib DeprecationWarning in ``inference_functions.py`` (:pr:`1584`)
     * Changes
     * Documentation Changes
     * Testing Changes
+        * Add handling of Python 3.11 error messages in ``test_table_accessor.py`` (:pr:`1584`)
```
ozzieD, line 14: remove this line

ozzieD (approved):

PR #1599, opened 2022-12-13
Improve handling of `Config` values for boolean inference

ParthivNaresh:
If a user attempts to update `boolean_inference_ints` as a config value to be an empty list, this results in an error during boolean inference since `[]` doesn't have positional indexers.


jeff-hernandez (commented):
Overall looks good! Do we want to add or update any tests for this?

ParthivNaresh:
@jeff-hernandez Closing this out since we have different logic in motion now!

PR #1600, opened 2022-12-14
Remove int from boolean predict

bchen1116:


woodwork/logical_types.py
```diff
@@ -968,7 +968,7 @@ def _coerce_boolean(series, null_invalid_values=False):
 
 def _transform_boolean(series, null_invalid_values):
     boolean_inference_list = config.get_option("boolean_inference_strings")
-    boolean_inference_list.append(config.get_option("boolean_inference_ints"))
+    boolean_inference_list.append(["1", "0"])
```
bchen1116, line 971: This allows users to pass in and properly cast [0,1] to boolean

ParthivNaresh: This means that columns with `"0" and "1"` won't be naturally inferred as `Boolean`/`BooleanNullable`, but if a user wants to specify `Boolean`/`BooleanNullable`, then we will appropriately mark `"0"` as `False` and `"1"` as `True`. Not including this line causes `series = series.apply(lambda x: valid.get(x, np.nan))` to replace all stringified values of `1s and 0s` with `np.nan`. Attempting to convert an `np.nan` in `transform` via `series = series.astype("bool")` evaluates all `nans` to `True`. Is my understanding correct @bchen1116?

bchen1116: Yep, this is correct!

woodwork/config.py
```diff
@@ -56,12 +56,9 @@
     "boolean_inference_strings": [
         ["yes", "no"],
         ["y", "n"],
-        ["1", "0"],
-        ["1.0", "0.0"],
         ["true", "false"],
         ["t", "f"],
     ],
-    "boolean_inference_ints": [1, 0],
```
jeff-hernandez, line 64: Do we still need the `boolean_inference_ints` key in the config to add int values `[1, 0]`?

ParthivNaresh: We can add them to `boolean_inference_strings` @jeff-hernandez , I believe that should be fine

docs/source/release_notes.rst
```diff
@@ -11,6 +11,7 @@ Future Release
         * Schema now maintains column order after renaming a column (:pr:`1594`)
     * Changes
         * Rename `backup_dtype` to `pyspark_dtype` (:pr:`1593`)
+        * Remove ``[0, 1]`` from boolean predictions (:pr:`1600`)
```
ParthivNaresh, line 14: Could we update this to say "Removed inference for ``["0", "1"]``, ``["0.0", "1.0"]``, and ``[0, 1]``" as ``Boolean`` logical types, but maintained forced inference of such values." or something like that?

woodwork/config.py
bchen1116: Should be fine to add them to the `boolean_inference_ints`!

woodwork/type_sys/inference_functions.py
```diff
@@ -110,6 +110,12 @@ def boolean_nullable_func(series):
                 return True
         except TypeError:  # Necessary to check for non-hashable values because of object dtype consideration
             return False
+    elif pdtypes.is_integer_dtype(series.dtype) and len(
+        config.get_option("boolean_inference_ints"),
+    ):
+        series_unique = set(series)
+        if series_unique == set(config.get_option("boolean_inference_ints")):
+            return True
```
ParthivNaresh, line 118: This only works for a value set as a list like `[0, 1]` right? We should consider making this capable of handling a list of lists so that the format matches that of `boolean_inference_strings`.
```
elif pdtypes.is_integer_dtype(series.dtype) and len(
        config.get_option("boolean_inference_ints"),
    ):
    series_unique = set(series)
    for boolean_list in config.get_option("boolean_inference_ints"):
        if series_unique == set(boolean_list):
            return True
```
What do you think?

bchen1116: Hmm, I don't think we should look towards automating other possible `int` inputs. Imo, there shouldn't be other int values that can be considered boolean the way `[0, 1]` are. What other possible values were you thinking?

ParthivNaresh: I agree that I can't personally think of other cases where open source users might want to add to this list, but if they did try they'd be limited in this regard. Either way I think it's totally fine if we don't add this here, it's not a priority functionality

bchen1116: I think it's fine to leave this out!

ParthivNaresh (approved):
Looks good to me!

jeff-hernandez (approved):

PR #1612, opened 2023-01-09
Update `Boolean` `transform` to avoid running a check for nans if the `dtype == bool`

ParthivNaresh:


Cmancuso:
Do we need to test that Value Error? Otherwise lgtm

woodwork/logical_types.py
```diff
@@ -223,15 +223,17 @@ def transform(self, series, null_invalid_values=False):
...
-                )
-            series.fillna(self.cast_nulls_as, inplace=True)
+        if not pdtypes.is_dtype_equal("bool", series.dtype):
+            if (is_dask and series.isna().any().compute()) or (
+                not is_dask and series.isna().any()
+            ):
+                if self.cast_nulls_as is None:
+                    raise ve
+                series.fillna(self.cast_nulls_as, inplace=True)
         series = _coerce_boolean(series, True)
```
bchen1116, line 237: Should this also be in the `if` statement? We shouldn't need to coerce if the type is 'bool'

bchen1116:
Might be useful for us to run our own LG tests to make sure this is ok!

ParthivNaresh:
@Cmancuso Added a test just to be sure!

bchen1116 (approved):
Nice!

PR #1614, opened 2023-01-12
Add ruff for linting and replace isort/flake8

gsheni:


bchen1116 (approved):
LGTM! Yay for faster linting and fixing

PR #1631, opened 2023-01-31
Add GitHub Actions cache to speed up workflows

gsheni:
### Lint Check  2 mins -> 30 secs
- https://github.com/alteryx/woodwork/actions/runs/4058547256/jobs/6985597115
- https://github.com/alteryx/woodwork/actions/runs/4058519011/jobs/6985539882
### Build Docs 4:14 mins -> 3:44 mins
- https://github.com/alteryx/woodwork/actions/runs/4058519013/jobs/6985540730
- https://github.com/alteryx/woodwork/actions/runs/4058547253/jobs/6985597958
### Install Test 1:22 mins -> 59 secs
- https://github.com/alteryx/woodwork/actions/runs/4058844848/jobs/6986244138
- https://github.com/alteryx/woodwork/actions/runs/4055891922/jobs/6979646823

### DIFF
--- a/.github/workflows/install_test.yaml
+++ b/.github/workflows/install_test.yaml
@@ -18,11 +18,11 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - name: Set up python ${{ matrix.python_version }}
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python_version }}
       - name: Checkout repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Build woodwork package
         run: make package
       - name: Install complete version of woodwork
--- a/.github/workflows/install_test.yaml
+++ b/.github/workflows/install_test.yaml
@@ -36,11 +36,11 @@ jobs:
       - name: Build woodwork package
         run: |
           make package
-      - name: Install woodwork with no dependencies
+      - name: Install woodwork with no dependencies (using cache)
         if: steps.cache.outputs.cache-hit == 'true'
         run: |
           python -m pip install -e "unpacked_sdist/[complete]" --no-deps
-      - name: Install woodwork with dependencies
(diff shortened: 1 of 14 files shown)


.github/workflows/install_test.yaml
```diff
@@ -17,19 +17,33 @@ jobs:
...
+          key: ${{ matrix.os- }}-${{ matrix.python_version }}-install-${{ env.pythonLocation }}-${{ hashFiles('**/pyproject.toml') }}-v01
       - name: Build woodwork package
-        run: make package
-      - name: Install complete version of woodwork
         run: |
-          python -m pip install --upgrade pip
-          pip config --site set global.progress_bar off
-          python -m pip install "unpacked_sdist/[complete]"
+          make package
+      - name: Install woodwork with no dependencies
```
sbadithe, line 39: Just curious, should the name include whether it was retrieved from a cache or not? Would it be useful for potential debugging?

.github/workflows/install_test.yaml
```diff
@@ -17,19 +17,33 @@ jobs:
...
+        if: steps.cache.outputs.cache-hit == 'true'
```
sbadithe, line 40: Also, I am not sure if I am reading this right, but why do we install if the cache-hit was true here, but only install if the cache-fit was false in `Install Woodwork with dependencies`?

.github/workflows/install_test.yaml
gsheni: Fixed:
https://github.com/alteryx/woodwork/pull/1631/commits/c8b01661ec5415a36cb41b2dc53d0c0575364e2c

.github/workflows/install_test.yaml
gsheni: Good catch, I had flipped the run tasks. Fixed:
https://github.com/alteryx/woodwork/pull/1631/commits/17f45f660d675ef21204a12c8b82f1714f497727

sbadithe (approved):

PR #1652, opened 2023-03-08
lint

bchen1116:
Testing

### DIFF
--- a/woodwork/statistics_utils/_get_dependence_dict.py
+++ b/woodwork/statistics_utils/_get_dependence_dict.py
@@ -103,7 +103,9 @@ def _get_dependence_dict(
     start_time = timer()
     if target_col is not None and target_col not in list(dataframe.columns):
         raise ValueError("target_col '{}' not in the dataframe".format(target_col))
-
+    boolean_columns = dataframe.ww.select(["Boolean", "BooleanNullable"]).columns.values
+    bool_to_int = {col: "IntegerNullable" for col in boolean_columns}
+    dataframe_with_bools_to_int = dataframe.ww.set_types(bool_to_int)
     returned_measures, calc_order, calc_max = _parse_measures(measures)
 
     unit = "calculations"
@@ -111,16 +113,22 @@ def _get_dependence_dict(
     # get valid columns for dependence calculations
     if "pearson" in calc_order:
         pearson_types = get_valid_pearson_types()
-        pearson_columns = _get_valid_columns(dataframe, pearson_types)
+        pearson_columns = _get_valid_columns(dataframe_with_bools_to_int, pearson_types)
         valid_columns = pearson_columns
     if "spearman" in calc_order:
         spearman_types = get_valid_spearman_types()
-        spearman_columns = _get_valid_columns(dataframe, spearman_types)
+        spearman_columns = _get_valid_columns(
+            dataframe_with_bools_to_int,
+            spearman_types,
+        )
         # pearson columns are a subset of spearman columns
         valid_columns = spearman_columns
     if "mutual_info" in calc_order:
         mi_types = get_valid_mi_types()
-        cols_to_drop = _find_large_categorical_columns(dataframe, max_nunique)
+        cols_to_drop = _find_large_categorical_columns(
+            dataframe_with_bools_to_int,
+            max_nunique,
+        )
         if len(cols_to_drop):
             warnings.warn(
                 "Dropping columns {} to allow mutual information to run faster".format(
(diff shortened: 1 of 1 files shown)


machineFL:

woodwork/statistics_utils/_bin_numeric_cols_into_categories.py
```diff
@@ -17,8 +17,8 @@ def _bin_numeric_cols_into_categories(schema, data, num_bins):
     """
     for col_name in data.keys():
         column = schema.columns[col_name]
-        if column.is_numeric or column.is_datetime:
-            # bin numeric features to make categories
+        if (column.is_numeric or column.is_datetime) and len(set(data[col_name])) > 2:
```
bchen1116, line 20: When there are 2 or fewer columns, this creates only 1 bin, which is inaccurate. Previously, this means that a column of [0, 1] values would be binned as [0] only. This allows us to maintain separate values for these scenarios, which is useful especially for boolean handling.

woodwork/statistics_utils/_get_dependence_dict.py
```diff
@@ -103,24 +103,33 @@ def _get_dependence_dict(
     start_time = timer()
     if target_col is not None and target_col not in list(dataframe.columns):
         raise ValueError("target_col '{}' not in the dataframe".format(target_col))
-
+    boolean_columns = dataframe.ww.select(["Boolean", "BooleanNullable"]).columns.values
+    bool_to_int = {col: "IntegerNullable" for col in boolean_columns}
```
bchen1116, line 107: Map boolean values to be `IntegerNullable` to allow them to be used in dependence calculations

Cmancuso:
Doesn't seem like there would be much, but did we observe any performance impact?

docs/source/release_notes.rst
```diff
@@ -22,7 +23,7 @@ Future Release
         * Run lint fix on latest dependency update pull requests (:pr:`1640`, :pr:`1641`)
 
     Thanks to the following people for contributing to this release:
-    :user:`gsheni`
+    :user:`gsheni`, :user:`bchen1116`
```
Cmancuso, line 26: @ParthivNaresh , you should be on here as well right?

woodwork/tests/accessor/test_statistics.py
```diff
@@ -108,10 +112,12 @@ def test_accessor_bin_numeric_cols_into_categories():
     assert data["ints1"].equals(pd.Series([0, 1, 3, 1], dtype="int8"))
     assert data["ints2"].equals(pd.Series([0, 1, 0, 1], dtype="int8"))
     assert data["ints3"].equals(pd.Series([0, 1, 3, 1], dtype="int8"))
+    assert data["ints4"].equals(pd.Series([0, 1, 1, 0], dtype="int8"))
```
Cmancuso, line 115: these are bools right but we are casting to int type? This will work for True / False representations and 0 / 1 representations?

docs/source/release_notes.rst
ParthivNaresh: @Cmancuso Ah yes my bad

woodwork/tests/accessor/test_statistics.py
bchen1116: Yep, it handles both bool and int representations!

bchen1116:
LG report attached.
There are noticable increases in time for pearson measurements, but this is due to the inclusion of `boolean` columns now in the calculations. The increase amounts don't raise concerns for me (the geoip2-ivp4.csv increase is due to there being only 1 column usable for pearson previously, but there being 3 now with this change).

There is a sizable (~50%) change in `validation_runtime` for the `EducationAndIncome` dataset, which shouldn't have been impacted by this change. I'm not sure why this change happened, if anyone has potential input. Ran LG twice and came with the same result.

[report.html.zip](https://github.com/alteryx/woodwork/files/10934132/report.html.zip)

ParthivNaresh (approved):
I think this is good to go. The performance regression doesn't make sense in `validate` but I don't think this PR is the culprit

woodwork/statistics_utils/_get_dependence_dict.py
```diff
@@ -103,24 +103,33 @@ def _get_dependence_dict(
...
+    dataframe_with_bools_to_int.ww.set_types(bool_to_int)
     returned_measures, calc_order, calc_max = _parse_measures(measures)
 
     unit = "calculations"
 
     # get valid columns for dependence calculations
     if "pearson" in calc_order:
         pearson_types = get_valid_pearson_types()
-        pearson_columns = _get_valid_columns(dataframe, pearson_types)
+        pearson_columns = _get_valid_columns(dataframe_with_bools_to_int, pearson_types)
```
simha104, line 117: Are the dependence calculations only being done on bool columns now?

bchen1116: It runs it on all columns! The boolean columns are just now being set properly for calculations.

Issue #1649, opened 2023-02-17
Allow `ww.type_system.add_type` a warning only option *or* provide and *exists* function


ulc0:
- As a [user/developer], I wish I could use Woodwork to add my new logical types without always erroring if it's already added. 

Current State:
- If type is not set, add_type works, delete_type fails
- If type is set, add_type fails delete_type works

Can we have an elegant solution to this problem? So I don't have to hardcode new logicals in my own fork?

ParthivNaresh:
@ulc0 Thanks for filing this issue, the new behaviour will allow for passing a `treatment` parameter to `add_type` and `remove_type`.

`add_type` will accept the values of `replace`, `ignore`, or the default of `None`.
- `replace` - If that logical type already exists, it will be unregistered and the parameters that were passed into the function call will be used instead.
- `ignore` - If that logical type already exists, any parameters that were passed into the function call will be ignored.
- `None` - The current behaviour will be defaulted to (raising an error)

`remove_type` will accept the values of `ignore` or the default of `None`.
- `ignore` - If that logical type doesn't exist, do nothing.
- `None` - A more descriptive error than what is currently provided will be raised (default).

PR #1661, opened 2023-03-20
Update add_type and remove_type to include a `treatment` parameter

ParthivNaresh:
Fixes: #1649

### DIFF
--- a/woodwork/type_sys/type_system.py
+++ b/woodwork/type_sys/type_system.py
@@ -1,3 +1,5 @@
+from typing import Callable
+
 import pandas as pd
 
 from woodwork.accessor_utils import _is_dask_series, _is_spark_series
@@ -135,7 +137,13 @@ class TypeSystem(object):
         self._default_relationships = self.relationships.copy()
         self._default_type = self.default_type
 
-    def add_type(self, logical_type, inference_function=None, parent=None):
+    def add_type(
+        self,
+        logical_type: LogicalType,
+        inference_function: Callable = None,
+        parent: LogicalType = None,
+        treatment: str = None,
+    ):
         """Add a new LogicalType to the TypeSystem, optionally specifying the corresponding inference function and a
         parent type.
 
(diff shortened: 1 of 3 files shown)


woodwork/type_sys/type_system.py
```diff
@@ -154,41 +166,57 @@ def add_type(self, logical_type, inference_function=None, parent=None):
...
+                for child in children:
+                    self.update_relationship(child, parent)
+
+            # Rebuild the relationships list to remove any reference to the removed type
+            self.relationships = [
+                rel for rel in self.relationships if logical_type not in rel
+            ]
+        else:
+            if treatment is None or treatment.lower() != "ignore":
+                raise ValueError(
```
simha104, line 217: Everything looks good to me! but should definitely have more a few more eyes look at this before merging.

simha104 (approved):

woodwork/type_sys/type_system.py
```diff
@@ -154,41 +166,57 @@ def add_type(self, logical_type, inference_function=None, parent=None):
             parent=parent,
         )
 
-        registered_ltype_names = [ltype.__name__ for ltype in self.registered_types]
-        if logical_type.__name__ in registered_ltype_names:
+        registered_ltype_names = {ltype.__name__ for ltype in self.registered_types}
+        ltype_is_registered = logical_type.__name__ in registered_ltype_names
+        if ltype_is_registered and treatment is None:
```
bchen1116, line 171: should we explicitly check that the types passed in are either `ignore`, `replace`, or `none`? In this case, a user could be passing in some typo, and the default behavior seems to treat it the same as `ignore`, which would be an issue.

woodwork/type_sys/type_system.py
```diff
@@ -145,6 +153,10 @@ def add_type(self, logical_type, inference_function=None, parent=None):
                 Defaults to None. If not specified, this LogicalType will never be inferred.
             parent (LogicalType, optional): The parent LogicalType, if applicable. Defaults to None. If not specified,
                 this type will be considered a root type with no parent.
+            treatment (string, optional): If the specified LogicalType already exists, determine the logic that should
+                be applied. Options are "replace", "ignore", or None (default). "Replace" will unregister the present
```
bchen1116, line 157: Nit: The options are all lowercase, but the next "Replace", "Ignore" mentioned are capitalized. It'd be cleaner to keep those lower as well.

PR #1666, opened 2023-04-04
Update Datetime format inference

### LINKED ISSUE #1665: Datetime Logical Type isn't able to infer dates with a format including 2 digit years and timezone
Datetime Logical Type isn't able to infer dates with a format including 2 digit years and timezone

#### How to reproduce:

```python
import pandas as pd
import woodwork as ww

dates = ['00-03-22 10:30:15+0000', '00-03-23 10:30:15+0000', '00-03-24 10:30:15+0000']
date_series = pd.DataFrame(dates)
date_series.ww.init(logical_types = {0: 'Datetime'})

print(date_series)
```
#### Output:

>     0
>     0 NaT
>     1 NaT
>     2 NaT

simha104:
- Update Datetime format inference to include formats with two digit year dates and timezones
- Closes #1665

-----


woodwork/utils.py
```diff
@@ -650,10 +650,12 @@ def _infer_datetime_format(dates, n=100):
             "%y/%m/%d",
             "%m/%d/%y %H:%M:%S",
             "%y/%m/%d %H:%M:%S",
+            "%y/%m/%d %H:%M:%S%z",
             "%d/%m/%y",
             "%y/%d/%m",
             "%d/%m/%y %H:%M:%S",
             "%y/%d/%m %H:%M:%S",
+            "%y/%d/%m %H:%M:%S%z",
```
ParthivNaresh, line 658: This doesn't support timezone for all formats with two digit years, we'd want the timezone appended to the end of every iteration of two digit years. I'd recommend adding it as a secondary check if two digit year inference fails

woodwork/tests/fixtures/datetime_freq.py
```diff
@@ -63,11 +63,10 @@ def pad_datetime_series(dates, freq, pad_start=0, pad_end=100):
             pd.date_range(start=dates[-1], periods=pad_end, freq=freq)[1:]
         ).to_series()
 
-    return (
-        pd.concat([head, pd.Series(dates), tail])
-        .reset_index(drop=True)
-        .astype("datetime64[ns]")
-    )
+    df = pd.concat([head, pd.Series(dates), tail]).reset_index(drop=True)
+    df = pd.to_datetime(df, utc=True)
```
ParthivNaresh, line 67: Is `utc=True` valid for the non-timezone use cases? I think this is converting all the datetimes to including timezones right? The tests are all passing so I think it's alright but I just wanted to check

Cmancuso: We ultimately convert everything to UTC anyway right? If we got something in a different timezone would this just convert it to UTC?

woodwork/utils.py
```diff
@@ -646,23 +646,32 @@ def _infer_datetime_format(dates, n=100):
         mode_fmt = fmts.mode().loc[0]  # select first most common format
     except KeyError:
         check_for_two_digit_years = [
-            "%m/%d/%y",
             "%y/%m/%d",
-            "%m/%d/%y %H:%M:%S",
```
Cmancuso, line 651: How come these were removed?

Cmancuso:
Couple of questions and needs a rebase, but otherwise looking good

woodwork/utils.py
simha104: It's because I add on the timestamps below in a different way that makes sure every format is paired up with a timezone rather than only a few

ParthivNaresh (approved):
Looks good!

PR #1682, opened 2023-04-18
remove repeated sorting in describe dict

### LINKED ISSUE #1042: Optimize `_get_describe_dict` to minimize repetitive computation
Many of the stats calculated by `_get_describe_dict` rely on having sorted data, (min, max, quartiles, top values, recent values), and the current implementation is likely repeating some of the sorting that happens in order to get these stats. 

For example, we do calculate all of the aggregation stats at once, but we have a separate call to `series.quantile([0.25, 0.5, 0.75])`. 

We should look into performing these calculations in a way that only sorts the data once, as most of the other computations after sorting will not be reliant on the length of the data.

simha104:
- Remove repeated sorting in describe dict
- Closes #1042

### DIFF
--- a/woodwork/statistics_utils/_get_describe_dict.py
+++ b/woodwork/statistics_utils/_get_describe_dict.py
@@ -152,14 +152,30 @@ def _get_describe_dict(
             agg_stats = ["count"]
         values = series.agg(agg_stats).to_dict()
 
+        mode = _get_mode(series)
+        # The format of the mode should match its format in the DataFrame
+        if _is_spark_dataframe(dataframe) and series.name in latlong_columns:
+            mode = list(mode)
+
+        if column.is_latlong:
+            nan_count = series.apply(_is_latlong_nan).sum()
+            count = len(series) - nan_count
+
+            values["nan_count"] = nan_count
+            values["count"] = count
+        else:
+            values["nan_count"] = series.isna().sum()
+
+        null_pct = values["nan_count"] / (values["nan_count"] + values["count"])
+        _use_manual_sort = null_pct < 0.05
+
         # Calculate other specific stats based on logical type or semantic tags
         if column.is_boolean:
             values["num_false"] = series.value_counts().get(False, 0)
             values["num_true"] = series.value_counts().get(True, 0)
         elif column.is_numeric:
-            percent_missing = 1 - (values["count"] / len(series))
-            _use_manual_sort = percent_missing < 0.5
-            if _use_manual_sort:
+            1 - (values["count"] / len(series))
+            if series.isna().sum():
                 series = series.sort_values(
                     ignore_index=True,
                 )
@@ -186,20 +202,6 @@ def _get_describe_dict(
                 values["first_quartile"] = quant_values[0]
                 values["second_quartile"] = quant_values[1]
                 values["third_quartile"] = quant_values[2]
-
-        mode = _get_mode(series)
-        # The format of the mode should match its format in the DataFrame
-        if _is_spark_dataframe(dataframe) and series.name in latlong_columns:
-            mode = list(mode)
-
-        if column.is_latlong:
-            nan_count = series.apply(_is_latlong_nan).sum()
-            count = len(series) - nan_count
-
-            values["nan_count"] = nan_count
-            values["count"] = count
-        else:
-            values["nan_count"] = series.isna().sum()
 
         values["mode"] = mode
         values["physical_type"] = series.dtype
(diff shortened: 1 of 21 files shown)


ParthivNaresh:
Can we get perf tests for this?

simha104:
> Can we get perf tests for this?

Yeah sure, is there a place in woodwork where I should put a performance test in? or should I just test it locally and provide the results here?

ParthivNaresh:
@simha104 Looks good! Can you squash these commits?

ParthivNaresh (approved):
LGTM

woodwork/statistics_utils/_get_describe_dict.py
```diff
@@ -25,6 +26,27 @@
...
+    Args:
+        N (pd.Series): Series is a list of values. Note N MUST BE already sorted.
+        percent (float): float value from 0.0 to 1.0.
+        count (int): Count of values in series
+
+    @return - the percentile of the values
+    """
+    k = (count - 1) * percent
+    f = math.floor(k)
+    c = math.ceil(k)
```
tamargrey, line 42: Could we get some more descriptive variable names and maybe some comments outlining what each step in the percentile calculation is? Or, since this seems to match the scipy implementation (from [this](https://stackoverflow.com/a/2753343/15489857) stack overflow answer), maybe just linking to that would. be enough.

tamargrey (commented):
Can we get a test that confirms that if nans are present, the current way gets used? I'm thinking you could mock `percentile` and check if that gets called or not.

woodwork/statistics_utils/_get_describe_dict.py
```diff
@@ -25,6 +26,27 @@
...
+def percentile(N, percent, count):
+    """
+    Find the percentile of a list of values.
+
+    Args:
+        N (pd.Series): Series is a list of values. Note N MUST BE already sorted.
+        percent (float): float value from 0.0 to 1.0.
+        count (int): Count of values in series
+
+    @return - the percentile of the values
```
bchen1116, line 40: nit on return formatting: 
```python
Returns:
     The percentile of the values.
```

woodwork/statistics_utils/_get_describe_dict.py
simha104: I've included the source now

woodwork/tests/accessor/test_statistics.py
```diff
@@ -1240,6 +1240,40 @@ def test_describe_accessor_method(describe_df):
...
+
+    # Test numeric with non-nullable ltypes
+    numeric_data = describe_df[["numeric_col"]].dropna()
+    for ltype in nullable_numeric_ltypes:
+        numeric_data.ww.init(
+            logical_types={"numeric_col": ltype},
+            semantic_tags={"numeric_col": "custom_tag"},
+        )
+        numeric_data.ww.describe()
+        assert mock_percentile.called
```
tamargrey, line 1265: By doing `mock_percentile.called` in a for loop, you're only really confirming that the first logical type has `percentile` called (since `mock_percentile.called` will stay `True` for the rest of the test). My recommendation would be to parametrize over the different nullable logical types; then, each time the test is created it has a fresh `mock_percentile`. 

Additionally, the check below with the non nullable types would run into a similar issue bc once the mock has been called, we don't learn anything from confirming that it's still called. I think it may just be simplest to make a totally separate test for the non nullable ltypes and only check that it's not called prior to the describe called and is called afterwards. I'm not totally in love with that, though, and would be open to other solutions!

woodwork/tests/accessor/test_statistics.py
```diff
@@ -1254,24 +1253,21 @@ def test_percentile_func_usage(mock_percentile, describe_df):
...
-    for ltype in non_nullable_numeric_ltypes:
-        numeric_data.ww.init(
-            logical_types={"numeric_col": ltype},
-            semantic_tags={"numeric_col": "custom_tag"},
-        )
-        numeric_data.ww.describe()
-        assert mock_percentile.called
 
+@patch.object(sys.modules["woodwork.statistics_utils._get_describe_dict"], "percentile")
+@pytest.mark.parametrize("non_nullable_numeric_type", [Integer, Age])
```
tamargrey, line 1258: one more thing! 

Can you add [Double, IntegerNullable, AgeNullable, AgeFractional] to the parametrization? It's a small thing, but we'll want to know that it's not just the type that is defining whether or not percentile gets used; it's whether nans are present. I would also update the test names to `test_percentile_func_not_called_with_nans` and `test_percentile_func_called_without_nans`

tamargrey (approved):
Once that last comment is done, you're good to go!

PR #1689, opened 2023-04-26
Remove `autouse=True` for LatLong fixtures

### LINKED ISSUE #1690: Remove `autouse=True` for LatLong Fixtures
- Making the LatLong fixtures autoused means that if a user does not have `pyspark` installed (for example), tests are skipped that don't need to be. Furthermore, it leads to unneeded repeated computation
- In a subsequent issue, we should investigate the use of autouse in other fixtures

sbadithe:
- Fixes #1690


machineFL:

PR #1692, opened 2023-05-01
Stop calculating `top_values` for Double columns with integer values

### LINKED ISSUE #1691: Remove top_values calculation for Double columns with integer values
We previously added logic into `_get_describe_dict` to calculate `top_values` for columns with the `Double` logical type whose values are all integers in https://github.com/alteryx/woodwork/issues/1203. We should remove that logic. This will allow us to just check if the logical type is one of the integer or age logical types instead of checking if every value could be converted to integer.

tamargrey:
- Previously, we needed to calculate top values for Double columns with integer values, but now that we no longer need that behavior, we should remove it from `_get_describe_dict`
- Closes #1691

### DIFF
--- a/woodwork/statistics_utils/_get_describe_dict.py
+++ b/woodwork/statistics_utils/_get_describe_dict.py
@@ -4,7 +4,15 @@ from typing import Any, Callable, Dict, Sequence
 import pandas as pd
 
 from woodwork.accessor_utils import _is_dask_dataframe, _is_spark_dataframe
-from woodwork.logical_types import Datetime, Integer, IntegerNullable, LatLong, Unknown
+from woodwork.logical_types import (
+    Age,
+    AgeNullable,
+    Datetime,
+    Integer,
+    IntegerNullable,
+    LatLong,
+    Unknown,
+)
 from woodwork.statistics_utils._get_histogram_values import _get_histogram_values
 from woodwork.statistics_utils._get_mode import _get_mode
 from woodwork.statistics_utils._get_numeric_value_counts_in_range import (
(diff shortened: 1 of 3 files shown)


woodwork/tests/accessor/test_statistics.py
```diff
@@ -1369,7 +1369,8 @@ def test_describe_callback(describe_df, mock_callback):
     assert mock_callback.total_elapsed_time > 0
 
 
-def test_describe_dict_extra_stats(describe_df):
+@pytest.mark.parametrize("use_age", [True, False])
+def test_describe_dict_extra_stats(use_age, describe_df):
```
tamargrey, line 1373: I wanted to make sure to cover age usage and confirm its behavior matches int and double usage, but I recognize that this may be a bit overkill. If anyone would rather I make a completely separate test that more minimally tests age vs integer usage, lmk.

bchen1116 (approved):
Not sure why the install for 3.11 is failing, but changes LGTM

woodwork/statistics_utils/_get_describe_dict.py
```diff
@@ -166,16 +174,19 @@ def _get_describe_dict(
...
-                    range_len = int(values["max"]) + 1 - int(values["min"])
-                    if range_len <= bins and (series % 1 == 0).all():
-                        values["top_values"] = _get_numeric_value_counts_in_range(
-                            series,
-                            _range,
-                        )
+                    if isinstance(
+                        column.logical_type,
+                        (Age, AgeNullable, Integer, IntegerNullable),
+                    ):
```
thehomebrewnerd, line 180: Not something we need in this PR, but might be nice if we had an `is_integer_type` property on the column schema to use instead of having to maintain a list of logical types here.

thehomebrewnerd (approved):

PR #1695, opened 2023-05-04
Lg airflow merge perf tests

petejanuszewski1:
- Pull Request Description (Looking glass performance tests run on merge through Airflow)

-----

### DIFF
--- a/.github/workflows/woodwork_main_airflow_performance_tests.yaml
+++ b/.github/workflows/woodwork_main_airflow_performance_tests.yaml
@@ -15,7 +15,7 @@ jobs:
     steps:
       - name: Generate ISO timestamp
         run: |
-          TIMESTAMP=$(date +"%Y-%m-%dT%H:%M:%S.%3Z")
+          echo "TIMESTAMP=$(date +"%Y-%m-%dT%H:%M:%S.%3Z")" >> $GITHUB_ENV
       - name: Checkout woodwork
         uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


machineFL:

.github/workflows/woodwork_main_airflow_performance_tests.yaml
```diff
@@ -0,0 +1,50 @@
...
+          curl --location --request POST '${{ secrets.AIRFLOW_BASE_URL }}dags/woodwork_run_tests_generate_report/dagRuns' \
+          -u '${{ secrets.AIRFLOW_WW_USER }}:${{ secrets.AIRFLOW_WW_PASS }}' \
+          --header 'Content-Type: application/json' \
+          --data-raw '{
+            "conf": {
+                  "description": null,
+                  "lg_branch": "lg_baton_airflow",
+                  "n_trials": 1,
+                  "pytest_args": {},
+                  "python_version": "3.8",
```
ParthivNaresh, line 40: Is there any benefit to using a more recent Python version here?

petejanuszewski1: that sets the python version on which we run the woodwork tests. New LG has support for 3.8, 3.9, 3.10, and 3.11. We were previously testing on 3.8, but if all ww deps are compatible we can increase the version. Might need to do some testing to verify first, then update the payload as needed.

PR #1702, opened 2023-05-15
Improve logical type inference

### LINKED ISSUE #1347: Default inference not picking up URL features on Restaurant_Likes.csv

```python
import pandas as pd
import woodwork as ww

X = pd.read_csv("/Users/freddy.boulton/Downloads/Restaurant_Likes.csv")
X.ww.init()

assert X.ww.schema.columns['Restaurant URL'].logical_type.type_string == 'categorical'
```

```
!pip freeze | grep -E "pandas|woodwork"
```
```
pandas==1.4.0
woodwork==0.14.0
```

image

[Restaurant_Likes.csv](https://github.com/alteryx/woodwork/files/8317278/Restaurant_Likes.csv)

bchen1116:
fix #1347

### DIFF
--- a/woodwork/type_sys/type_system.py
+++ b/woodwork/type_sys/type_system.py
@@ -91,6 +91,12 @@ DEFAULT_RELATIONSHIPS = [
     (Integer, Age),
     (IntegerNullable, AgeNullable),
     (IntegerNullable, Integer),
+    (Unknown, EmailAddress),
+    (Unknown, IPAddress),
+    (Unknown, PhoneNumber),
+    (Unknown, PostalCode),
+    (Unknown, Timedelta),
+    (Unknown, URL),
 ]
 
 DEFAULT_TYPE = Unknown
@@ -399,11 +405,13 @@ class TypeSystem(object):
         types_to_check = [
             ltype for ltype in self.root_types if ltype != NaturalLanguage
         ]
+        print(types_to_check)
         series_nan_cast = _replace_nans(series)  # Will change dtype
         if series_nan_cast.count() == 0:
             return Unknown()
 
         type_matches = get_inference_matches(types_to_check, series_nan_cast)
+        print(type_matches)
         if len(type_matches) == 0:
             # Check if this is NaturalLanguage, otherwise set
             # type to default type (Unknown). Assume that a column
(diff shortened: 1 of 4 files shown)


woodwork/type_sys/type_system.py
```diff
@@ -399,11 +405,13 @@ def get_inference_matches(types_to_check, series, type_matches=[]):
         types_to_check = [
             ltype for ltype in self.root_types if ltype != NaturalLanguage
         ]
+        print(types_to_check)
```
Cmancuso, line 408: I assume we want to remove these prints?

Cmancuso:
Can we run evalML tests against this to see how their suite will react?

woodwork/type_sys/type_system.py
```diff
@@ -91,6 +91,12 @@
     (Integer, Age),
     (IntegerNullable, AgeNullable),
     (IntegerNullable, Integer),
+    (Unknown, EmailAddress),
+    (Unknown, IPAddress),
+    (Unknown, PhoneNumber),
```
Cmancuso, line 96: should any of these have categorical parents?

bchen1116: My idea here was that these values don't necessarily have categorical-like values, since there's an infinite number of possible values and when they appear in datasets, they have the potential to be all-unique. This is different from when `CountryCode` or `CurrencyCode` appear, since there's a limited number of values these could be. That's why I decided to opt for `Unknown` as the parent, but let me know what you think!

Cmancuso: Hmm, yeah I mean it definitely depends on the dataset honestly. If there are few enough distinct values, it probably should be categorical, and if not, it should be Unknown. I don't suppose there's any way to represent something like that currently is there?

bchen1116: Yep, currently no, but we only use the parent-child relationships defined to determine which value we prefer to default to. If a dataset matches with our Categorical inference and URL inference, it'll default to URL now. I believe if it matches with Categorical and Unknown, it will choose the first match, which will be Categorical.

bchen1116:
All tests on EvalML passed locally!
image

Cmancuso (approved):

PR #1712, opened 2023-06-13
remove autouse from latlong fixtures

### LINKED ISSUE #1690: Remove `autouse=True` for LatLong Fixtures

simha104:
- Pull Request Description (replace this text with your description)
- Closes #1690

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -8,10 +8,12 @@ Release Notes
     * Enhancements
     * Fixes
     * Changes
+      * latlong fixtures are no longer autouse (:pr:`1712`)
     * Documentation Changes
     * Testing Changes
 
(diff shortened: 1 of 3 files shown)


docs/source/release_notes.rst
```diff
@@ -8,10 +8,12 @@ Release Notes
     * Enhancements
     * Fixes
     * Changes
+      * latlong fixtures are no longer autouse (:pr:`1712`)
```
bchen1116, line 11: can we put `latlong` inside double ticks

PR #1713, opened 2023-06-16
optimize boolean inference

### LINKED ISSUE #1714: Optimize boolean inference
EvalML’s recent 0.65.0 suffered a degradation in performance for a few of their larger datasets. This is likely due to the additional boolean inference that occurs in WW, especially for object dtype columns. This issue tracks optimizing the inference and transformation of boolean columns. 

Quick possible fixes that would help are:
- Don’t generate the set of possible boolean values in the inference. Include that in the config directly
- Don’t generate the mapping of booleans in the transform. Include that in the config or somewhere globally.

simha104:
- Optimize boolean inference
- Closes #1714

-----


machineFL:

simha104 (commented):
There is a very small/negligible performance improvement that can be seen in the `boolean_heavy` section of the report
[Notebook.pdf](https://github.com/alteryx/woodwork/files/11803301/Notebook.pdf)

bchen1116:
Hmm, I'm curious if there are other things we can do to further optimize this inference and transformation. For instance, maybe replacing the `series.apply(lambda x:...)` function in the transform with a `series.map(boolean_transform_mappings)` equivalent? `map` seems to cut the time significantly compared to `apply`. 

Also, seems like casting a series into a set takes longer than calling `.unique()`, which could be another small thing we change to speed this up a little
image

simha104 (commented):
New performance improvement of 39% in Boolean Inference!
[report.pdf](https://github.com/alteryx/woodwork/files/11835713/report.pdf)

woodwork/type_sys/inference_functions.py
```diff
@@ -115,17 +115,10 @@ def boolean_nullable_func(series):
         series_no_null = series.dropna()
         try:
             series_no_null_unq = set(series_no_null)
```
sbadithe, line 117: Should we apply the `.unique()` suggestion here as well?

woodwork/type_sys/inference_functions.py
```diff
@@ -115,17 +115,10 @@ def boolean_nullable_func(series):
...
-                {False},
-            ]:
+            if series_no_null_unq.issubset({False, True}):
                 return True
-            series_lower = set(str(s).lower() for s in set(series_no_null))
-            if series_lower in [
-                set(boolean_list)
-                for boolean_list in config.get_option("boolean_inference_strings")
-            ]:
+            series_lower = set(str(s).lower() for s in series_no_null_unq)
```
sbadithe, line 120: Just wondering, if the length of `series_no_null_unq` is greater than two, is there a scenario where we still need to do this computation?

sbadithe: I can't comment on GitHub on this particular line in the MR, but I was wondering the following about null handling: 

On the parent function `boolean_func`, we call `boolean_nullable_func` and then check if there are any nulls -- `not series.isnull().any()`. I think we can leverage short circuiting and flip these two for potential speedup in boolean inference.

woodwork/type_sys/inference_functions.py
simha104: I decided not to do that because `.unique` only really provides a speed improvement for series with an `integer` dtype. I could put it in there for consistency though

PR #1733, opened 2023-07-15
Cache results of `IntegerNullable` and `BooleanNullable` inference 

### LINKED ISSUE #1033: Improving type inference performance
There are potentially several ways type inference efficiency could be improved. This issue is being created to capture these ideas.

- For Dask we use the first partition of data for type inference and for Koalas the first 100,000 rows of data. For pandas all rows are used. We could use only a portion of the data for pandas, which might speed up some inference functions (possibly sacrificing inference accuracy).
- If some type of sampling is implemented as suggested above, this could be a user controllable config setting - specifying the method (random vs first n) and the number of rows to use so users have more control over the speed/accuracy tradeoff involved.
- The current `col_is_datetime` function attempts to call `pd.to_datetime` to determine if the column can be cast to datetime. We might want to look into using the code in `_infer_datetime_format` being implemented in PR #1025 to see if a similar approach could be used in `col_is_datetime` that might be more efficient. Perhaps if we can quickly infer a datetime format, we can assume the column contains datetime data without having to perform the full conversion.
- We should look for instances where inference functions compute and use the same pieces of information. If duplicate calculations are being performed, we can cache these results and share them among functions to eliminate the duplicate calculations.

sbadithe:
- Partially fixes #1033
- Currently, the inference function for the Boolean logical type calls the inference functions for BooleanNullable. However, considering that the Boolean is a child of its corresponding Nullable type, we know that the Nullable inference function must have already been computed by the time we are performing Boolean inference. By caching the result of the BooleanNullable inference, we can avoid recomputing its result. 

- We can apply this same argument to Integer/IntegerNullable to speed up performance. 

Preliminary performance testing points to a ~84% speed up on `boolean_heavy` data and a ~78% speed up on `numeric_heavy` data.


machineFL:

woodwork/type_sys/inference_functions.py
```diff
@@ -100,8 +106,13 @@ def double_func(series):
     return False
 
 
-def boolean_func(series):
-    if boolean_nullable_func(series) and not series.isnull().any():
+def boolean_func(series, is_boolean_nullable=None):
```
sbadithe, line 109: Admittedly, adding an argument that defaults to None is probably not where we want to land long-term, but I think this is a fair short-term fix. I think a better long term solution would be to generalize the idea of caching in the `get_inference_matches` function itself.

woodwork/type_sys/type_system.py
```diff
@@ -391,9 +391,23 @@ def get_inference_matches(types_to_check, series, type_matches=[]):
             check_next = []
             for logical_type in types_to_check:
                 inference_func = self.inference_functions.get(logical_type)
-                if inference_func and inference_func(series):
-                    type_matches.append(logical_type)
-                    check_next.extend(self._get_children(logical_type))
+                if inference_func:
```
sbadithe, line 394: I wasn't sure how much we would gain by adding a test for the special logic here. I'm leaning towards the thought that testing this would be testing implementation details and not the desired end result.

bchen1116 (approved):
LGTM! Nice job speeding this up dramatically!

PR #1862, opened 2024-05-10
Use filter arg to safe extract archives

### LINKED ISSUE #1860: Add filter arg to tarfile.extractall call to safely deserialize dataframes
Update `tarfile.extractall` call to safely deserialize dataframes.

thehomebrewnerd:
Closes #1860 

The filter argument was added in these Python versions: 3.9.17, 3.10.12, 3.11.4, 3.12.0. If the user is running an older Python version an error will be thrown and they will need to upgrade to one of these patch releases (or newer) to extract the archive.


jeremyliweishih (approved):
LGTM

